PR #5, opened 2017-07-21
added live ticker feature

db383:
added live ticker feature to wallet dashboard, polling bittrex api for prices, refresh every 5 seconds.


Ejhfast:
Merged after a few fixes through https://github.com/CityOfZion/neo-electron-wallet/pull/6

PR #7, opened 2017-07-23
Wallet improvements

cinooo:
Some general bug fixes, ui improvement and starting to add some validations to sending transactions

### DIFF
--- a/app/actions/index.js
+++ b/app/actions/index.js
@@ -74,10 +74,14 @@ export function toggleAsset(){
 };
 
 // dashboard
-
-export function togglePane(pane){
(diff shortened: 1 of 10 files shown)


Ejhfast:
Hey cinooo, I like the clipboard functionality for the private and public keys! The send page in this pull request seems to be broken, though? Clicking send no longer toggles the mini window size and confirm appears on the right hand side? I also think we want explicit error messages for address/amount validation, as opposed to simply not showing the confirm button

screen shot 2017-07-25 at 10 26 35 pm

cinooo:
Heya! Yep agree with everything you've said, let me add those validation messages

PR #16, opened 2017-07-30
Local storage

nickemmons:
Integrated local storage to store your wallets for future logins.  When you generate a new wallet, your wif key is added to the local storage.  All the wallets you generate are displayed in a grid below the login input.  When you click on a wallet, you are automatically logged in and redirected to the dashboard.
screen shot 2017-07-30 at 3 38 40 am


Ejhfast:
Hey @EmmonsVentures, looks like a great start! A few questions and comments:

1. Are wifs encrypted with a separate user password before being loaded in local storage? I think this is necessary for security. 
2. I think the interface we want for multiple accounts would support login with a single user password, that would then decrypt all the wif keys associated with that user's accounts. This looks a bit different than what you are doing now? This way we can also safely generate a multi-key backup file.
3. More minor and we can work out the details, but I think we can do account navigation through a more list-like interface, maybe sliding in off the left column? Jaxx has some good inspiration here

EmmonsVentures:
Hey @Ejhfast thanks for the feedback! Those are good points about the password protection.  I'll get to work on that asap.  And I'll check out Jaxx to get a better idea about the list layout.  I'll send some pics to double check it's going in the right direction.  I'll get to work on all of this stuff first thing in the morning.

anfn101:
Another even more minor thing- it would be nice if you could add the ability to change the identification of certain wallets (like "Work Wallet" or "College Funds", custom) instead of Wallet 1, 2, 3, ... although 1, 2, 3, could be the defaults. Don't focus on this yet, but could be good to have.

canesin:
giving my 2cents since @Ejhfast is the lead on this:

well I personally disagree with storing the WIF in local storage... but COZ guys know I`m a security freak.. my view is that there should be a only a stored password for memory encryption of the WIF and that a better usage for local storage would be to hold a address book, encrypted with the same password.

wouterbruinings:
I agree with canesin. I would not want to use the wallet if I knew it was stored the WIF unencrypted.

Ejhfast:
@wouter that won't be happening, don't worry

danielfm123:
You can use that neo web wallet https://www.myneowallet.io

danielfm123:
You can use that neo web wallet https://www.myneowallet.io

Ejhfast:
hey @nickemmons, closing since this is now available in 0.0.5

PR #18, opened 2017-07-31
Setup author format for deb packing

canesin:
This is the format for email and url string in npm package.. it is required for building deb and rpm packages.


Ejhfast:
Thanks!

canesin:
This is replaced by #23

PR #24, opened 2017-08-01
Alpha release

Ejhfast:
version 0.0.2

### DIFF
--- a/app/components/TransactionHistory.js
+++ b/app/components/TransactionHistory.js
@@ -28,7 +28,19 @@ class TransactionHistory extends Component {
       <div className="columnHeader">Transaction History</div>
       <div className="headerSpacer"></div>
       <ul id="transactionList">
-        {this.props.transactions.map((t) => <li><div className="txid" onClick={() => openExplorer(getExplorerLink(this.props.net, t.txid))}>{t.txid.substring(0,32)}</div><div className="amount">{t.amount} {t.type}</div></li>)}
+        {this.props.transactions.map((t) => {
+          let formatAmount;
+          if (t.type === "NEO"){ formatAmount = parseInt(t.amount); }
+          else{ formatAmount = parseFloat(t.amount).toPrecision(5); }
+          // ignore precision rounding errors for GAS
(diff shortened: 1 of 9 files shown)


app/components/TransactionHistory.js
```diff
@@ -28,7 +28,19 @@ class TransactionHistory extends Component {
...
       <div className="headerSpacer"></div>
       <ul id="transactionList">
-        {this.props.transactions.map((t) => <li><div className="txid" onClick={() => openExplorer(getExplorerLink(this.props.net, t.txid))}>{t.txid.substring(0,32)}</div><div className="amount">{t.amount} {t.type}</div></li>)}
+        {this.props.transactions.map((t) => {
+          let formatAmount;
+          if (t.type === "NEO"){ formatAmount = parseInt(t.amount); }
+          else{ formatAmount = parseFloat(t.amount).toPrecision(5); }
+          // ignore precision rounding errors for GAS
+          if ((formatAmount > 0 && formatAmount < 0.001) || (formatAmount < 0 && formatAmount > -0.001)){
+            formatAmount = 0.0.toPrecision(5);
```
weppos: @Ejhfast I just happened to find this commit as a result of a couple of issues I was investigating:

1. All my GAS transactions appeared as `0.000` (e.g. `0.0000312` -> `0.000`)
2. NEO transactions appeared to be formatted as integer (e.g. `10.24` -> `10`)

and it looks like this seems to be the culprit. I tried to understand what kind of "error on fractional" you were experiencing, but that would require quite a bit of investigation.

For what is worth, it's a quite common habit to use 8 digits precision for crypto currencies. I'm well aware of Floating number rounding errors, that seems to be the main reason here. For the same reason, money amounts are generally handled with integers (in this case it would be an int where the right-most 8 digits are the decimal part) or a decimal library (JS is not my primary language, but I happened to find a couple of implementations including [`big.js`](https://github.com/MikeMcl/big.js))

I may provide a PR that fixes the 0 values and int rounding, if you don't mind to provide some details about the issue you were facing here.

PR #25, opened 2017-08-02
added comma delimination to NEO-USD live display

db383:
Added comma separation to live USD display.
Original: $12345.78 -> Now: $12,345.78

### DIFF
--- a/app/wallet/api.js
+++ b/app/wallet/api.js
@@ -85,21 +85,17 @@ export const getBalance = (net, address) => {
 /**
  * @function
  * @description
- * Hit the bittrex api getticker to fetch the latest BTC to ANS price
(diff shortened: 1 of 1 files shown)


Ejhfast:
sorry for the delay on this, but could you update to use the cmc api (we swapped bittrex for that in master)

db383:
Updated NEO - USD conversion query from using two bittrex api queries to a single coinmarketcap api query.

Ejhfast:
hey sorry for confusion, so the api has been moved to neon-js, which is now an import. can you pull there instead? thanks!

Ejhfast:
i didn't notice right away as you were pull requesting to alpha-release branch

Issue #32, opened 2017-08-05
Replace Bittrex API method ANS => BTC => USD with single Coinmarketcap query


Ejhfast:
Single query is better than the two we have now: https://api.coinmarketcap.com/v1/ticker/NEO/?convert=USD

vergilbg:
Does anybody know how bittrex gets the conversion rate from btc to usd?  
In the UI,  when you look at a market,  for example neo,  above the btc value shows (last) the usd value is shown too,  but this usd value is not available through the api. It would be useful either making usd value available on api or if bittrex can share the api they use for the conversion (assuming they use external data for the conversion). 

The above url wouldn't  be very accurate. I have tested and there is difference in btc values coming frim bittrex and from coinmarketcap. 

Whatb I have done to make this bit more accurate,  is getting btc and usd values from coinmarketcap,  find the conversion for 1 btc (btc/usd)  then apply that to the btc fetched from bittrex (btc*usd).  That still gannot guarantee the accuracy.  Requests must be fired simultaneously. If there is something better,  please let me know.

babatundebusari:
@vergilbg interested in this as well...will like to know how to get the conversion on bittrex as well...not sure if they use USDT for their conversion

PR #35, opened 2017-08-06
Updated getMarketPriceUSD function to use USDT-NEO Bittrex API call

hbibkrim:
#32 

Fixing Github issue

By updating the bittrex getticker URL to use coinmarketcap

Manual test

Nope

Nope

- [ ] Unit tests written?

### DIFF
--- a/app/wallet/api.js
+++ b/app/wallet/api.js
@@ -73,20 +73,15 @@ export const getBalance = (net, address) => {
 /**
  * @function
  * @description
- * Hit the bittrex api getticker to fetch the latest BTC to ANS price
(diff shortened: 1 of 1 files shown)


hbibkrim:
Possible further work:
- Introduce some error handling if the getticker response.success returned false

Ejhfast:
yeah error handling is important, i am actually worried our current build will break when bittrex updates their API. @fabwa suggested that we should shift over to coinmarketcap, which i agree with. would you be able to update the pull with that API instead?

hbibkrim:
@Ejhfast Updated

PR #36, opened 2017-08-06
Unit test infrastructure

shawnmclean:
Unit Testing

 - Tests that the Login component/container is calling the LOGIN action as proof that infrastructure works.
 - Test the combination of the reducer
 
 - Jest and Enzyme testing frameworks used.
 - Made some changes to the Login component to make it more testable by replacing ref node input with accepting the event target's value directly.
 - Refactored reducers into its own file with `index.js` being the combiner.

Manual logging in to TestNet and playing around.

Executing tests with command `jest`, `npm install -g jest` may be needed.

I'd recommend moving the redux parts of `Login` (and other components where this pattern is used) to its own container and have a dummy Login component. This allows jest to snapshot the generated html from the component for review and testing the container does not rely on any DOM testing code.

- [x] Unit tests written?

### DIFF
--- a/__mocks__/fileMock.js
+++ b/__mocks__/fileMock.js
@@ -1,0 +1,1 @@
+module.exports = 'test-file-stub';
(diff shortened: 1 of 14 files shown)


Ejhfast:
looks great! merged

PR #37, opened 2017-08-06
Redux refactored to use ducks pattern

shawnmclean:
Unit Tests and maintenance

  - This groups constants, actions and reducers into their own singular module file. Keeps files consolidated and easier to maintain.
  - Documentation made clearer.
  - Version is being pulled from `package.json` instead of being hard coded.

Full regression test on TestNet.

- [x] Unit tests written?


Ejhfast:
looks great! unfortunately the current build is breaking for me, not sure what is going on there?

screen shot 2017-08-07 at 2 56 05 pm

could you also integrate the refactored lib neon-js changes from current master?

shawnmclean:
@Ejhfast Pulled in latest master and fixed conflicts. Try again.

Ejhfast:
Awesome work, thanks!

Issue #46, opened 2017-08-07
Can't copy/paste the addresses on Ubuntu Linux.


jaronfitz:
I can click the copy buttons and highlight/use copy shortcuts, but it doesnt actually capture.

Ejhfast:
also, would be great to allow right click copy/paste on windows/mac

dop89:
Hi,

I have experienced the same issue. The reason is the second parameter of the `clipboard.writeText()` function. Without this "type" parameter, everything works fine on Arch Linux. I can create a PR for this fix.

PR #50, opened 2017-08-08
fixes issue with not working copy-to-clipboard button

dop89:
https://github.com/CityOfZion/neon-wallet/issues/46

It enables the user to use the copy-to-clipboard button again. Tested on Arch Linux

the `clipboard.writeText()` function has a next to the actual text a second type parameter. This type parameter seems to make problems. Without this optional parameter, the text is copied to the clipboard properly. 

See here: https://github.com/electron/electron/blob/master/docs/api/clipboard.md

I tested in on my local PC (Arch Linux with Gnome 3.24)

I don't think so

No

- [ ] Unit tests written? -> No

### DIFF
--- a/app/components/CreateWallet.js
+++ b/app/components/CreateWallet.js
@@ -44,12 +44,12 @@ class CreateWallet extends Component {
       <div className="keyList">
         <span className="label">Public Address:</span>
         <span className="key">{this.props.address}</span>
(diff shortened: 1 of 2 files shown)


canesin:
@dop89, [join us in slack](https://join.slack.com/t/neosmarteconomy/shared_invite/MjIyOTc2OTU4NDk3LTE1MDIxNzE2NTctZTE1Mjg5ZWE3Yw)! I will provide testnet NEO/GAS to you to test your contributions.

Ejhfast:
thank you!

Issue #44, opened 2017-08-07
should also have a chinese yuan(RMB) ticker from yunbi


ethanonfire:
thank you Ethan!

Ejhfast:
good idea, we can integrate this one we have a settings pace, other currencies as well

Ejhfast:
closing this given the (unfortunate) china exchange situation, hopefully we can re-open at some future time

PR #51, opened 2017-08-08
Enables switching of price between USD and EUR

dop89:
https://github.com/CityOfZion/neon-wallet/issues/44

Currently, the price of NEO is only displayed (hardcoded) in USD. With my approach, the currency is more dynamic and can be switched in the first approach between USD and EUR. Adding further currencies should be easy. 

I created a new React Component (Balance), with is responsible for fetching the balance and converting it into the current market price. This component is reusable and can be used in other parts of the application. With a click on the currency label, the price is switched between USD and EUR. For proper currency formatting, a node.js library was used. 

Unfortunately I don't know how to receive NEO on TestNet. Therefore, I only checked it with passing a manual price into the `getMarketPrice` method. If you can provide me test NEO, I'll do some test transaction and check whether the price is updated or not. 

The `getMarketPrice` method should be moved to neon-js. I'll create a separate PR for this. After that, this method can be deleted in this codebase and imported from the new neon-js version.  

I'm pretty new to React and this was basically my first approach. Please consider this PR more like an suggestion instead of a finished and well tested implementation. I'm open for your feedback. 

- [ ] Unit tests written?

### DIFF
--- a/app/actions/index.js
+++ b/app/actions/index.js
@@ -23,12 +23,13 @@ export function newWallet(){
   }
 }
 
-export function setBalance(neo, gas, price){
+export function setBalance(neo, gas, price, currencyCode){
   return {
     type: types.SET_BALANCE,
     Neo: neo,
     Gas: gas,
-    price: price
+    price: price,
+    currencyCode: currencyCode
   }
 }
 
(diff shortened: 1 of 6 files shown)


Ejhfast:
@dop89 we'd really love this feature, but implemented on the setting page now available in 0.0.5 -- if you add there, happy to merge

shawnmclean:
Closing due to inactivity. Please open a new pull request if you decide to continue on this enhancement. :)

PR #58, opened 2017-08-09
Add test for Claim, CreateWallet, Logout component

shionwahch:
None.

Test coverage for components.

N/A

N/A

N/A

Have not covered some of the dispatches in Claim component that uses timeout.

- [ ] Unit tests written?
N/A


Ejhfast:
@shionwahch sorry for the delay on this, dev was moving too fast to keep up with PRs! if you fix the merge conflicts with 0.0.5 I would be happy to test and merge

Issue #59, opened 2017-08-09
Rounding of GAS value should be floor.


canesin:
When rounding amount of avaliable GAS it should be floored to precision (not rounded), a user that tries to transfer at precision limit (0.001) my be mistaken of his possessions.

**proposed algo**:

let gas_disponible_padded = gas_disponible * 10000;
let gas_disponible_display_padded = floor(gas_disponible_padded);
let gas_disponible_display =  gas_disponible_display / 10000;

PR #62, opened 2017-08-10
Round Gas value down to prevent trying to oversell

hbibkrim:
#59 

Preventing users from trying to sell more gas than they own because displayed amount is rounded up

Using suggested solution

Manual test on Testnet

Nope

Nope

- [ ] Unit tests written?


hbibkrim:
Or would you prefer to round gas down during initiateGetBalance?

canesin:
This looks good @hbibkrim, if @Ejhfast don't give a opinion I will merge latter..

Ejhfast:
looks good, I will merge once I do a quick independent test (do not like merging code without running myself), @canesin feel free to do so yourself if you have time

Ejhfast:
thanks @hbibkrim! sorry about how slow i've been so slow on these

PR #139, opened 2017-08-22
fix #72 Feeback when trying to log in with a wrong key

### LINKED ISSUE #72: Feeback when trying to log in with a wrong key
**NEON version:** 
alpha 0.0.3c

**I'm submitting a ...**  (check one with "x")
[ ] bug report
[X] feature request

**Current behavior:**
When an invalid key is provided to the login input clicking on the login button will do nothing.

**Expected behavior:**
When an invalid key is provided to the login input a visual feeback is shown by clicking on the login button.
The feedback will inform the key is not valid or could'nt be loaded

**Steps to reproduce:**
Try inserting some random string in the login input.

rjloura:

Fixes issue #72 

Added the text "Private Key is not in proper WIF format." when entry to WIF input text box was not in proper format.  Also fixed the css so that the Login button would show disabled.

Tested with values of "foo", "a", "1", and a property formatted WIF private key generated by the neon-wallet.

N/A
N/A
- [ ] Unit tests written?

### DIFF
--- a/app/components/Login.js
+++ b/app/components/Login.js
@@ -17,6 +17,11 @@ let Login = ({ dispatch, loggedIn, wif }) =>
     <div className="login">
       <div className="logo"><img src={logo} width="60px"/></div>
       <input type="text" placeholder="Enter your private key here (WIF)" onChange={(e) => onWifChange(dispatch, e.target.value)} />
+        {!loggedIn && wif && wif.length > 0 ? (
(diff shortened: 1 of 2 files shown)


rjloura:
neobadkey

Ejhfast:
Thanks for this PR! Unfortunately this pull happened while 0.0.5 was under development, and 0.0.5 supports this feature now, but this was great idea and we appreciate the contribution

Issue #29, opened 2017-08-03
Have ability to save private key onto the wallet


flavoredtaco:
Instead of having to frequently import private key every time you open the wallet, implementing a way to save accounts associated with different private keys would be more efficient.

j4zz:
I’d like this, but I’d like it to be optional (not that taco implied otherwise, just to be explicit).

canesin:
The problem @flavoredtaco is to securely store the keys. The straight forward method is to encrypt a file using a password, but this is orders of magnitude (by common password standards approx 1000000000000000000000000000000000000000000000000000000000000) less safe, maybe if we use a very strong password we can do much better, but then what is the difference of using just the key ? The best second option is if a camera is available we can read a private key QR code (just a easier way to login with private key).. on mobile you also have the option to use fingerprint to encrypt the storage, we can probably with a lot of effort use some kind of face recognition if a camera is available also.

Sure we can be also lazy and do the password thing, most people will be happy - but I would also like to have the option to login with the key, if not I`m not using it.

coranos:
Could this be done the way the neo-gui does it?
They use AES Encryption to decrypt a file on disk, and use the password to unlock the file and read the private key.
A keylogger could get the password and send it to a remote location, where it would be useless without the file on disk.

You could also, if you wanted, add TOTP, like Google Authentication.
https://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm

That would give them a limited window of time to steal your stuff.

canesin:
@coranos you are relying in a 3th party to hold the encryption keys... there must be better ways to do so.

Ejhfast:
@coranos neo-gui does more or less what @canesin says, encrypting private key with user password and saving. TOTP requires unnecessary centralization

I am open to storing encrypted version of private key like neo-gui, but agree we should also keep old way of logging in for users who want to save private key in their own, secure way

coranos:
I think you have what I said backwards. We all know how neo-gui works, I'm asking for the neon wallet to work the same.

TOTP doesn't rely on a third party to hold encryption keys, you can implement it all yourself. It's literally "I sent a number to your phone, tell me what the number is, you have 30 seconds".

Ejhfast:
@coranos i understood what you meant about neo-gui

i didn't understand what you meant by TOTP (now I do), that could make sense but is unrelated to this issue (saving accounts)

dennish007:
[...]

PR #140, opened 2017-08-24
Introduce "Remember my login" feature to automatically load the dashboard

Plasma:
#29 

Introduces a new optional feature, "Remember my login", so that you don't need to re-paste the private key each time opening Neon wallet.

By storing the private key in a settings file, we can refer to it later upon app startup and perform an automatic login.

Manual testing.

Yes. It's worth noting this is a simple implementation, and the storage of the private key is unencrypted.

Yes. We should discuss appropriate storage of the private key. 

I investigated using https://github.com/atom/node-keytar to store the private key in the OS keychain, but there were complication/packaging issues. It also may be more prudent to require a password entry instead of using the keychain.

- [ ] Unit tests written?

image

### DIFF
--- a/app/components/Login.js
+++ b/app/components/Login.js
@@ -1,9 +1,10 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router';
-import { login } from '../modules/account';
+import { login, setRememberLogin, getStoredPrivateKey } from '../modules/account';
 import CreateWallet from './CreateWallet.js'
 import { getWIFFromPrivateKey } from 'neon-js';
+import { withRouter } from 'react-router';
 
 const logo = require('../images/neon-logo2.png');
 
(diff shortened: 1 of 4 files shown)


Plasma:
This PR should be considered a simple proof of concept.

I would like to discuss the proper storage of the private key, as this currently stores it in a plain text settings file.

AshRolls:
Thank you for this PR Plasma. As you say this is a simple proof of concept implementation and we shouldn't merge this to master, plain text storage of private key is not acceptable. 

Additionally there is work already in process for implementing NEP-2 ( https://github.com/neo-project/proposals/pull/2 ) into NEON that this PR may interfere with.

Chat further in slack, and thanks again for this effort, it shows you are a willing and capable developer.

canesin:
@Plasma, I will close this since @Ejhfast is working on a NEP2 implementation. As a rule of thumb we never store anykind of key as plaintext.

Issue #27, opened 2017-08-02
On 'Reload' user is logged out w/ no UI feedback


golivera:
When you are logged into the wallet and you `Reload` the wallet then you get logged out of your account.

The transaction history is wiped and there is no balance.

Possible Improvements
- Redirect back to the login screen
- Show user logout message

![screen shot 2017-08-02 at 3 22 42 pm](https://user-images.githubusercontent.com/1169005/28897640-77f586ec-7796-11e7-88f7-61b9bd6ef539.png)

Ejhfast:
@golivera can you give more steps to replicate? what is "reload"? the refresh button? is this version 0.0.3?

golivera:
MacOSX
V 0.0.3

The `reload` is the native electron reload that you can see under `View > Reload`? On Mac it is `command+r`.

PR #141, opened 2017-08-25
Updated MacOS toolbar

hbibkrim:
#27 

#27 

Removed Reload submenu item - shortcut no longer as a result.

Tested on MacOS Sierra

Removed few other submenu items and added links to CoZ/Github/Slack/Reddit (could remove if unnecessary).

Windows/Linux should be unaffected but as I only have a Mac, I can only test on that.
The most minimalistic a Menu can have is Neon menu item with nothing inside it. I've left the basic commands and the link to Dev tools. In future PR, can display Toggle Dev Tools only when running in debug mode?

- [ ] Unit tests written?


marksuman:
Would it be better to get the Reload functionality to work instead of removing the option from the menu?

hbibkrim:
I don't think the reload functionality brings anything to the application that logging out doesn't. It looked like boilerplate code left over and not an intentional feature

danielfm123:
You can use that neo web wallet https://www.myneowallet.io

canesin:
@CityOfZion/council someone with OSX to test ?

Ejhfast:
thanks @hbibkrim!

PR #144, opened 2017-08-27
Fix link to NEO sub-Reddit

weppos:
I admit this is a very trivial change, very low priority. This PR fixes the link to the Reddits sub-channel that was improperly linking to the Slack account.


canesin:
tks

PR #155, opened 2017-09-04
Feature/refresh helper text

robinvanb:
No existing issue.

When I first used Neon it wasn't quite clear to me what the two arrows in the middle meant, and there was no tooltip.

Added a tooltip to the refresh balance button.

Check how you used tooltips elsewhere, then I ran the application and verified it worked.

Added *.iml (IntelliJ config file) to .gitignore

No.

- [ ] Unit tests written?
No/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ release-builds/*
 app/dist/*
 app/dist
 .DS_Store
+*.iml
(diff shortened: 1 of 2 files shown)


Ejhfast:
awesome, thanks!

PR #164, opened 2017-09-07
Fixed issue with treating key read as success when it fails

mhuggins:
None that I know of.

If keys fails to load from storage, it is still treated as a successful read and may result in saving of bad data, for example.

Check for error result and display a user-friendly error message.

TODO: I haven't yet.  I will add tests/smoke tests if this otherwise looks good.

No.

No.

- [ ] Unit tests written?
Not yet.


Ejhfast:
thanks! will test and look at this in a bit

shawnmclean:
Closing due to inactivity. Please open a new pull request if you decide to continue on this enhancement. :)

PR #172, opened 2017-09-08
Initial suggested issue template

Satoshinaire:
None.

Users requesting general software support on the developer GitHub

Creating an issue template to prompt users seeking support to go to Slack

Not possible

No

This is just a suggestion, feel free to disregard and create something else that suits

- [ ] Unit tests written?


Ejhfast:
great idea!

PR #177, opened 2017-09-09
Use password type for the private key field for security reason

marcdelalonde:
Some user feedbacks

Security issues

Change the html field type

- [ ] Unit tests written?

### DIFF
--- a/app/components/LoginPrivateKey.js
+++ b/app/components/LoginPrivateKey.js
@@ -41,7 +41,7 @@ let LoginPrivateKey = ({ dispatch, loggedIn, wif, history }) =>
     <div className="login">
       <div className="loginForm">
         <div className="logo"><img src={logo} width="60px"/></div>
-        <input type="text" placeholder="Enter your private key here (WIF)" ref={(node) => wif = node}/>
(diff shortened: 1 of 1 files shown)


darfraider:
You need to make the changes in both login forms right?  Encrypted Key login and Private Key login?  I think that would be a good idea.

Satoshinaire:
I think the main benefit of this change is psychological more than anything. But that's not nothing.

Agree too that if you're going to do it, you should do it everywhere that makes sense.

darfraider:
It’s interesting that this PR is not merged into the master but when I cloned the master and built it, when I ran it, the encrypted key login ad well as private key login enters fields are shadowed with *. Not sure where this is handled because the html code inserted in the js methods are still input type of “text” (instead of “psssword” as proposed by this PR).

Ejhfast:
This is on the roadmap, but with support for a toggle -- some people are copying from paper wallets and want to see the key as they type in.

If you add this feature to all non-password inputs turned off/on by a toggle, would be happy to accept!

shawnmclean:
Solved in #69

PR #181, opened 2017-09-10
feat: add ability to load JSON wallet encrypted private keys

Shou:
We add the ability to load a JSON encoded wallet based on the format
used in the original 'Antshares' online wallet.

None AFAIK. However, I've wanted this myself, and seen it requested elsewhere.

Allows users to load their JSON wallet encrypted private key.

I added a "Select wallet" button to the "Login using an encrypted key" section that opens a file dialog.

By successfully decrypting my own JSON wallet key.

Usage of Flexbox in the login sections, which should be an improvement. Otherwise, other 'changes' are almost just additions.

- [ ] Unit tests written?

### DIFF
--- a/app/components/DisplayWalletKeys.js
+++ b/app/components/DisplayWalletKeys.js
@@ -13,7 +13,10 @@ let key_name;
 
 const saveKey = (dispatch, encWifValue) => {
   storage.get('keys', (error, data) => {
-    data[key_name.value] = encWifValue
+    data[key_name.value] = {
+      type: 'neon',
+      privateKey: encWifValue
+    };
     dispatch(sendEvent(true, "Saved key as "+key_name.value))
     storage.set('keys', data);
     setTimeout(() => dispatch(clearTransactionEvent()), 5000);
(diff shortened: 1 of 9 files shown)


Ejhfast:
Thanks! Will take a look at this soon

Ejhfast:
@Shou can you add this the setting page (where Neon loads its own encrypted key files)? or were you intending this as something more like a one-time-login using a file feature? we think the feature is really valuable, but would be good to be consistent

either way, could you get in touch with me on the slack (unignorant) -- CoZ would like to reward you some Neo

Shou:
Sure, that sounds like a better idea! I've started working on that now. Although the encryption model is slightly different with the Neo Wallet, so it needs to store some extra data (master key and initialization vector), or to decrypt the wallet and convert it which requires password input alongside the import button. Maybe we could change the current wallet storage format to be more flexible? Like a list of objects, a version field, type fields (for different wallet formats), etc. This is better addressed more thoroughly in separate PR though.

Joined the Slack, I'm `/team/U78GQH3CK` under the same nickname as here.

shawnmclean:
Hey @Shou, do you have any intentions of fixing the conflicts in this pull request?

Shou:
@shawnmclean sorry for the response delay here! I've been rather busy but I will complete the rebasing.

dvdschwrtz:
@Shou there have been a lot of major changes to the app since this has been opened. Let me know if you would like help in terms of updating and merging this feature.

dvdschwrtz:
@Shou I am going to close this PR tomorrow if we don't hear back from you. We are also upgrading to NEP-6 which is a unified format for storing wallets soon

PR #192, opened 2017-09-24
Add missing return statement to validateForm()

ixje:
@shawnmclean I can't comment to your question below because of repository restrictions, so I'm editing this message to answer you.
I might later on, my order of priority would be this: 

- add tests to my own mobile wallet (I'm using neon-wallet as reference for the SDK)
- add tests to neon-js SDK
- add tests here.


shawnmclean:
Any chance you can add some tests around this? If not no worries, the fix is still good.

shawnmclean:
@Ejhfast Do you know why @ixje cannot comment or what restriction is happening? I thought anyone could comment on PRs. The conversation isn't locked either.

ixje:
@shawnmclean something seems to have changed, because now I can. Maybe because I just registered on de neo developers page? I had not seen such a restriction before elsewhere, so I'm just guessing here

shawnmclean:
Merging this in now, feel free to create another one with your tests :)

PR #205, opened 2017-09-28
mocking neon-js calls to remove async issues

dvdschwrtz:

I did not mock the async calls to neon-js when I originally wrote these tests, which is why I had async issues and needed try/ catch blocks and long setTimeouts. Now the 'neon-js' calls are mocked and it is working correctly.

Testing the neon wallet

By adding in mock functionality for 'neon-js' calls

running 'npm test'

there was one place I left a TODO comment where I put the same code for my axios mock adapter in 2 files. I was unable to make this DRY with my initial attempt and feel that this is an important PR to improve the current tests, so I created this PR and will come back to that minor issue later.

- [ ] Unit tests written?

### DIFF
--- a/__tests__/components/WalletInfo.test.js
+++ b/__tests__/components/WalletInfo.test.js
@@ -7,15 +7,27 @@ import { SEND_TRANSACTION, CLEAR_TRANSACTION } from '../../app/modules/transacti
 import { SET_HEIGHT } from '../../app/modules/metadata';
 import { SET_CLAIM } from '../../app/modules/claim';
 import WalletInfo from '../../app/components/WalletInfo';
+
+// TODO research how to move the axios mock code which is repeated in NetworkSwitch to a helper or config file
 import axios from 'axios';
 import MockAdapter from 'axios-mock-adapter';
 import { clipboard } from 'electron';
+import { version } from '../../package.json';
+
+const axiosMock = new MockAdapter(axios);
+axiosMock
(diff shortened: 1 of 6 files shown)


__tests__/components/WalletInfo.test.js
```diff
@@ -137,16 +134,12 @@ describe('WalletInfo', () => {
       SET_CLAIM
     ];
     setTimeout(() => {
```
shawnmclean: Do we still need the setTimeout here?

__tests__/components/WalletInfo.test.js
```diff
@@ -109,26 +117,15 @@ describe('WalletInfo', () => {
     ];
     deepWrapper.find('.refreshBalance').simulate('click');
     setTimeout(() => {
```
shawnmclean: Same here

__tests__/components/WalletInfo.test.js
```diff
@@ -86,13 +98,9 @@ describe('WalletInfo', () => {
     expect(clipboard.writeText.mock.calls.length).toBe(0);
     deepWrapper.find('.copyKey').simulate('click');
     setTimeout(() => {
```
shawnmclean: Here too.

dvdschwrtz:
@shawnmclean I took it out of the first one that checked copy to clipboard was called. I do need the setTimeouts for the other 2 places because the store.getActions() method needs that extra cycle in the event loop in order to return the correct results.

Issue #216, opened 2017-10-06
Globally refactor code and divide into more coherent system of components and containers


Ejhfast:
If someone wants to tackle this, would be very valuable. This codebase is becoming a bit ugly

evgenyboxer:
I'll do it!

PR #218, opened 2017-10-07
First round of linting

evgenyboxer:
This is part of the refactoring - https://github.com/CityOfZion/neon-wallet/issues/216
Since I'll be moving files around, I might as well lint them too. This PR is a good start (airbnb styling)

This PR solves many issues relating to inconsistent JS style across the repo.

Installed the latest eslint packages and ran the --fix.

Currently, some tests are broken, but this is unrelated to this PR. Also, I checked that the wallet works, and it does.

We need to merge this one quickly, since we may encounter conflicts.


evgenyboxer:
After speaking with @Ejhfast I'll adjust the linter to use the same we have on neon-js.

PR #220, opened 2017-10-09
Linting (StandardJS)

evgenyboxer:
This is part of the refactoring - #216
Since I'll be moving files around, I might as well lint them too.

This PR solves many issues relating to inconsistent JS style across the repo.

Used StandardJS + StandardJS for React + Fixed ALL lint errors (TONS) + Added PropTypes + Removed Deadcode + camelCase variables + Deconstructing props/state + more stuff.

Currently, some tests are broken, but this is unrelated to this PR. Also, I checked that the wallet works, and it does.


shawnmclean:
Can you run `yarn install` to get the yarn.lock updated. It seems its causing some issues with caching based on the checksum of that file.

PR #233, opened 2017-10-10
mapDispatchToProps refactor example

dvdschwrtz:

Potential refactor - I like mapDispatchToProps better than the current style because you do not have to pass dispatch around or explicitly call it. 

Code refactor

by testing the app

- [ ] Unit tests written?

### DIFF
--- a/app/components/Settings.js
+++ b/app/components/Settings.js
@@ -9,7 +9,7 @@ import { forEach, map } from 'lodash'
 import fs from 'fs'
 import storage from 'electron-json-storage'
 
-const {dialog} = require('electron').remote
+const { dialog } = require('electron').remote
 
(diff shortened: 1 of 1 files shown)


evgenyboxer:
I like this! Good improvement.
A step further in my view would be to move all the actions into their module/duck file. Keep all that logic away from the component. Using something like `redux-thunk` can help with that. It can fire actions (using dispatch) and is awesome for the async stuff you want.

evgenyboxer:
Assuming this is your duck - 

`export const someAction = () => (dispatch: DispatchType) => {
    // you can also get stuff from state if you want
    dispatch(someAction)
   ...
   // more dispatching..
}`

PR #234, opened 2017-10-11
Empty Field Checking and User Notification Messages

phetter:
The wallet doesn't check for empty input fields and gives the user no notification message if a field is required or empty. This PR addresses several of those cases, but not necessarily all of them.

### DIFF
--- a/app/components/EncryptKey.js
+++ b/app/components/EncryptKey.js
@@ -26,24 +26,29 @@ const generateNewWallet = (dispatch) => {
     setTimeout(() => dispatch(clearTransactionEvent()), 5000)
     return
   }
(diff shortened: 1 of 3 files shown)


shawnmclean:
@dvdschwrtz could you review this and see if the setTimeout stuff still applies here under your changes?

dvdschwrtz:
@shawnmclean I have enabled mock timers for some of the tests so far. The setTimeouts in the components won't be affected. In the near future, once I pull the async logic from the components into the modules folder as I discussed with @evgenyboxer, the setTimeouts might be removed. I am not really sure yet. In either case, it should be fine to merge this in.

PR #237, opened 2017-10-12
UI/UX - Input Form Field Submission on Enter Keypress

phetter:
The workflow of "login using a saved wallet" feels a bit off in terms of user experience so this PR is meant to fix that.

The "select a wallet" dropdown has been moved above the passphrase input box and now the login button can be triggered by pressing enter on the passphrase input box. 

So now the workflow is:

1. Select a wallet
2. Type in passphrase
3. Press enter or click Login button to submit form

Currently, this change only affects LoginLocalStorage.js ("login using a saved wallet") as I wanted to see how the feature would be received. If the feature is appreciated, I will work to make this change consistent for all input form fields.


hal0x2328:
Nice, this was something that bugged me too, I was considering working on this exact fix.

shawnmclean:
Nice, can you add a test for this component, which simulates the keypress and that the right actions were dispatched?

app/components/LoginLocalStorage.js
```diff
@@ -60,6 +60,12 @@ let LoginLocalStorage = class LoginLocalStorage extends Component {
     }))
   }
 
+  handleKeyPress = (event) => {
+    if (event.key === 'Enter') {
```
evgenyboxer: Please use constants.
We now have a `constants.js` under /core folder.
Can be something like:
```
export const KEYS = {
  ENTER: 'Enter',
}
```

app/components/LoginPrivateKey.js
```diff
@@ -35,6 +35,12 @@ class LoginPrivateKey extends Component {
     onWifChange(dispatch, verifyPrivateKey, history, wif)
   }
 
+  handleKeyPress = (event) => {
+    if (event.key === 'Enter') {
```
evgenyboxer: Same comment like above ^

phetter:
I'm going to close this. The attempt to merge with the new changes basically nullifies all the work I did on testing and form submission last week.

phetter:
Still revising tests...

phetter:
Sorry for all the commits. The refactor tripped me up but I wanted to resolve all this while it was fresh and since evgeny was nice enough to offer conflict resolution suggestions. I dropped the login test for the enter keypress for now because the refactor was tripping me up. I will continue the unti test work, including this one, under another issue/pr.

evgenyboxer:
I'll send another conflict resolution soon.

PR #246, opened 2017-10-14
initital refactor of component and container architecture

dvdschwrtz:

Start to refactor the container and component architecture and move more logic into the actions/ modules files.

Refactoring the neon wallet

creating a new container folder structure with an index.js connect wrapper and its corresponding .jsx file

I QA tested it a bit. further review and QA testing of the login with private key and then sending from the dashboard function would be great!

- [ ] Unit tests written?


dvdschwrtz:
@evgenyboxer please review and let me know what you think of the way I implemented the changes. Also, I added flow to some of the modules files that I modified when pulling logic out of the containers. I don't know if this is correct or not, as the modules files do not have flow incorporated into them already.

evgenyboxer:
Looks really good! Good job! +1 for Flow in modules too.

The only thing I would change is the usage of Wrapper (DashboardWrapper).
I think it’s okay to use it without Wrapper and do a default export connect(...) in the containers without a dedicated const for it.

What do you think?

dvdschwrtz:
Thanks! Appreciate the review. I made the change. should be good to go now.

evgenyboxer:
Cheers.
By the way, I believe that you don’t need the ‘/index’ in the end of the import.
‘containers/Dashboard’ should lead to the index.js file without /index

PR #256, opened 2017-10-19
Refactor several components

dvdschwrtz:

refactoring the neon wallet to switch container and component architecture and move logic into modules

refactor neon wallet

by refactoring the TokenSale, NetworkSwitch, TransactionHistory, and WalletInfo components to switch container and component architecture and move logic into modules

manual QA testing works but still need to fix some of the tests 

no

- [ ] Unit tests written?


dvdschwrtz:
This PR is bigger than I originally intended it to be, but a lot of components were using similar methods, such as initiateGetBalance for example, so I needed to refactor multiple components at once in order to get this to work

PR #251, opened 2017-10-18
Fix flow and lint errors

evgenyboxer:
**What current issue(s) from Trello/Github does this address?**
Failed Flow/Lint errors.

**What problem does this PR solve?**
Fixed Flow/Lint errors.

**How did you solve this problem?**
yarn flow && yarn lint

**How did you make sure your solution works?**
No failed errors found!

PR #274, opened 2017-10-25
Adding flow check to CircleCI configuration

arithmetric:
I mentioned this idea in #251.

Currently the flow checks are not performed as part of the CircleCI tasks, so a flow error does not trigger CI error reporting to GitHub.

Added the `yarn flow` command to the build job. I added it there because I thought it made sense to couple with `yarn lint`.

The CircleCI report for this PR will demonstrate it.

no

no

- [n/a] Unit tests written?


arithmetric:
`yarn flow` was run as part of this CircleCI build: https://circleci.com/gh/CityOfZion/neon-wallet/334

Issue #230, opened 2017-10-10
Basic Integration Test using Spectron


shawnmclean:
We need a basic workflow testing using some UI testing tool for electron such as Spectron and executed as a step during the CI process.

Reason: We want to ensure that the app will still load with basic functionalities when there are code changes.

Suggested Workflow: `Login with known Wif on testnet` > `View Balance and Transactions`

thuansb:
Hi @shawnmclean, I will take this task. What I need to do is just implement the test and do the pull request later right?

PR #280, opened 2017-10-28
Run e2e test on CircleCI

thuansb:
#230 
Config to run integration test on CI

- [ ] Unit tests written?


shawnmclean:
Nice work!

PR #291, opened 2017-10-31
 Show wallet gas and total value, show neo/gas prices

slipo:
None

Currently the app just shows the NEO wallet value and excludes GAS.

Now the app will show the quoted NEO/GAS prices and the NEO/GAS wallet values as four separate data items. In addition it shows a sum total wallet value.

There are automated tests. I also manually checked the app to see the values.

It's switching from bittrex API to coinmarketcap because the latter has the GAS USD market.

- [x] Unit tests written?
yes!


shawnmclean:
usd gas

I'm approving the code, the design, I'm not sure if anyone else wants to give any feedback on that.

evgenyboxer:
PR looks good.
I’m terms of the design, I think it’s better to drop the value of a single Neo/Gas.

Under each asset just write the total, without the wording “NEO/GAS Wallet”, as it is inferred from the context.

I agree that having a total for both is a good idea. In terms of the wording, maybe instead of “Wallet”, just “Total”.

slipo:
Is this what we're looking for? If so I'll push it out.

designchangesneon

slipo:
OK, those changes were looking pretty good to me so I pushed them out. Let me know if there's anything else to do.

shawnmclean:
Looks much cleaner!

PR #293, opened 2017-11-02
ledger update

dvdschwrtz:

Adding ledger support for Neon Wallet

Updates ledger code to refactor javascript and integrate with neon-js. improved user feedback during ledger usage

added Methods to neon-js and refactored wallet code

Tested with ledger

no. I will add tests for the ledger code soon when we revisit testing for the wallet after the refactor.

no

- [ ] Unit tests written?


evgenyboxer:
Looks good!

PR #305, opened 2017-11-07
Hides the token sale page for 0.0.7 release

evgenyboxer:
Token sale is currently broken

Hides the button (does not remove the actual code)


henryfour:
Hi, evgenyboxer:
  Can people still use 0.0.6 version to participate in token sales on NEO? Will there be any secrity problem?
  Looking forward to your reply, Thank you very much!

PR #331, opened 2017-11-17
Display current NEO and GAS prices

slipo:
None

Would be nice to see the current NEO and GAS prices in the UI.

Shown in display now.
display prices

Tests + view it in the app

No

No, pretty simple. 

- [x] Unit tests written?
Yes!


app/containers/PriceDisplay/index.jsx
```diff
@@ -0,0 +1,11 @@
+// @flow
```
evgenyboxer: Could you please rename to index.js (we use .jsx for React components)

app/containers/PriceDisplay/PriceDisplay.jsx
```diff
@@ -0,0 +1,40 @@
...
+import styles from './PriceDisplay.scss'
+import classNames from 'classnames'
+
+type Props = {
+  neoPrice: number,
+  gasPrice: number
+}
+
+export default class PriceDisplay extends Component<Props> {
+  render () {
```
evgenyboxer: I think this could be better written as a functional component - 
`const PriceDisplay = ({ ... }: Props) => { ... return (...) }`

evgenyboxer: Also, could you move the PriceDisplay to the components section? Then from the dashboard container pass the props it needs.

evgenyboxer:
Looks good! Just a few comments.

app/containers/PriceDisplay/index.jsx
evgenyboxer: Actually this might not be relevant, as this will not be a container.

slipo:
Sure thing. All done as requested, I believe. I kept the index.js but now it just does a pass-through load of the component like the others that are similar. Simplified the tests as well to just do a snapshot check. Let me know if there's anything else.

Issue #265, opened 2017-10-20
Give warning when API is down or wallet lacks connectivity


Ejhfast:

evgenyboxer:
On it

skeller88:
In the slack support channel, I read that when the wallet is unexpectedly showing zero balances, it's due to an API connectivity issue. So adding this feature would have the added bonus of cutting down on questions from people about where their NEO funds are.

Satoshinaire:
Yeah, absolutely. I think this actually ties in with #9 as well regarding persistent state (perhaps there should be an issue specifically about persistent state that both of these issues relate to).

It'd be good to keep the balance cached locally so that the 0 doesn't freak people out, even when updated balances can't be obtained due to API outage or network connectivity.

PR #346, opened 2017-11-22
Fix for: Give warning when API is down or wallet lacks connectivity #265

slipo:
Github #265

Shows an error if the app can't get wallet balance

On init, it's using the standard refresh method which already has the correct error checking

Disconnect internet.

No

No

- [x] Unit tests written?
Yes!


slipo:
Forgot to attach the screenshot
screen shot 2017-11-22 at 1 27 52 pm

PR #350, opened 2017-11-23
Better formatting

slipo:
No

2 things:
* Numbers that are 1000+ now have commas, ie 1,123,123.00 instead of 1123123.00
* The total wallet value (neo + gas) would be 0 if gas value is 0. This is common if the user has never claimed. Reworked the logic there.

Using numerals package

Wallet screen + tests

No

I needed to install the mime package even tho I'm not adding anything that requires it.

- [x] Unit tests written?


dvdschwrtz:
This looks good. I think Gas is 8 decimals by the way. Not sure how we ended up with 7 in the wallet. I will merge this and we can fix that later

PR #352, opened 2017-11-23
Enter in textfields submits

slipo:
All login password fields submit the form when you hit enter

### DIFF
--- a/app/components/PasswordField/PasswordField.jsx
+++ b/app/components/PasswordField/PasswordField.jsx
@@ -1,9 +1,15 @@
 // @flow
 import React, { Component } from 'react'
 
+import { omit } from 'lodash'
+
 import FaEye from 'react-icons/lib/fa/eye'
 import FaEyeSlash from 'react-icons/lib/fa/eye-slash'
 import passwordFieldStyles from './PasswordField.scss'
+
+type Props = {
+  onEnterKey: Function
+}
 
 type State = {
   showKey: boolean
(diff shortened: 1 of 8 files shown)


app/components/PasswordField/PasswordField.jsx
```diff
@@ -14,6 +20,19 @@ class PasswordField extends Component<Props, State> {
     showKey: false
   }
 
+  constructor () {
+    super()
+    this._handleKeyPress = this._handleKeyPress.bind(this)
+  }
+
+  _handleKeyPress (target) {
+    if (target.charCode === 13) {
```
evgenyboxer: Please use a constant, like `ENTER_KEY` (defined in constants)

app/components/PasswordField/PasswordField.jsx
```diff
@@ -14,6 +20,19 @@ class PasswordField extends Component<Props, State> {
     showKey: false
   }
 
+  constructor () {
+    super()
+    this._handleKeyPress = this._handleKeyPress.bind(this)
```
evgenyboxer: You can use arrow functions instead of binding.

So just have a `handleKeyPress = (target) => { }` (without the constructor)

evgenyboxer:
I was wondering, maybe it should be a higher order component?

`withOnEnter(component)`

and then each component will have to implement an "onEnterKey" method, which will trigger the form submission (instead of passing this for every field)

In the higher order component you register to listen the key events, and once you detect the ENTER key, just fire the registered method on the wrapped component.

app/components/PasswordField/PasswordField.jsx
slipo: hey, good tip! thanks

slipo:
I'm not sure the HOC is the best approach. It definitely works for the current design with single pages. But once the UI gets more complicated and perhaps there are overlays or multiple forms shown at once, then you have to start testing if your form is in focus somehow. In any case I'll code it up quickly as an alternative PR and we can decide from there.

slipo:
Closing this out in favor of this much simpler approach #354

PR #375, opened 2017-11-30
fix remaining test issues

dvdschwrtz:

- [ ] Unit tests written?


dvdschwrtz:
Looks like the e2e inegration tests are also failing. Let's merge this and fix that on another PR

Issue #336, opened 2017-11-18
Can no longer copy/paste "send to" address.


summertooth:
I used to be able to copy my send to address from an exchange and paste in Neon, can no longer do this.  Seems like this should be enabled as typing these addresses will lead to many errors.  Thank you for your time.

comountainclimber:
What version are you running? Neon => About Neon

rockaBil-E:
i upgraded v006 two weeks ago, and now7, and found this 'paste' bug in both (on my machine).  I re-loaded 5 (pastes fine) and will wait for an resolution in 8.

summertooth:
v7, newest version, just installed yesterday.  @comountainclimber

stevesbrain:
@rockaBil-E @summertooth Are you both trying to right click and user copy + paste via right click menu's? Try Ctrl+C shortcut for copy, and Ctrl+V shortcut for paste.

chris40taylor:
Hi, I had the same problem with the cut and paste but followed stevesbrain above and it worked.
Will this issue be rectified?

bleach182:
Okay so has this been solved? I am new to this and don't really understand what @dvdschwrtz  commits are about. I when to the #336 and #376 but do not understand what I am supposed to do with that code..

Satoshinaire:
@bleach182 You don't need to do anything, but the issue should be resolved in the next release.

PR #376, opened 2017-11-30
Add Cut/Copy/Paste context menus on input fields (#336)

jmsaulnier:
#336

Right Click on input doesn't do anything. You should see a menu with cut / copy / past / select all options

Add Electron menu input config

Tests

### DIFF
--- a/main.js
+++ b/main.js
@@ -81,10 +81,7 @@ app.on('ready', () => {
     }
 
     const inputMenu = Menu.buildFromTemplate([
-      {label : 'Cut', accelerator : 'CmdOrCtrl+X', click() { mainWindow.webContents.cut(); }},
-      {label : 'Copy', accelerator : 'CmdOrCtrl+C', click() { mainWindow.webContents.copy(); }},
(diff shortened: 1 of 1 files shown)


dvdschwrtz:
I am not seeing any difference to the copy and paste functionality when I use these changes. What does it mean that it adds a context menu? @evgenyboxer can you also take a look at this

jmsaulnier:
@dvdschwrtz I updated my PR comment, a right click on input elements should display a menu with cut / copy / past / select all options

main.js
```diff
@@ -80,6 +80,17 @@ app.on('ready', () => {
       Menu.setApplicationMenu(menu)
     }
 
+    const inputMenu = Menu.buildFromTemplate([
```
dvdschwrtz: per our slack discussion, please change to only allow for 'paste'. thanks and look forward to working with you on the wallet :)

jmsaulnier: @dvdschwrtz done 👌

Issue #363, opened 2017-11-28
Support for other currencies


slipo:
This would be a currency selection field on the settings page. Then the frontend would show that currency instead of USD. I'll take this on if desired.

PR #378, opened 2017-12-01
Support for non-USD Currencies

slipo:
#363 

Now you can set your currency to anything supported by Coin Market Cap

New option on settings page

Tests + change currency on settings page, view wallet.

I copied and pasted an api method from neon-js. When we upgrade to the new version we can use the theirs.

Changed the design slightly

- [x] Unit tests written?


app/modules/price.js
```diff
@@ -0,0 +1,121 @@
+// @flow
+import axios from 'axios'
+import asyncWrap from '../core/asyncHelper'
+import { DEFAULT_CURRENCY_CODE } from '../core/constants'
+
+/* @todo remove below when we upgrade neon-js and can use their getPrice function */
+const CURRENCY = ['aud', 'brl', 'cad', 'chf', 'clp', 'cny', 'czk', 'dkk', 'eur', 'gbp', 'hkd', 'huf', 'idr', 'ils', 'inr', 'jpy', 'krw', 'mxn', 'myr', 'nok', 'nzd', 'php', 'pkr', 'pln', 'rub', 'sek', 'sgd', 'thb', 'try', 'twd', 'zar']
```
dvdschwrtz: maybe use Object.keys on the currency from the core constants file instead of creating this list

PR #379, opened 2017-12-01
Don't run coveralls when pushing

slipo:
None

Currently when you push to the repo it tries to run coveralls coverage test and gives this error:
```
neon-wallet/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

Don't run try to trigger coveralls locally

Pushed this branch, didn't have to use '--no-verify'

No

For some more information:
https://github.com/nickmerwin/node-coveralls/issues/166

- [ ] Unit tests written?


slipo:
Just pushed a nicer version which allows us to continue using `yarn test` locally... For this to work, somebody needs to change the Circle CI build configurations to use `yarn test-ci` instead of `yarn test`

shawnmclean:
@slipo you can also make this change in the `/.circleci/config.yml` file.

slipo:
ahh cool, didn't realize. done.

Issue #271, opened 2017-10-24
Warn user when address being sent to has no transaction history


Ejhfast:
ARK addresses apparently use exactly the same naming conventions as NEO addresses, so even with all the other checks, this would still be helpful for preventing users from hurting themselves

evgenyboxer:
Should this be first implemented in Neon-JS? How do you check for this right now?

Ejhfast:
Not neon-js, as this is more a UI feature. Just look up transaction history
of address being sent to, if it is empty, display warning. neon-js will
allow you to do that API call to lookup

evgenyboxer:
Right okay!
In terms of the UX of this - How do you imagine the warning to look like? Do you think it should be a sticky message at the top (one you hit "send"), or it should be displayed next to the "confirm pane" button we have?

Also, unrelated - but I would like to rework the "confirmation button" with a small confirmation modal. (Once you hit send and its valid -> I open the confirmation modal -> Repeating all the send info + the warning mentioned above for empty history)

Ejhfast:
Yeah, maybe if we rework the confirmation we can build the warning into
that? I like the modal idea

evgenyboxer:
Yep. sounds great! I'll work on this.

dvdschwrtz:
closing this issue as it has been merged to dev

PR #397, opened 2017-12-05
Warn user when address being sent to has no transaction history

slipo:
https://github.com/CityOfZion/neon-wallet/issues/271

Check NeonDB tx history for recipient address transactions. If there are none, show warning

Check the UI sending to fresh throwaway address.

screen shot 2017-12-05 at 9 53 26 am

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/SendModal/ConfirmDisplay.jsx
+++ b/app/components/Modals/SendModal/ConfirmDisplay.jsx
@@ -6,6 +6,7 @@ import Button from '../../Button'
 import Loader from '../../Loader'
 import { formatBalance } from '../../../core/formatters'
 import { openExplorerAddress } from '../../../core/explorer'
+import asyncWrap from '../../../core/asyncHelper'
 
 import styles from './SendModal.scss'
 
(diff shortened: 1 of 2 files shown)


app/components/Modals/SendModal/ConfirmDisplay.jsx
```diff
@@ -17,24 +19,60 @@ type Props = {
...
+}
+
+class ConfirmDisplay extends React.Component<Props, State> {
+  state = {
+    addressChecked: false,
+    addressCheckedMessage: ''
+  }
+
+  checkTransactionHistory = (net, address) => {
+    api.neonDB.getTransactionHistory(net, address).then((transactions) => {
```
evgenyboxer: I believe you can do something like: (taken from transactions.js)

```
  let message = ''

  const [err, transactions] = await asyncWrap(api.neonDB.getTransactionHistory(net, address))
  if (err) {
  	message = 'Warning: recipient address has no activity in its transaction history. Please be sure the address is correct before sending.'
  	}
  } else if (!transactions || !transactions.length) {
      message = 'Warning: recipient address has no activity in its transaction history. Please be sure the address is correct before sending.'  	
  }

  this.setState({
  	addressChecked: true,
  	addressCheckedMessage: message
  })
```

This way you setState only once + use the async/await.
Also - `(net: NetworkType, address: string)`

PR #400, opened 2017-12-06
Improve Ledger Login Process

dvdschwrtz:

I added a setInterval and refactored the login process for ledger so it automatically detects if you plug-in or unplug your ledger from the login page

- [ ] Unit tests written?

### DIFF
--- a/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
+++ b/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
@@ -27,15 +27,15 @@ export default class LoginLedgerNanoS extends Component<Props, State> {
 
   componentDidMount () {
     const { ledgerNanoSGetInfoAsync } = this.props
-    const intervalId  = setInterval(async () => {
+    const intervalId = setInterval(async () => {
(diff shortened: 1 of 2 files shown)


app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -16,14 +16,32 @@ type Props = {
   publicKey: string,
   history: Object
 }
+type State = {
+  intervalId: number
+}
+
+export default class LoginLedgerNanoS extends Component<Props, State> {
+  state = {
+    intervalId: ''
```
evgenyboxer: Any reason this is on the state? (vs just a `let` outside of the class)

dvdschwrtz: @evgenyboxer I read that you need to maintain the interval id in state with react if you want to make sure that the interval is cleared when the component unmounts

dvdschwrtz: although I think it might work either way. I believe this is best practice

PR #405, opened 2017-12-07
refactor send asset function code

dvdschwrtz:

- [ ] Unit tests written?

### DIFF
--- a/app/modules/transactions.js
+++ b/app/modules/transactions.js
@@ -67,24 +67,18 @@ export const sendTransaction = (sendAddress: string, sendAmount: string, symbol:
 
     const isHardwareSend = !!publicKey
 
-    // TODO: Consolidate this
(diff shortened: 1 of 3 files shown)


app/modules/transactions.js
```diff
@@ -67,25 +67,19 @@ export const sendTransaction = (sendAddress: string, sendAmount: string, symbol:
 
     const isHardwareSend = !!publicKey
```
evgenyboxer: Is there a better check for this?

evgenyboxer: Maybe we can define the type of the login? Feels like this condition is only known to us..

dvdschwrtz: yeah, that makes sense. I will incorporate those changes for the ledger login page as well

PR #411, opened 2017-12-11
Upgrade to neon-js 2.2.1

slipo:
Fixes the issue with prices getting rounded to integers:
https://github.com/CityOfZion/neon-js/commit/a8d8bb18f1f6a5fbf395c92ee1a35a690906a0b5

Adjustment made for the following commit which solves the isNaN mystery:
https://github.com/CityOfZion/neon-js/commit/4504367ea9375e9e4bdad81655f42b538f706ffa

### DIFF
--- a/app/components/Modals/TokenInfoModal/TokenInfoModal.jsx
+++ b/app/components/Modals/TokenInfoModal/TokenInfoModal.jsx
@@ -33,15 +33,15 @@ const TokenInfoModal = ({ hideModal, token }: Props) => (
           </tr>
           <tr>
             <td>Name:</td>
-            <td>{token.info.name}</td>
+            <td>{token.name}</td>
           </tr>
           <tr>
             <td>Total Supply:</td>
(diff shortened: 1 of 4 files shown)


evgenyboxer:
We probably need to use `getToken` (https://github.com/CityOfZion/neon-js/blob/master/src/api/nep5.js#L66) instead of getBalance/getTokenInfo because this call would give you both of the things we need (info+balance), instead of doing 2 calls.

Also, it seems that the `getToken` will return null when the balance is incorrect? (https://github.com/CityOfZion/neon-js/blob/master/src/api/nep5.js#L86)

slipo:
It seems to me that `getToken` is not working how we need it to. I've created an issue here:
https://github.com/CityOfZion/neon-js/issues/104

So we can keep this issue open till that's resolved or merge and handle this todo separately.

slipo:
Looks like the fix will get into master soon.

https://github.com/CityOfZion/neon-js/pull/106

I can update the code later today and point at the hash with the fix.

PR #413, opened 2017-12-12
notify user ledger support for NEP5 is not yet available and disable …

dvdschwrtz:
…functionality

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Dashboard/index.js
+++ b/app/containers/Dashboard/index.js
@@ -2,7 +2,7 @@
 import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
 
-import { logout, getAddress } from '../../modules/account'
+import { logout, getAddress, getHardwareLogin } from '../../modules/account'
 import { getBlockExplorer, getNetwork } from '../../modules/metadata'
 import { getNotifications, showErrorNotification } from '../../modules/notifications'
 import { getNEO, getGAS, getTokens, getIsLoaded, loadWalletData } from '../../modules/wallet'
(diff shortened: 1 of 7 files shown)


app/modules/transactions.js
```diff
@@ -72,9 +72,11 @@ export const sendTransaction = (sendAddress: string, sendAmount: string, symbol:
     let sendAssetFn
     if (symbol === ASSETS.NEO || symbol === ASSETS.GAS) {
       sendAssetFn = () => api.neonDB.doSendAsset(net, sendAddress, publicKeyOrWif, sendAsset, asyncSigningFunction)
-    } else {
+    } else if (!isHardwareSend) {
       const scriptHash = getScriptHashForNetwork(net, symbol)
       sendAssetFn = () => api.nep5.doTransferToken(net, scriptHash, publicKeyOrWif, sendAddress, adjustDecimalAmountForTokenTransfer(parsedSendAmount), 0, asyncSigningFunction)
+    } else {
+      return rejectTransaction('ledger support is not yet ready for sending NEP5 tokens')
```
evgenyboxer: Looks good! Maybe capitalize `ledger` (like the previous message)

app/containers/Dashboard/index.js
```diff
@@ -19,7 +19,8 @@ const mapStateToProps = (state: Object) => ({
   GAS: getGAS(state),
   tokens: getTokens(state),
   loaded: getIsLoaded(state),
-  explorer: getBlockExplorer(state)
+  explorer: getBlockExplorer(state),
+  hardwareLogin: getHardwareLogin(state)
```
evgenyboxer: How about `isHardwareLogin` to indicate its a boolean

app/modules/transactions.js
dvdschwrtz: @evgenyboxer will do. also, this should never be called since I disabled the send button, but always good to make sure. we can put this back once ledger support is added

PR #419, opened 2017-12-15
Replaced readme slack link with discord link

mhuggins:
N/A

The README still linked to Slack instead of Discord.

I replaced the Slack link and text with Discord link and text.

I clicked the link.

No.

No.

- [ ] Unit tests written?
N/A


slipo:
Thanks!

PR #422, opened 2017-12-16
Added address book management for sending assets

mhuggins:
Address Book

Finishes #9.

It allows users to save & name addresses that they commonly send assets to.  There are simple mechanisms for saving (only valid addresses) with custom unique names, as well as deleting addresses that are no longer needed.

I modified the existing "Send" modal to extend the behavior of the recipient address text input.  It is now composed of a more complex set of react components that represent a select input w/ drop-down of existing values (if any values have been added yet).

When the user focuses on the input, it expands the drop-down to show existing values.  As the user starts typing, the values are filtered (by both address & name) to match the user's search term.

When a new address has been typed into the address input, a "save" icon is displayed so that the user can save it.  If that is clicked, an interactive tooltip allows the user to enter a name for the address before saving.

When a user wishes to delete an address book entry, they can click the trash can icon that appears next to each entry in the drop-down.

I smoke tested it manually, but I did not write any automated tests for it yet.

* I widened the SendModal to ensure that the recipient address is still completely visible even when the "save" icon is displayed.
* I removed the auto-focus of the recipient input since it automatically made the dropdown appear.  If this is not desired, I can revert that change and instead not show the dropdown until the input is clicked or the user starts typing.

* I'm new to Flow, so I easily could have messed up some of the syntax around that without knowing it.
* I added error handling to the redux actions for validation around local storage callbacks, but I didn't see this handled anywhere else.  I'm sure it could use some cleanup, as there's a little bit of duplicated code with the error handling.
* I didn't add any SCSS variables for the colors or spacing used.  I can do that if the styles look good, or I can refactor the styles first.
* This is my first contribution to a CoZ project.  Is there anything else I should do as part of submitting this?
* [ ] Unit tests written

### DIFF
--- a/app/components/Inputs/SelectInput/Dropdown.scss
+++ b/app/components/Inputs/SelectInput/Dropdown.scss
@@ -1,0 +1,7 @@
+.dropdown {
+  position: relative;
+
+  .content {
+    position: absolute;
+  }
+}
--- a/__tests__/components/Inputs/AddressInput/AddressInput.test.js
+++ b/__tests__/components/Inputs/AddressInput/AddressInput.test.js
@@ -1,0 +1,42 @@
+import React from 'react'
+import { shallow, mount } from 'enzyme'
+import AddressInput from '../../../../app/components/Inputs/AddressInput/AddressInput'
+
+describe('AddressInput', () => {
+  const props = {
+    addresses: {
+      'Lenny': 'AW4FD7bz6PF2QadFKF8qXUT7tNmWgvXZc4',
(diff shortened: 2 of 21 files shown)


app/components/Inputs/SelectInput/Dropdown.scss
```diff
@@ -0,0 +1,11 @@
+.dropdown {
+  position: relative;
+
+  .anchor {
+    // TODO
```
mhuggins: ☝️ I can remove this.

mhuggins:
🎉  Thanks guys!

PR #440, opened 2017-12-26
add ledger NEP5 support

dvdschwrtz:

- [ ] Unit tests written?


dvdschwrtz:
I will include these changes with my other PR which needs to cover some of the same code. Closing this for now

Issue #421, opened 2017-12-16
Update stored wallet to NEP6 standard


slipo:
https://github.com/neo-project/proposals/pull/13

PR #444, opened 2017-12-30
NEP-6 wallet support

slipo:
https://github.com/CityOfZion/neon-wallet/issues/421

All projects should follow the NEP-6 wallet standard so they can share wallets

Some notes:
I leave the keys.json storage file behind, just in case.

When we upgrade the wallets we only have labels and encrypted keys. This means that we can’t fill in the address fields until they log into the accounts.

I first tried to use neon-js which was pretty cool but it throws an exception if you try to export an Account without an address. It wasn’t too much work to do it all custom in the end.

Try to use naming inline with the NEP-6 standard (wallets contain accounts, in settings you add accounts not keys)

Automated + manual tests

QA Tests worth doing (any others?)
 * Starting with legacy key file, it should migrate to NEP-6. Delete an account, restart app and make sure that account is still missing.
 * Starting with no wallet, test recovering NEP-6 wallet
 * Starting with no wallet, test recovering legacy wallet
 * Adding accounts after the two above
 * Starting with no wallet, test adding 2 accounts, logging into both works 
 * Recovering NEP-6 / Legacy after already having accounts in the wallet
 * Logging into recovered legacy account
 * Logging into recovered NEP-6 account
 * Deleting all those accounts
 * Recovering same recovery file multiple times (won’t allow key to be in wallet twice)
 * Export everything, delete all accounts, import

No.

Having wallet accounts without public addresses could pose a challenge if other programs the user is importing into don't support that. There's not really anything we can do though.

- [x] Unit tests written?


slipo:
Per this issue:
https://github.com/electron-userland/electron-builder/issues/2260

I upgraded electron-builder and now the build works.

app/containers/Settings/Settings.jsx
```diff
@@ -102,28 +105,33 @@ export default class Settings extends Component<Props, State> {
...
-      text: `Please confirm deleting saved wallet - ${key}`,
+      text: `Please confirm deleting saved wallet - ${label}`,
       onClick: () => {
         // eslint-disable-next-line
-        storage.get('keys', (error, data) => {
-          delete data[key]
-          storage.set('keys', data)
-          setKeys(data)
+        storage.get('userWallet', (error, data) => {
+          data.accounts.map((account, index) => {
```
evgenyboxer: Can we do some error handling here?
` if (error) { ... } else { delete wallet account }`
Also can we use something like:
`const keys = reject(data.accounts, { key });`
Also, why do we need to check the label when deleting? isn't a key is unique?

__tests__/modules/generateWallet.test.js
```diff
@@ -29,31 +35,183 @@ describe('generateWallet module tests', () => {
 
   describe('newWallet tests', () => {
     const payload = account
-    const expectedAction = Object.assign({}, { payload }, { type: NEW_WALLET })
+    const expectedAction = Object.assign({}, { payload }, { type: NEW_WALLET_ACCOUNT })
```
evgenyboxer: Do we need an Object.assign here?
Can it be - 
`const expectedAction = { payload, type: NEW_WALLET_ACCOUNT }`

app/modules/generateWallet.js
```diff
@@ -27,22 +28,152 @@ export function newWallet (account: Object) {
...
+    label: label,
+    isDefault: isDefault, // Make the first account the default
+    lock: false,
+    key: key,
+    contract: {},
+    extra: null
+  }
+}
+
+export const createEmptyWallet = () => {
```
evgenyboxer: How about this - 
```
export const createEmptyWallet = ({
  ...DEFAULT_WALLET,
  scrypt: DEFAULT_SCRYPT
})
```

app/modules/generateWallet.js
```diff
@@ -27,22 +28,152 @@ export function newWallet (account: Object) {
...
+        resolve()
+      }
+    })
+  })
+}
+
+export const walletHasKey = (wallet, key) => {
+  let foundKey = false
+
+  each(wallet.accounts, (account) => {
```
evgenyboxer: Use .some - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

`wallet.accounts.some((account) => account.key === key)`

PR #446, opened 2017-12-30
WIP - (do not merge) Neoscan

dvdschwrtz:

A couple of things need to happen before this is pulled in. First, the changes I made to the neoscan API need to be in production, which should happen later tonight or tomorrow morning. Second, I am pointing at my local neon-js changes I made. I will need to merge those changes into neon-js and repoint the package.json to the CoZ repo before this is merged as well. 

- [ ] Unit tests written?

### DIFF
--- a/__mocks__/neon-js.js
+++ b/__mocks__/neon-js.js
@@ -25,7 +25,7 @@ neonjs.api = {
     claimGas: promiseMockGen(mockConfigResponse),
     sendAsset: promiseMockGen(mockConfigResponse),
     doInvoke: promiseMockGen(mockConfigResponse)
-  }
+  },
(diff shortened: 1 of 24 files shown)


dvdschwrtz:
A couple of things need to happen before this is pulled in. First, the changes I made to the neoscan API need to be in production, which should happen later tonight or tomorrow morning. Second, I am pointing at my local neon-js changes I made. I will need to merge those changes into neon-js and repoint the package.json to the CoZ repo before this is merged as well.

dvdschwrtz:
Also, leaving this comment to remind myself to resolve an issue with the tx flow types I need to look at again before everything is merged

PR #448, opened 2017-12-31
Added DeepBrain Coin

mhuggins:
N/A

It supports sending & showing balance for DBC (DeepBrain Coin), similar to RPX.

I added the MainNet & TestNet constants for the DBC hash.

I loaded the wallet and confirmed that I can view the 0 DBC balance & view DBC token information by clicking on the balance.  **NOTE:** I was unable to test transferring DBC due to having a 0 balance.

N/A

See **NOTE** above.

- [x] Unit tests written?

### DIFF
--- a/__tests__/modules/wallet.test.js
+++ b/__tests__/modules/wallet.test.js
@@ -13,6 +13,11 @@ describe('wallet module tests', () => {
     NEO: 0,
     GAS: 0,
     tokens: {
+      DBC: {
+        balance: 0,
+        scriptHash: 'b951ecbbc5fe37a9c280a76cb0ce0014827294cf',
+        symbol: 'DBC'
+      },
       RPX: {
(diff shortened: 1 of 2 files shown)


mhuggins:
The test failure doesn't appear to relate to my change.

slipo:
Just fyi: I had the same issue with the build. Here was my fix: https://github.com/CityOfZion/neon-wallet/pull/444/commits/bfe2090bca73e396884f626c8ecc8645ef7dd570

mhuggins:
@slipo Thanks, rebasing against `dev` fixed it!

ministoat:
Using a ledger I cannot send dbc with this version, error message says not supported

mhuggins:
@ministoat I don't have a ledger to test against.  I will have to see if someone else can help out.

Issue #253, opened 2017-10-18
Multiple sends


canesin:
In neo-gui is possible to add many transactions to a list before sending the tx.
The UX I envision is a one where you create a "multiple recipents" transaction by clicking a "+add" button instead of the "send", and it adds the transaction to the transaction list. that appears collapsed above the QR-Code of the address.
After the user finishes adding the transactions he can click on the send button by the side of it.

That combined with #9 would make really easy to send to many destinations quickly and reliably, like when we need to send weekly rewards.

evgenyboxer:
I’ll do it!

canesin:
@evgenyboxer my idea of UX/UI:

![image](https://user-images.githubusercontent.com/353648/31804833-3896f3ce-b52a-11e7-889c-37eca75e6318.png)

mhuggins:
I don't mind taking this on if no one has started on it.  I think it could be done more easily with the new UX.  I may just need help on the back-end portion with how to include multiple addresses when actually creating the transaction.

PR #451, opened 2018-01-02
Added sending to multiple recipients (and sending assets + tokens)

mhuggins:
multiple-recipients

#253

Currently, a transaction can only include a single asset or token going to a single recipient address.

I revamped the screens within the SendModal component, and I replaced the existing `api.neonDB.doSendAsset` and `api.nep5.doTransferToken` calls with the newer `api.sendAsset` and `api.doInvoke` functions instead.  The `doInvoke` function is only being used if the user is attempting to transfer tokens (e.g.: RPX), otherwise the `sendAsset` function is used.

Smoke tested with sending assets and tokens across multiple addresses.  Verified transactions on block explorer:

* https://testnet.neotracker.io/tx/4ce187fd80beccd16c8885b474ab05efa7dff3f3c5e2e239656c3806fc1a2d43
* https://testnet.neotracker.io/tx/345a03a9ab7c44ed1b6522e1736ab695b7075b2afaecebc22d5d8cbc28af0d7f

I made a few changes and added some improvements as I went:

* Completely changed the layout of the `SendModal` when entering the asset/amount/address.
* Replaced the asset dropdown with a `SelectInput` that includes autocompletion.
* Updated the `NumberInput` to include a "MAX" button.  This replaces the 25%, 50%, 75%, 100% buttons that were there previously.
* Added a new `Address` component that represents the clickable address for viewing an address in the selected block explorer.
* Replaced any calls to `<button>` with the `<Button>` component instead.  Ensured it's being rendered anywhere it was changed.
* Extracted a `withAddressCheck` higher-order component to separate the loading state on the send verification screen from actually displaying the table.
* Added an "agree" checkbox on the verification screen to confirm before sending.

I worked with @snowypowers for a few hours over Slack tonight to diagnose some issues I was running into with getting the transactions to work, and I want to make sure he gets credit for his efforts.  HUGE thank you to him for his help!

- [x] Unit tests written?


app/core/nep5.js
```diff
@@ -1,4 +1,12 @@
...
+
 import { COIN_DECIMAL_LENGTH, toFixedDecimals } from './formatters'
 
-export const adjustDecimalAmountForTokenTransfer = (value: number) => toFixedDecimals(value, COIN_DECIMAL_LENGTH) * 10 ** COIN_DECIMAL_LENGTH
+export const adjustDecimalAmountForTokenTransfer = (value: number) => {
+  return toFixedDecimals(value, COIN_DECIMAL_LENGTH) * 10 ** COIN_DECIMAL_LENGTH
+}
+
+export const sendTokens = (net, scriptHash, publicKeyOrWif, sendAddress, sendAmount, gasCost = 0, signingFunction = null) => {
+  return api.nep5.doTransferToken(net, scriptHash, publicKeyOrWif, sendAddress, sendAmount, gasCost, signingFunction)
```
mhuggins: ✂️ This change wasn't needed, can remove.

flow-typed/declarations.js
```diff
@@ -58,8 +58,14 @@ declare type TokenInfoType = {
...
 declare type SymbolType = $Keys<typeof TOKENS> | 'NEO' | 'GAS'
 
-declare type TokenType = {
+declare type TokenType = { // eslint-disable-line no-undef
   symbol: SymbolType,
   balance: number,
   info: TokenInfoType
 }
+
+declare type SendEntryType = { // eslint-disable-line no-undef
```
mhuggins: I was getting eslint errors without adding these comments here.  I can remove them if there's a better fix.

mhuggins:
@evgenyboxer This is working and ready for a review now. 😄

PR #499, opened 2018-01-08
Add RHT token (Redeemable HashPuppy Token)

weppos:
DISCLAIMER:

I am not affiliated with RHT. I just happen to [discover it recently](https://steemit.com/cryptocurrency/@weppos/rht-token-redeemable-hashpuppy-token) and being a NEO user and developer, I was curious to see how complicated would be to add the token to the wallet.

Feel free to close/reject the PR if this is not in line with the current dev roadmap.

---

No existing ticket.

This PR adds the new RHT token. See https://neotracker.io/asset/2328008e6f6c7bd157a342e789389eb034d9cbc4

I added the token to the token list.

I tested it manually compiling the app. I also run the test suite with `yarn test`.

Nope.

- [x] Unit tests written?

**Notes**

I was not able to find a reliable test token. There are several tests in the TestNet (e.g. https://testnet.neotracker.io/asset/f9572c5b119a6b5775a6af07f1cef5d310038f55) but none of them seems to be the final one.

Therefore, I did not include the token in the TestNet list.

I also tried to find if there was any convention for tokens (e.g. alphabetically) but it looks like there wasn't. I added the token in alpha order compared to the existing entries.

### DIFF
--- a/__tests__/modules/wallet.test.js
+++ b/__tests__/modules/wallet.test.js
@@ -27,6 +27,11 @@ describe('wallet module tests', () => {
         balance: '0',
         scriptHash: 'ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9',
         symbol: 'RPX'
+      },
+      RHT: {
+        balance: '0',
+        scriptHash: '2328008e6f6c7bd157a342e789389eb034d9cbc4',
+        symbol: 'RHT'
       }
     },
     transactions: []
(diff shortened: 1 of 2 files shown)


mhuggins:
TestNet token appears to be here: https://testnet.neotracker.io/asset/400cbed5b41014788d939eaf6286e336e7140f8c

weppos:
@mhuggins I'm not sure that's correct. They created several tests:

https://testnet.neotracker.io/asset/400cbed5b41014788d939eaf6286e336e7140f8c
https://testnet.neotracker.io/asset/d3de84c166d93ad2581cb587bda8e02b12dc37ca
https://testnet.neotracker.io/asset/f9572c5b119a6b5775a6af07f1cef5d310038f55

clshack:
The right token seems this one:
https://neotracker.io/asset/2328008e6f6c7bd157a342e789389eb034d9cbc4


Referring to the dev twitter:
https://twitter.com/hash_puppies/status/949601097005617152

weppos:
@clshack yes, that's correct, and this is the token that it's used in this PR for the MainNet.
The question about the token was for the TestNet.

PR #501, opened 2018-01-08
rht - redeemable hashpuppy token constant

phetter:
This adds hashpuppy support for neon-wallet

### DIFF
--- a/app/core/constants.js
+++ b/app/core/constants.js
@@ -67,14 +67,16 @@ export const MODAL_TYPES = {
 export const TOKENS_TEST = {
   DBC: 'b951ecbbc5fe37a9c280a76cb0ce0014827294cf',
   RPX: '5b7074e873973a6ed3708862f219a6fbf4d1c411',
-  QLC: '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5'
+  QLC: '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5',
+  RHT: '2328008e6f6c7bd157a342e789389eb034d9cbc4'
 }
 
 // MainNet
 export const TOKENS = {
   DBC: 'b951ecbbc5fe37a9c280a76cb0ce0014827294cf',
   RPX: 'ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9',
-  QLC: '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5'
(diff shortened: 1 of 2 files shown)


phetter:
Okay so in all honesty I didn't notice that this was a duplicate (nearly). I simply didn't check the existing PRs since I thought there was no way someone would've added it. What a coincidence. This version does have the test token added though.

weppos:
@phetter FYI the test token doesn't exist, reason I kept it out on purpose.
https://testnet.neotracker.io/asset/2328008e6f6c7bd157a342e789389eb034d9cbc4

valdor27:
wow this is actually quite a coincidence haha

app/core/constants.js
```diff
@@ -67,14 +67,16 @@ export const MODAL_TYPES = {
...
+  RHT: '2328008e6f6c7bd157a342e789389eb034d9cbc4'
 }
 
 // MainNet
 export const TOKENS = {
   DBC: 'b951ecbbc5fe37a9c280a76cb0ce0014827294cf',
   RPX: 'ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9',
-  QLC: '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5'
+  QLC: '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5',
+  RHT: '2328008e6f6c7bd157a342e789389eb034d9cbc4'
```
mhuggins: You probably want this as the TestNet value: https://testnet.neotracker.io/asset/400cbed5b41014788d939eaf6286e336e7140f8c

weppos: I'm not sure this is correct. There are 4 other RTH test tokens that were created after. It looks like they run several tests, without having a final one.

mhuggins: Might be worth pinging their team then.

mhuggins: Just a thought.  Not sure if it's really vital. 😄

clshack: Token seems to be alive in the latest version:
https://neotracker.io/asset/2328008e6f6c7bd157a342e789389eb034d9cbc4

Also refer to the dev website:
https://twitter.com/hash_puppies/status/949601097005617152

dvdschwrtz: There was another PR with the same token value that I merged already before I saw these comments. we will remove it when testing if it is not the correct value. I am closing this for now

PR #528, opened 2018-01-14
WIP: Token sale

dvdschwrtz:

- [ ] Unit tests written?


app/modules/sale.js
```diff
@@ -1,57 +1,120 @@
...
+  )
 
   const [error, rpcEndpoint] = await asyncWrap(api.neonDB.getRPCEndpoint(net)) // eslint-disable-line
-  const [err, balance] = await asyncWrap(api.nep5.getTokenBalance(rpcEndpoint, _scriptHash, account.address)) // eslint-disable-line
-  const [e, response] = await asyncWrap(api.neonDB.doMintTokens(net, _scriptHash, wif, toMint, 0))
+  const [err, balance] = await asyncWrap(
+    api.nep5.getTokenBalance(rpcEndpoint, _scriptHash, account.address)
+  ) // eslint-disable-line
+
+  const intents = [[NEO, neoToMint], [GAS, gasToMint]]
```
mhuggins: I don't think `NEO` and `GAS` are symbols here, should these be `ASSETS.NEO` and `ASSETS.GAS`?

PR #533, opened 2018-01-15
 FIX: application gets in bad state if incorrect json filed is used when restoring wallet

ReganL:
… recovery method

Does not fix any open issues

Application gets in an unusable state if an invalid JSON file is imported into wallet recovery in the Settings page. The code is assuming a String type for the account keys and does not handle some other types.

A basic schema approach to new accounts which checks for a valid type when setting keys. Instead of the current approach which accepts the import and then crashes on an Array.slice method if an incorrect type appears in the account.key it only accepts string key types and uses the Constructor to generate a new account object which has all of the relevant properties either passed through or defaults are set

Tested importing various JSON files that were not Neon wallet files. A good test example is just the project's package.json file

No special changes.

I will look into validating the file import at a later stage and verifying file is of correct format and adding notices if file is incorrect if this is desired. For now this just prevents a crash

Fixed small typo in notification Settings.jsx

- [ ] Unit tests written?
TBC

### DIFF
--- a/app/containers/Settings/Settings.jsx
+++ b/app/containers/Settings/Settings.jsx
@@ -128,7 +128,7 @@ export default class Settings extends Component<Props, State> {
       onClick: () => {
         storage.get('userWallet', (readError, data) => {
           if (readError) {
-            showErrorNotification({ message: `An error occurred reading previosly stored wallet: ${readError.message}` })
+            showErrorNotification({ message: `An error occurred reading previously stored wallet: ${readError.message}` })
           }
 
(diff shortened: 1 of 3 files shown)


dvdschwrtz:
great job! welcome to neon wallet :)

ReganL:
Cheers @dvdschwrtz . I'd be willing to participate a bit more actively on this project. Let me know if you have a channel or Trello board where issues and features are logged, or discussions are had, and I'll participate/assist where I can!

PR #539, opened 2018-01-15
Add custom tokens

evgenyboxer:
Sorry for the massive PR. But this was indeed a lot of work.
It adds a "Manage Tokens" modal to manage user defined NEP5 tokens. Tokens can be defined for each network (Right now only Test/Main).


app/components/Modals/BaseModal/BaseModal.jsx
```diff
@@ -11,14 +13,25 @@ type Props = {
     width?: string,
     height?: string,
     className?: string,
+    bodyClassName?: string,
```
mhuggins: This class isn't being used anywhere.

app/components/Modals/TokenModal/Row.jsx
```diff
@@ -0,0 +1,67 @@
...
+    return (
+      <div className={styles.row}>
+        <input
+          className={classNames(styles.rowLabel, {[styles.rowError]: isSymbolInvalid})}
+          type='text'
+          readOnly={!token.isUserGenerated}
+          disabled={!token.isUserGenerated}
+          defaultValue={token.symbol}
+          placeholder='symbol'
+          ref={(node) => { this.labelInput = node }}
```
mhuggins: Would a better name for this be `symbolInput`?

app/components/Modals/TokenModal/Row.jsx
```diff
@@ -0,0 +1,67 @@
...
+          onChange={(e) => token.isUserGenerated && onChangeSymbol(e.target.value)}
+        />
+        <input
+          className={classNames(styles.rowURL, {[styles.rowError]: isScriptHashInvalid})}
+          type='text'
+          readOnly={!token.isUserGenerated}
+          disabled={!token.isUserGenerated}
+          placeholder='script hash'
+          ref={(node) => { this.urlInput = node }}
```
mhuggins: Would a better name for this be `scriptHashInput`?

app/components/Modals/TokenModal/TokenModal.jsx
```diff
@@ -0,0 +1,150 @@
...
+          <form onSubmit={(e) => {
+            e.preventDefault()
+            this.saveAndValidateTokens()
+          }}>
+            <div className={styles.rowsContainer}>
+              {tokens.filter(token => token.networkId === networkId).map((token: TokenItemType) => (
+                <Row
+                  token={token}
+                  isSymbolInvalid={errorItemId === token.id && errorType === 'symbol'}
+                  isScriptHashInvalid={errorItemId === token.id && errorType === 'scriptHash'}
```
mhuggins: "symbol" and "scriptHash" on these two lines don't match the flow type:

```js
type InputErrorType = 'label' | 'url'
```

app/components/Modals/TokenModal/utils.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const getNewTokenItem = (networkId: string) => ({
+  id: uniqueId(TOKEN_ID_PREFIX),
+  symbol: '',
+  scriptHash: '',
+  networkId,
+  isUserGenerated: true
+})
+
+export const validateTokens = (tokens: Array<TokenItemType>) => {
```
mhuggins: I think giving the return value of this object a flowtype would have made it obvious that `InputErrorType` wasn't matching in my previous comment.  e.g.:

```js
type TokenValidationResult = {
  errorItemId: number | null,
  errorType: InputErrorType | null,
  errorMessage: string | null
}

export const validateTokens = (tokens: Array<TokenItemType>): TokenValidationResult => { ... }
```

mhuggins:
@evgenyboxer I told the wallet guys that this can be merged, but the feedback might be a nice follow-on if you feel like addressing it.  I can always address it too if you prefer. 😄

PR #542, opened 2018-01-16
Boilerplate for refactoring redux actions & maintaining loading state

mhuggins:
N/A

This adds boilerplate for easily implementing API interactions and exposing a loading/failure state in the UI associated with each call.

I implemented redux-saga into the code base along with a handful of helpful action creators & higher-order components that work hand-in-hand for rendering data or loading/failure state based upon the result of any function call (e.g.: neon-js API calls).

*Next Steps: (TBD after this is merged)*

* Add tests for new boilerplate code.
* Add documentation for action generators & higher-order components.
* Move boilerplate code into an internal npm package (neon-redux?) that can be published in the future.
* Replace more in-component action calls with container loading & visible loading state.

I smoke tested that changing the network updates anywhere the network is referenced.

* There are a few placeholder TODO's that I will be circling back on.  It was becoming too big of PR, so I kept what's here as a cutoff point and will be continuing after this gets merged, as long as that works for the team.
* I had to include a deprecated `getNetwork` call until some more refactoring can happen on existing module methods.  My intention is to convert these into higher-order components that expose methods that can be used for making calls (such as claiming gas).

Because there's a combination of thunk + saga code in here, it might look like more work to start with.  As this refactor continues, the benefits will become more clear, and the code quality will appear cleaner and more succinct.

*PLEASE help to thoroughly test this!  The main thing to check is if everything works as expected when switching between MainNet and TestNet.*

**Unit tests written?**
Not for the boilerplate code yet, I'm planning to circle back on this next week.


mhuggins:
@evgenyboxer I just rebased this against your network changes since I had a ton of merge conflicts (we were both working on changes around getting/setting network).  I haven't smoke tested this again after doing so to ensure it didn't break anything.  This could use some thorough QA regardless though.

mhuggins:
@evgenyboxer I fixed the merge conflicts between both our changes around the network state.  I had to add a few temporary functions (I included a new "deprecated.js" file to denote this) while in the process of refactoring.  As I move more logic over to use these new higher order component approach, we'll be able to drop these functions.

PR #546, opened 2018-01-17
Token sale UI

evgenyboxer:
Implement the token sale UI


app/components/Modals/TokenModal/utils.js
```diff
@@ -1,11 +1,8 @@
 // @flow
-import { uniqueId } from 'lodash'
-
-const TOKEN_ID_PREFIX = '_ug'
+import uuidv4 from 'uuid/v4'
 
 export const getNewTokenItem = (networkId: string) => ({
-  id: uniqueId(TOKEN_ID_PREFIX),
-  symbol: '',
+  id: uuidv4(),
```
mhuggins: Nice improvement.

app/core/sale.js
```diff
@@ -0,0 +1,86 @@
...
+  GAS: number
+) => {
+  let message
+
+  if (neoToMint < 0 || gasToMint < 0 || (neoToMint === 0 && gasToMint === 0)) {
+    message = 'You must send positive amounts of NEO or GAS.'
+    return [false, message]
+  }
+
+  if (neoToMint && parseFloat(neoToMint) !== parseInt(neoToMint)) {
```
mhuggins: We should avoid adding more `parseFloat` calls.  Can we change this to:

```js
if (neoToMint && !toBigNumber(neoToMint).isInteger()) { ... }
```

app/modules/sale.js
```diff
@@ -1,58 +1,199 @@
...
   const net = getNetwork(state)
+  const address = getAddress(state)
+  const isHardwareLogin = getIsHardwareLogin(state)
+  const signingFunction = getSigningFunction(state)
 
   const account = new wallet.Account(wif)
-  if (!toBigNumber(neoToSend).isInteger()) {
-    dispatch(showErrorNotification({ message: 'You cannot send fractional NEO to a token sale.' }))
+  const neoToMint = toNumber(neoToSend)
+  const gasToMint = toNumber(gasToSend)
```
mhuggins: I think we should keep these as a string when calling `validateMintTokensInputs`, then let `toBigNumber` be called internally from there.

app/core/sale.js
```diff
@@ -0,0 +1,86 @@
+// @flow
+export const validateMintTokensInputs = (
+  neoToMint: number,
+  gasToMint: number,
```
mhuggins: These would need to be `string` instead of `number` if we want to go with the `toBigNumber` approach here.

app/core/sale.js
```diff
@@ -0,0 +1,86 @@
...
+    message = 'You cannot send fractional NEO to a token sale.'
+    return [false, message]
+  }
+
+  if ((neoToMint && isNaN(neoToMint)) || (gasToMint && isNaN(gasToMint))) {
+    message = 'Please enter valid numbers only'
+    return [false, message]
+  }
+
+  if (neoToMint > NEO) {
```
mhuggins:
```js
if (toBigNumber(neoToMint).gt(NEO)) { ... }
```

app/core/sale.js
```diff
@@ -0,0 +1,86 @@
...
+    message = 'You do not have enough NEO to send.'
+    return [false, message]
+  }
+
+  if (gasToMint > GAS) {
```
mhuggins: etc.

app/modules/sale.js
```diff
@@ -1,58 +1,199 @@
...
+  gasToSend: string,
+  scriptHash: string,
+  gasCost: string = '0'
+) => async (dispatch: DispatchType, getState: GetStateType) => {
   const state = getState()
   const wif = getWIF(state)
-  const NEO = getNEO(state)
+  const publicKey = getPublicKey(state)
+  const NEO = toNumber(getNEO(state))
+  const GAS = toNumber(getGAS(state))
```
mhuggins: We should use `toBigNumber` on these, or just pass the string through to any internal functions we call.

PR #572, opened 2018-01-18
version 0.1.1

Ejhfast:

- [ ] Unit tests written?


mhuggins:
Commit d891984e596c2232e84c63b5d43af3a70f6ab164 was not a part of the 0.1.1 release.

mhuggins:
We now have 0.1.1 in master as a result of merging #606.

Issue #540, opened 2018-01-15
Add md5 Checksum Values to the Website


feuillet:
Guys, please add the md5 Checksum values to the Website.
For stuff like Coin Wallets this is definitely a must.
Thanks.

PR #584, opened 2018-01-20
updates to neonwallet.com

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/540

This PR also includes a request https://api.github.com/repos/CityOfZion/neon-wallet/releases/latest which will no longer make it necessary to update download links every time we release.

When the application loads it defaults to `config.js` but after making http request overrides with latest release information.

Compiled build tested and new request functions to https://api.github.com/repos/CityOfZion/neon-wallet/releases/latest tested extensively.

This now contains a minified build which will be a better UX for the average viewer of the page (theres also no need to update download links every release)

All of the latest can be found here https://github.com/comountainclimber/neon-website

- [ ] Unit tests written?


comountainclimber:
Thanks for taking a look @mhuggins

PR #592, opened 2018-01-22
FIX duplicate accounts can be saved when generating a new wallet

ReganL:
No current issues

When generating a wallet, the user can save the wallet in accounts. Current behaviour allows the user to save duplicates of the newly generated account. This can cause an issue if the user tries to delete one of the accounts at any point. The delete functionality deletes by key and will delete both instances of an account in state, even though one of them remains in the current view.

Utilising the same function used to validate wallet imports, we check if the key exists in state and display a warning if it already exists.

Generate a new wallet and click the save button twice. First save shows a success notification. Second save shows a warning with the account address in the notification.

- [ ] Unit tests written?


dvdschwrtz:
@Ejhfast can we invite @ReganL to the CoZ slack channel?

PR #611, opened 2018-01-27
Replaced balance, claim, & transaction history actions/reducers with new actions architecture

mhuggins:
N/A

This is a continuation of the redux refactor from PR's #542, #575, #586, and #595. It replaces most of the claim & wallet actions & reducers with a new set of actions for fetching balances, available claims, and transaction history.  It also renames the accountActions to authActions for clarity.

In addition to this work, I also:

* Removed the complex `loadWalletData` function and replaced it with a batch request in the new redux architecture (a long time coming!).
* Introduced several new higher-order components:
  * `withResponsiveAction`: perform an action when a prop changes or condition is met.
  * `withReset`: resets data in a portion of the store when a prop changes or condition is met.
  * `withLogoutReset`: resets action data when the user logs out.
* Cleaned out all data related to the current account when the user logs out.
* Added a `testHelpers.js` file to help consolidate repetitive test code.
* Prevent tokens data from being cleared on a bad balance fetch.

I added new actions for fetching balances, claims, and transaction history using the new actions architecture.

I smoke tested and addressed any test failures associated with these changes.

N/A

N/A

- [x] Unit tests written?


evgenyboxer:
Fantastic work @mhuggins

mhuggins:
@evgenyboxer Rebased.

mhuggins:
Looks like another intermittent failure on these end-to-end tests.  These were working prior to rebasing.

PR #625, opened 2018-01-30
Add tnc script hash

pigxo:

- [ ] Unit tests written?


mhuggins:
This shouldn't be needed anymore since users can add their own script hashes in the settings screen.  That feature was added in v0.1.0.

dvdschwrtz:
@pigxo do you work for trinity? happy to have you as a contributor to NEO :) I am closing this PR as I agree with @mhuggins. TNC should notify their holders with instructions on adding the token to neon wallet. Thanks!!

PR #654, opened 2018-02-03
Backfilled tests for action creators and saga

mhuggins:
N/A

This fills some holes with regards to the test suite around some of the new redux architecture that was introduced.

I added tests!

`yarn test`

N/A

N/A

- [x] Unit tests written?


mhuggins:
Merging since this doesn't change any behavior, just adds tests.

PR #668, opened 2018-02-08
Add NRV and Enable Verification

brianlenz:
* Added the NRV token sale script hash.

* Enabling mintTokens verification by default.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,8 +21,8 @@ jobs:
       - checkout
       - restore_cache:
           key: neon-wallet-{{ checksum "yarn.lock" }}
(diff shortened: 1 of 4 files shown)


buddhaman8:
Can the same be done for QLC?

brianlenz:
Removing `sudo` in the `config.yml` for circleci fixed the sudo error:

> /bin/bash: sudo: command not found
> Exited with code 127

But that revealed a different issue with building the linux x64 binary:

> Error: Exit code: ENOENT. spawn gm ENOENT

Fixed that by adding "graphicsmagick" to the `apt-get install`.

Now we have the pretty ✅

PR #676, opened 2018-02-11
Fixed signing function assignment for ledger login

mhuggins:
N/A

When I refactored the login actions, I neglected to set the signing function for ledger logins.  I wasn't able to test this since I don't have a ledger yet, but I happened to notice that it has no way of being set since it's not being passed in as an argument.  This should fix it.

I set the signing function to the function defined in the ledger module, which matches the way `master` branch sets the signing function.

I can't since I don't have a ledger.  Ledger login may be broken regardless of this change, someone needs to test it out on the `dev` branch.

N/A

N/A

- [ ] Unit tests written?


mhuggins:

PR #677, opened 2018-02-11
Added NEO/GAS price chart to dashboard

mhuggins:
N/A

This implements the NEO/GAS price chart based upon @asalib's designs.  (Proper fonts will be added in a follow-up PR.)

price-history-chart

I added a new action set to fetch the data from cryptocompare.com's API, and I added the recharts package for rendering the chart in a new dashboard panel.  The user can toggle between NEO & GAS assets, as well as 1 month, 1 week, and 1 day timeframes.

I smoke tested it.  I'll be adding tests in a follow-up PR.

* New `recharts` package introduced.
* I increased the window width from 1000px to 1280px to match Alf's designs.

N/A

- [ ] Unit tests written?


mhuggins:

PR #678, opened 2018-02-12
Added tests for block height actions

mhuggins:
N/A

This backfills some tests around block height fetching actions.

I added tests.

I ran the test suite.

N/A

N/A

- [x] Unit tests written?


mhuggins:

PR #703, opened 2018-02-14
Updated logout button to display at bottom of sidebar

mhuggins:
N/A

It moves the logout button to the bottom of the sidebar to match the designs.

*Before:*
sidebar-before

*After:*
sidebar-after

I used flexbox to align the sidebar sections.

Looked at the sidebar.

N/A

N/A

- [ ] Unit tests written?


evgenyboxer:
Looks great.

I'm merging this despite the build failing. Its not caused by your PR.

PR #704, opened 2018-02-14
Added tests for claim actions

mhuggins:
N/A

This backfills tests for the claim actions.

I added tests!

I ran the test suite.

N/A

N/A

- [x] Unit tests written?


mhuggins:
@dvdschwrtz I also have this PR from before, I can reopen against the `rebase` branch if needed.  What's your preference?

PR #706, opened 2018-02-14
Fixes validation of tokens when sending a transaction

mhuggins:
N/A

When the new actions were introduced, the shape of tokens in the redux store changed.  This was reflected properly when adding/validating individual recipients being added to a transaction in the send modal, but it was not accounted for when confirming/submitting to actually process & send the transaction.

This fixes the balances shape to match the expectations of the validation function.

I smoke tested it and observed the shape in the redux store & breakpoints in the code.

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/app/core/deprecated.js
+++ b/app/core/deprecated.js
@@ -1,5 +1,5 @@
 // @flow
-import { get, omit } from 'lodash'
+import { get, omit, pick } from 'lodash'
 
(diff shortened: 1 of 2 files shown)


mhuggins:
This is failing because the dev branch is failing right now.  The dev branch should pass again once rebasing against master is complete.

PR #720, opened 2018-02-16
Please allow hiding tokens on the Neon wallet display. Under manage tokens, currently cannot be changed

MudWar:

- [ ] Unit tests written?


mhuggins:
This is a good suggestion, but I believe it should be marked as an issue, not a pull request.

MudWar:
Oh yeah, sorry, first post here

PR #751, opened 2018-02-21
Feature/add esc key prop to modal

comountainclimber:

@canesin pointed that its frustrating UX to have some modals close via the `esc` key. This PR adds prop to the `BaseModal` component so that it can condtionally close via `esc` this prop is called
`shouldCloseWithEscapeKey`.

By adding a new prop to `BaseModal`

Testing locally

- [ ] Unit tests written?


app/components/Modals/BaseModal/BaseModal.jsx
```diff
@@ -30,11 +31,12 @@ const BaseModal = ({
...
   bodyClassName,
   style,
-  onAfterOpen
+  onAfterOpen,
+  shouldCloseWithEscapeKey
 }: Props) => (
   <ReactModal
     isOpen
-    onRequestClose={hideModal}
+    onRequestClose={() => shouldCloseWithEscapeKey ? hideModal() : undefined}
```
mhuggins: :neckbeard: `shouldCloseWithEscapeKey && hideModal()` is a little more concise.

comountainclimber: nice will change that thanks @mhuggins

comountainclimber:
@mhuggins I made a separate branch and PR with all commits squashed down into one https://github.com/CityOfZion/neon-wallet/pull/752

PR #761, opened 2018-02-22
Update Ledger code

snowypowers:
We are using outdated ledger to communicate with the Ledger Nano S. While there are currently no issues, we should be relying on the ledger repo for latest updates.

**Features**

- Spanking new ledger code dependent on LedgerHQ's ledgerjs
- Configurable BIP44 built in so you can now select the other accounts hidden in your ledger.
- Maintained by me so you can ask for features :)
- I will probably pull this out into a seperate repo in the future but lets just roll with this for now.

It adds the Ledger repo and refactors the existing code to rely on the new ledger code.

- Add ledger dependency
- Write new ledger code
- Hook up the exports through `ledgerNanoS.js`

Manual testing with my own ledger. Due to the multitude of changes ongoing with `dev`, I would like to put off intensive testing till the transition internally is complete. For example, the current claim gas function does not wait for the first transaction to clear. However, the 3 core functionalities (sendAsset, claimGas and doInvoke) are tested.

I removed a lot of dependencies that were basically left over from the days before `neon-js`.

I added a line to automatically open devtools when running the wallet in dev. Helps a lot in windows because nothing I do can open the devtools. And its dev only.

The new ledger code is in the file `neonLedger.js`
The wallet is dependent on the exports from `ledgerNanoS.js`

This setup allows me to write whatever I want while keeping the old interface.

- [ ] Unit tests written?


app/ledger/neonLedger.js
```diff
@@ -0,0 +1,199 @@
...
+
+  return r + s
+}
+
+export const getPublicKey = async (acct: number = 0): Promise<string> => {
+  const ledger = await NeonLedger.init()
+  try {
+    return await ledger.getPublicKey(acct)
+  } finally {
+    ledger.close()
```
mhuggins: Since `ledger.close` is a promise, should we either `await` or `return` to ensure it processes?

app/ledger/neonLedger.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+}
+
+export const getDeviceInfo = async () => {
+  const ledger = await NeonLedger.init()
+  try {
+    return await ledger.device.device.getDeviceInfo()
+  } finally {
+    ledger.close()
```
mhuggins: Same here.

app/ledger/neonLedger.js
```diff
@@ -0,0 +1,199 @@
...
+  try {
+    const data = typeof unsignedTx !== 'string' ? tx.serializeTransaction(unsignedTx, false) : unsignedTx
+    const publicKey = await ledger.getPublicKey(acct)
+    const invocationScript = '40' + await ledger.getSignature(data, acct)
+    const verificationScript = wallet.getVerificationScriptFromPublicKey(publicKey)
+    const txObj = tx.deserializeTransaction(data)
+    txObj.scripts.push({ invocationScript, verificationScript })
+    return tx.serializeTransaction(txObj)
+  } finally {
+    ledger.close()
```
mhuggins: Same here.

app/ledger/neonLedger.js
```diff
@@ -0,0 +1,199 @@
...
+  const ledger = await NeonLedger.init()
+  try {
+    const data = typeof unsignedTx !== 'string' ? tx.serializeTransaction(unsignedTx, false) : unsignedTx
+    const invocationScript = '40' + await ledger.getSignature(data, acct)
+    const verificationScript = wallet.getVerificationScriptFromPublicKey(publicKeyEncoded)
+    const txObj = tx.deserializeTransaction(data)
+    txObj.scripts.push({ invocationScript, verificationScript })
+    return tx.serializeTransaction(txObj)
+  } finally {
+    ledger.close()
```
mhuggins: Same here.

app/ledger/neonLedger.js
```diff
@@ -0,0 +1,199 @@
...
+
+  let s = response.substring(sStart, sEnd)
+
+  while (r.length < 64) {
+    r = '00' + r
+  }
+
+  while (s.length < 64) {
+    s = '00' + s
+  }
```
mhuggins: :neckbeard: You could use `String.prototype.padStart` in place of both these for-loops:

```js
r = r.padStart(64, '0')
s = s.padStart(64, '0')
```

app/ledger/neonLedger.js
snowypowers: I would say these are just fire and forgets? There isnt anything about closing properly and I havnt seen any errors due to closing.

app/ledger/neonLedger.js
snowypowers: This is inherited code lol I will clean up the parsing tonight. I have a good idea of what is happening there.

app/ledger/neonLedger.js
mhuggins: I considered that, I wasn't sure if it was an active decision though.  I wanted to call it out in case it wasn't.

app/ledger/neonLedger.js
mhuggins: FWIW, I use the :neckbeard: emoji to represent "if you feel like it".  Don't feel obligated to refactor it.

snowypowers: Cleaner code is always better :) Anyway looks like I would be adding on stuff if Coranos replies with the error code I should look out for.

app/ledger/neonLedger.js
mhuggins: Turns out the async is needed to prevent HID errors around accessing a closed device.

PR #800, opened 2018-03-03
Added short-term fix for resetting redux store keys on logout

### LINKED ISSUE #796: dev branch: Stale wallet data displayed when loading new wallet
This appears to be a new issue on the `dev` branch as compared with 0.1.4.  On 0.1.4, any time you opened a new wallet, you would see the green "loading" bars displayed until the wallet's data was loaded. On the `dev` build, those loading bars appear to have been removed and instead you see the _previous_ wallet's balances displayed temporarily until the new wallet's balances are fetched and updated in the UI. This is probably just a minor issue, but if there is a network issue, then it can lead to old balances being displayed for longer which is unexpected and undesirable.

I think the previous behavior of not showing any balances until the data is fetched is the preferable UX.

mhuggins:
Fixes #796.

This resets the authenticated user's data when they logout from the wallet.  Without this, the next login during the same app session will end up displaying the previous account's data until the new data has loaded.

This is a temporary quick fix.  The issue is that calling `reset` on a batch action is not resulting in each associated action's `reset` function being called even though it should be.  (Specifically calling `reset` on accounts action should trigger `reset` for balances, claims, and transaction history.)

This fixes the issue for release.  I will circle back on a long term fix for batch resets.

Before fix:
* Logged in as user "A".
* Saw loading spinner, followed by user "A" account info.
* Logged out.
* Logged in as user "B".
* Saw user "A" account info, followed by user "B" account info.

After fix:
* Logged in as user "A".
* Saw loading spinner, followed by user "A" account info.
* Logged out.
* Logged in as user "B".
* Saw loading spinner, followed by user "B" account info.

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/app/containers/App/index.js
+++ b/app/containers/App/index.js
@@ -5,7 +5,9 @@ import { compose } from 'recompose'
 
 import appActions from '../../actions/appActions'
 import authActions from '../../actions/authActions'
-import accountActions from '../../actions/accountActions'
+import balancesActions from '../../actions/balancesActions'
+import claimsActions from '../../actions/claimsActions'
+import transactionHistoryActions from '../../actions/transactionHistoryActions'
 import networkActions from '../../actions/networkActions'
 import withFetch from '../../hocs/api/withFetch'
 import withReload from '../../hocs/api/withReload'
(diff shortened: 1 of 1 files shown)


app/containers/App/index.js
```diff
@@ -61,5 +63,8 @@ export default compose(
 
   // Remove stale data from store on logout
   withLogoutReset(authActions),
-  withLogoutReset(accountActions)
+  // TODO: replace these three calls with `withLogoutReset(accountActions)` once batch reset is fixes
```
mhuggins: Oops.  Wrote "fixes" instead of "fixed".

PR #807, opened 2018-03-05
Load balance

dvdschwrtz:

Adds Load Balancing to Neon Wallet

Points the methods towards neoscan with neonDB as a fallback

using the api.loadBalance method

QA testing and running the test suites

- [ ] Unit tests written?

### DIFF
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -16,7 +16,7 @@ type Props = {
 export const ID = 'BALANCES'
 
 async function getBalances ({ net, address, tokens }: Props) {
-  const endpoint = await api.neonDB.getRPCEndpoint(net)
+  const endpoint = await api.loadBalance(api.getRPCEndpointFrom, { net })
 
   // token balances
   const promises = tokens.map(async (token) => {
@@ -37,7 +37,7 @@ async function getBalances ({ net, address, tokens }: Props) {
 
   // asset balances
   promises.push((async () => {
-    const assetBalances = await api.neonDB.getBalance(net, address)
+    const assetBalances = await api.loadBalance(api.getBalanceFrom, { net, address })
 
     return {
       [ASSETS.NEO]: assetBalances.assets.NEO.balance.toString(),
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -40,8 +40,8 @@ async function getBalances ({ net, address, tokens }: Props) {
     const assetBalances = await api.loadBalance(api.getBalanceFrom, { net, address })
 
     return {
-      [ASSETS.NEO]: assetBalances.assets.NEO.balance.toString(),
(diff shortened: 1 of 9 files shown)


app/actions/balancesActions.js
```diff
@@ -37,11 +37,11 @@ async function getBalances ({ net, address, tokens }: Props) {
...
   // asset balances
   promises.push((async () => {
-    const assetBalances = await api.neonDB.getBalance(net, address)
+    const assetBalances = await api.loadBalance(api.getBalanceFrom, { net, address })
 
     return {
-      [ASSETS.NEO]: assetBalances.assets.NEO.balance.toString(),
-      [ASSETS.GAS]: assetBalances.assets.GAS.balance.round(COIN_DECIMAL_LENGTH).toString()
+      [ASSETS.NEO]: assetBalances.balance.assets.NEO.balance.toString(),
+      [ASSETS.GAS]: assetBalances.balance.assets.GAS.balance.round(COIN_DECIMAL_LENGTH).toString()
```
mhuggins: Does the shape of this endpoint differ between neoscan and neonDB?  Does one have a `balance` key and the other doesn't?

app/core/formatters.js
```diff
@@ -34,8 +34,10 @@ export const formatBalance = (
   return formatGAS(balance, shortDisplay)
 }
 
-export const toFixedDecimals = (value: ValueType, decimals: number): string =>
-  toBigNumber(value).toFixed(COIN_DECIMAL_LENGTH)
+export const toFixedDecimals = (value: ValueType, decimals: number): string => {
+  const decimalLength = (decimals || decimals === 0) ? decimals : COIN_DECIMAL_LENGTH
```
mhuggins: Nice, killed a bug in the process!

app/modules/api/batch.js
```diff
@@ -7,7 +7,7 @@ import {
 } from '../../values/api'
 
 type State = {
-  batch: true,
+  batch: boolean,
```
mhuggins: I had this as `true` because a state should never come throw here with `batch: false` FWIW.

app/actions/balancesActions.js
dvdschwrtz: I can take a look again but I am confused as to why yours worked before. Both endpoints should return this same format. Will take another look

mhuggins: You know, actually it might have the data under both `assets` and `balance.assets`.  I remember seeing duplicate data structure coming back from neonDB when I added this.

snowypowers: The duplicate data structure was when i was transiting to the new model and retained the old models for a while before deprecation. The old code looks correct though.

dvdschwrtz: Thanks @snowypowers :)

PR #808, opened 2018-03-05
Claim

dvdschwrtz:

claiming gas working on dev branch

allows user to claim both spent and unspent tx outputs 

by sending NEO to yourself, then polling with getRawTransaction to see when the transaction has been processed before claiming gas.

manual testing and update the tests to pass

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Claim/Claim.jsx
+++ b/app/containers/Claim/Claim.jsx
@@ -30,6 +30,9 @@ export default class Claim extends Component<Props> {
     }
     if (finalizeClaim && !prevProps.finalizeClaim) {
       doClaimNotify()
+      if (this.intervalId) {
+        clearInterval(this.intervalId)
+      }
     }
   }
 
--- a/app/containers/Claim/Claim.jsx
+++ b/app/containers/Claim/Claim.jsx
@@ -10,17 +10,32 @@ type Props = {
   doClaimNotify: Function,
   setClaimRequest: Function,
   doGasClaim: Function,
+  checkClaimStatus: Function,
   claimRequest: boolean,
   disableClaimButton: boolean,
   claimAmount: string,
+  finalizeClaim: boolean
 }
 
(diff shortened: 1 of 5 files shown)


app/containers/Claim/Claim.jsx
```diff
@@ -10,17 +10,35 @@ type Props = {
...
+    const { claimRequest, doClaimNotify, setClaimRequest, finalizeClaim, checkClaimStatus } = this.props
     if (claimRequest && !prevProps.claimRequest) {
       setClaimRequest(false)
+      this.intervalId = setInterval(checkClaimStatus, POLL_FREQUENCY)
+    }
+    if (finalizeClaim && !prevProps.finalizeClaim) {
       doClaimNotify()
+      if (this.intervalId) {
+        clearInterval(this.intervalId)
+      }
```
mhuggins: Since these three lines are repeated twice (here and in `componentWillUnmount`), it seems like we could put them into a function called `stopPolling` or something like that.

app/modules/claim.js
```diff
@@ -134,17 +150,48 @@ export const doGasClaim = () => async (
...
+  getState: GetStateType
+) => {
+  const state = getState()
+  const net = getNetwork(state)
+  const txid = getTransactionId(state)
+  const endpoint = await api.loadBalance(api.getRPCEndpointFrom, { net })
+  const response = await rpc.Query.getRawTransaction(txid).execute(endpoint)
+  if (response.result && response.result.blockhash && response.result.confirmations) {
+    dispatch(setTransactionId(''))
+    dispatch(setFinalizeClaim(true))
```
mhuggins: This is the only place we're calling `setFinalizeClaim`.  It seems like we need to call `setFinalizeClaim(false)` somewhere in order to reset it, otherwise only one claim per app launch will work (if I'm interpreting correctly).

PR #814, opened 2018-03-07
Reverted load balancing change for transaction history

### LINKED ISSUE #813: dev branch: GAS and NEO errors
At first, I thought this was a Ledger-specific issue (#798), but after further testing, it seems to happen with saved wallets, too.

The issue happened initially when I tried to login with my Ledger. Upon clicking the "Use Ledger Nano S" button, the UI turned white, and there were multiple errors in the console:

```
TRANSACTION_HISTORY request action failed. Error: new BigNumber() not a number: GAS
at raise (http://localhost:3000/dist/bundle.js:32508:25)
    at http://localhost:3000/dist/bundle.js:32496:33
    at new BigNumber (http://localhost:3000/dist/bundle.js:31511:67)
    at toBigNumber (http://localhost:3000/dist/bundle.js:15311:10)
    at http://localhost:3000/dist/bundle.js:2047:177
    at Array.map (native)
    at _callee$ (http://localhost:3000/dist/bundle.js:2039:61)
    at tryCatch (http://localhost:3000/dist/bundle.js:128492:40)
    at Generator.invoke [as _invoke] (http://localhost:3000/dist/bundle.js:128730:22)
    at Generator.prototype.(anonymous function) [as next] (http://localhost:3000/dist/bundle.js:128544:21)
```

```
Uncaught TypeError: Cannot read property 'NEO' of null
    at mapAccountDataToProps (index.js:30)
    at Function.mapStateToProps [as mapToProps] (withData.js:38)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at Function.detectFactoryAndVerify (wrapMapToProps.js:52)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at handleFirstCall (selectorFactory.js:26)
    at pureFinalPropsSelector (selectorFactory.js:74)
    at Object.runComponentSelector [as run] (connectAdvanced.js:26)
    at Connect.initSelector (connectAdvanced.js:178)
    at new Connect (connectAdvanced.js:119)
```

I've seen the same issue occur when using a saved wallet, but it doesn't happen every time. I've seen it happen when switching from MainNet to TestNet, but that also works most of the time.

I'm only seeing the issue about 33% of the time.

I don't know if this correlation is causation yet, but it seems that the issue occurs when using Neoscan for requests like:

https://api.neoscan.io/api/main_net/v1/get_address_neon/XX
https://api.neoscan.io/api/main_net/v1/get_claimable/XX

whereas it works when using Neon wallet DB:

http://api.wallet.cityofzion.io/v2/address/history/XX
http://api.wallet.cityofzion.io/v2/address/balance/XX
http://api.wallet.cityofzion.io/v2/address/claims/XX

It seems that this Neoscan API call is used in both cases (success and failure scenarios):

https://api.neoscan.io/api/main_net/v1/get_balance/XX

Let me know if there are more details you need from me!

mhuggins:
Fixes #813.

Load balancing was added in several spots recently, but load balancing the transaction history is not working due to a bug in neon-js.  Specifically, the neonDB and neoscan API's return different values, and the neoscan API sometimes returns bad data.

I reverted the load balancing of transaction history to only fetch from neonDB based upon @snowypowers advice:

> i went with the alternative implementation  as it would allow you to just iterate the balance object for easy retrieval of the amts
> anyway dont wait on this change, i want to add more tests to make sure both sides sync up properly. will probably take a day more
> on a side note, i notice that neoscan doesnt have all transactions that neonDB shows and it also takes 8s just to pull the data.

Smoke tested.

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/app/actions/transactionHistoryActions.js
+++ b/app/actions/transactionHistoryActions.js
@@ -14,7 +14,7 @@ type Props = {
 export const ID = 'TRANSACTION_HISTORY'
 
 export default createRequestActions(ID, ({ net, address }: Props = {}) => async (state: Object) => {
-  const transactions = await api.loadBalance(api.getTransactionHistoryFrom, { net, address })
+  const transactions = await api.neonDB.getTransactionHistory(net, address)
 
   return transactions.map(({ NEO, GAS, txid }: TransactionHistoryType) => ({
     txid,


app/actions/transactionHistoryActions.js
```diff
@@ -14,7 +14,7 @@ type Props = {
 export const ID = 'TRANSACTION_HISTORY'
 
 export default createRequestActions(ID, ({ net, address }: Props = {}) => async (state: Object) => {
-  const transactions = await api.loadBalance(api.getTransactionHistoryFrom, { net, address })
+  const transactions = await api.neonDB.getTransactionHistory(net, address)
```
brianlenz: Is this the only API that has issues with neoscan? I wasn't able to single out the point-of-failure in my tests, but I'm sure you guys were able to confirm that...

mhuggins: Good question.  I'll verify after merging this change in.

brianlenz: Cool, I'm testing now, as well...

brianlenz: Looks solid to me! I haven't been able to repro the issue... 👍

PR #816, opened 2018-03-08
Hide zero-balance tokens in send modal

### LINKED ISSUE #746: Only show NEP5 tokens with balances
As the number of NEP5 tokens grows in the NEO ecosystem, it makes sense to only list the NEP5 tokens that have a balance in your wallet. Neon would still need to check the balance for every registered NEP5 token, but it would be a nice optimization to only list the tokens that have a non-zero balance. This would be similar to how neotracker.io shows NEP5 token balances, only listing those that are applicable.

mhuggins:
Fixes #746.

In my [previous PR](https://github.com/CityOfZion/neon-wallet/pull/815), I updated the token list on the dashboard to hide zero balances.  This updates the send modal to exclude tokens with a zero balance from the dropdown.  NEO & GAS are always available as options.

I checked the symbol to keep NEO & GAS and check the balance to keep any non-zero amount.

Opened the send modal for various addresses that don't have NEO/GAS or only have 1-2 types of tokens.

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/SendModal/AddRecipientDisplay.jsx
+++ b/app/components/Modals/SendModal/AddRecipientDisplay.jsx
@@ -2,12 +2,14 @@
 import React from 'react'
 import classNames from 'classnames'
 import ArrowForward from 'react-icons/lib/md/arrow-forward'
+import { pickBy } from 'lodash'
 
 import Button from '../../Button'
 import NumberInput from '../../NumberInput'
 import AddressInput from '../../Inputs/AddressInput'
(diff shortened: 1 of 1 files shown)


mhuggins:
@brianlenz Thanks for calling that out.  I think we should track that as a separate issue even though it has a visible impact on the send modal.  The real issue is, like you said, that we should probably show a loading spinner when switching between MainNet and TestNet.

mhuggins:
Opened #826 to track that issue.

PR #828, opened 2018-03-09
Fixed issue with missing asset key in balances response

mhuggins:
N/A

Sometimes when we get an API response back for asset balances, it *only* includes a GAS key *or* a NEO key; not both.  In such circumstances, we can assume the balance is '0'.  Without this change, an error is raised, and the app just shows a white screen.

I short-circuited to return '0' for either asset balance if the key is not set.

Logged into an account that has only ever had GAS and no NEO.

N/A

@snowypowers This might be worth fixing on the neon-js side of things.

- [ ] Unit tests written?

### DIFF
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -38,11 +38,13 @@ async function getBalances ({ net, address, tokens }: Props) {
   // asset balances
   promises.push((async () => {
     const assetBalances = await api.loadBalance(api.getBalanceFrom, { net, address })
+    const { assets } = assetBalances.balance
 
-    return {
-      [ASSETS.NEO]: assetBalances.balance.assets.NEO.balance.toString(),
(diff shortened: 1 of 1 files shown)


snowypowers:
this is **intended**. The `Balance` structure is as such:
```
Balance {
  net: string
  address: string
  assetSymbols:  string[]
  assets: { [index: string]: AssetBalance }
  tokenSymbols: string[]
  tokens: { [index: string]: number }
}
```

If the key does not exist in `assetSymbols`, it will not exist in `assets`. This structure:

1. Accomodates future possible assets (and tokens) added so our object doesnt balloon in size.
2. Allows us to isolate assets within an object allowing for easy iteration.
3. prevents collision of symbols between assets and tokens.

mhuggins:
That makes sense, thanks @snowypowers.  I'm going to leave this change as-is for now, but it sounds like in the future we may want to loop over the `assetSymbols` object.

snowypowers:
Yeah there isn't much use right now because there isnt any new assets being created but it is a feature built into the system.

PR #854, opened 2018-03-12
Fixed private key login to only permit private key or WIF

mhuggins:
N/A

Users are currently able to login with their address or public key instead of WIF or private key.  Nothing can actually be done other than looking at publicly accessible data, but we should not incorrectly authenticate in the first place.

I tested that the string provided to the authentication function is a WIF or private key.

I smoke tested and added tests.

N/A

N/A

- [x] Unit tests written?

### DIFF
--- a/__mocks__/neon-js.js
+++ b/__mocks__/neon-js.js
@@ -87,7 +87,10 @@ neonjs.wallet = {
   Account: jest.fn(() => { return { address } }),
   getVerificationScriptFromPublicKey: jest.fn(() => scriptHash),
   isAddress: jest.fn(() => true),
-  isNEP2: jest.fn(() => true)
+  isNEP2: jest.fn(() => true),
+  isWIF: jest.fn(() => false),
+  isPrivateKey: jest.fn(() => false),
+  isPublicKey: jest.fn(() => false)
 }
 
 module.exports = neonjs
--- a/__tests__/actions/authActions.test.js
+++ b/__tests__/actions/authActions.test.js
@@ -1,0 +1,132 @@
+import { wallet } from 'neon-js'
+
+import { wifLoginActions } from '../../app/actions/authActions'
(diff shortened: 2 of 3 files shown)


__mocks__/neon-js.js
```diff
@@ -87,7 +87,10 @@ neonjs.wallet = {
   Account: jest.fn(() => { return { address } }),
   getVerificationScriptFromPublicKey: jest.fn(() => scriptHash),
   isAddress: jest.fn(() => true),
-  isNEP2: jest.fn(() => true)
+  isNEP2: jest.fn(() => true),
+  isWIF: jest.fn(() => false),
+  isPrivateKey: jest.fn(() => false),
+  isPublicKey: jest.fn(() => false)
```
mhuggins: I think these global stubs are a bad idea long term and we should work to get rid of them, favorite stubbing in tests where needed.

__tests__/actions/authActions.test.js
```diff
@@ -0,0 +1,132 @@
+import { wallet } from 'neon-js'
+
+import { wifLoginActions } from '../../app/actions/authActions'
+
+describe('authActions', () => {
+  describe('wifLoginActions', () => {
+    describe('request', () => {
+      const wif = 'KxB52D1FGe5xBn6YeezNwj7grhkHZxq7bv2tmaCPoT4rxApMwMvU'
+      const address = 'AM22coFfbe9N6omgL9ucFBLkeaMNg9TEyL'
```
snowypowers: This address doesn't correspond to its WIF or private key. Intended?

mhuggins: It doesn't matter, but it was not intended.  I'll update regardless.

mhuggins: Fixed.

PR #857, opened 2018-03-12
Fixed balance calculation for non-USD currencies

### LINKED ISSUE #843: [Bug] Price always show USD value
The price functionality always reflect the USD value even if you change it to any other currency. 

The symbol changes but the number doesn't. So I see my 1 NEO at 95 USD and also at 95 JPY.

mhuggins:
Fixed #843.

The price functionality always reflects the USD value even if you change it to any other currency.  The symbol changes, but the number doesn't.  For example, you might see 1 NEO at 95 USD and also at 95 JPY.

I separated the call for prices data to occur after settings data have been loaded.  This is necessary since we don't know the user's desired currency until settings are loaded, and the currency is needed to make the token prices fetch request.

I loaded up the dashboard with the currency set to USD and not USD, confirming that it shows not just different symbols but also different/correct values.

N/A

N/A

- [x] Unit tests written?

### DIFF
--- a/__tests__/components/App.test.js
+++ b/__tests__/components/App.test.js
@@ -14,7 +14,19 @@ const initialState = {
   api: {
     APP: {
       batch: true,
-      mapping: ['NETWORK', 'PRICES', 'SETTINGS']
+      mapping: ['ACCOUNTS', 'BLOCK_HEIGHT', 'SETTINGS']
+    },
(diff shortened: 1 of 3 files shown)


snowypowers:
Branch tested on windows. Tried 3 different currencies and worked as expected.

PR #858, opened 2018-03-13
upgrade neon-js to 3.4.4

snowypowers:
- Fixes sales by ordering the signatures correctly
- Uses the new transaction format for past transactions

As this changes the format of transactions returned, it probably breaks most of the transaction related tests. The alternative is to use `3.3.3` which contains only the signature fix. Close this if you do that.

On a side note, I tried out the price PR and it doesn't seem to be working.


snowypowers:
Closing as we will use 3.3.3 instead.

PR #864, opened 2018-03-14
Replace global neon-js test mock with local stubbing

mhuggins:
N/A

The global stubbing of neon-js introduces problems when you want to use aspects of neon-js as is, without stubbing.  It also depends on a convoluted & time consuming process of updating a complex object whenever the neon-js dependency changes.  Instead of a global dependency stub, I suggest that we only mock neon-js API calls for specific tests that need them.

I removed the global mock for neon-js and replaced it with local mocks specific to any individual test that needs it.

In addition to adding mocks to tests that need them, it also replaces axios-mock-adaptor with nock, which allows us to disable all network requests.  This will help to quickly identify any calls that need to be stubbed for new tests being added.

I ran the test suite and fixed any test issues.

N/A

@snowypowers I attempted to mock neon-js API calls like this:

```js
import { api } from 'neon-js'
jest.spyOn(api, 'getWalletDBHeightFrom').mockImplementation(...)
```

I received an error that `getWalletDBHeightFrom` is a getter and must be mocked as such, so I changed to this:

```js
import { api } from 'neon-js'
jest.spyOn(api, 'getWalletDBHeightFrom', 'get').mockImplementation(...)
```

That in turn raised an error that the getter is not `configurable` (i.e.: it is set to `configurable: false` instead of `configurable: true`).  As such, I had to work around this by freezing load balancing, and stubbing neoscan instead:

```js
jest.spyOn(api.neoscan, 'getWalletDBHeight').mockImplementation(...)
```

Is there a way we can update neon-js to allow stubbing on the `api` object directly?

- [x] Unit tests written?


snowypowers:
thats weird... its not a getter...

mhuggins:
@snowypowers Could it be something with the way webpack compiles it?  I'm not really sure.

snowypowers:
Looks like its babel's way of interpreting `export * from ./file`

https://stackoverflow.com/questions/36510224/es6-re-exported-values-are-wrapped-into-getter

mhuggins:
Lame.  I guess we could just export our own object.  I wonder if there's a way to avoid that behavior otherwise.

PR #869, opened 2018-03-14
Update README.md with extra NEP5 tokens

robmoorman:
Just to give more insight in newly supported NEP5 tokens at a glance

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
 * Import/Export wallet accounts (NEP6 Standard)
 * View balance
 * View prices for GAS and NEO in multiple currencies
-* Send GAS, NEO, and certain NEP5 tokens (RPX, DBC, QLC, IAM, ONT and THOR)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,7 @@
 * Import/Export wallet accounts (NEP6 Standard)
 * View balance
 * View prices for GAS and NEO in multiple currencies
-* Send GAS, NEO, and certain NEP5 tokens (RPX, DBC, QLC)
+* Send GAS, NEO, and certain NEP5 tokens (RPX, DBC, QLC, IAM, ONT and THOR)
```
mhuggins: Thanks for the contribution!  I would suggest we remove the list of tokens (since any NEP5 token can be added by users now) and just change this line to:

> Send GAS, NEO, and any NEP5 token

robmoorman: Hi @mhuggins that would be a better pick indeed!

PR #870, opened 2018-03-14
JS Dev updates

robmoorman:
Updated docs to latest dependency versions in use
Added nvmrc file to support correct nodejs version


snowypowers:
@mhuggins are we looking to merge this? Been sitting for quite long and even the updates are really outdated. Would be nice to merge it and follow it up with a new PR that updates the versions to the correct ones being used (electron version)

Issue #878, opened 2018-03-16
Block people from sending to known phishing addresses


dvdschwrtz:
We have this repo with a list of blocked addresses https://raw.githubusercontent.com/CityOfZion/phishing/master/blockedAddresses.json. In neon wallet, we need to do a fetch request for this data from github and make sure that no one can send to any addresses on this list. It should notify them through the UI that they have input a phishing address and block them from proceeding

PR #887, opened 2018-03-17
Blocked people from sending to addresses listed in CityOfZion/phishing

Taconut:
Fixes issue #878 

Up until this commit, people may have unintentionally sent tokens to malicious addresses

The first time someone tries to send tokens to an address, neon-wallet fetches the contents of the JSON file at the root of CityOfZion/phishing and compares its contents against the address. If there is a match, the transaction is halted and a message is displayed notifying the user that they just tried to send tokens to a phishing address. Subsequently, the application uses a cached copy of the JSON file.

I tested "AasqbuVKaCARqoWfp64ygM5fxYBqmMMgpq" as my malicious address and "Adr3XjZ5QDzVJrWvzmsTTchpLRRGSzgS5A" as my non-malicious address

I added node-fetch as a dependency. I was going to add electron-fetch instead, but it was acting up and I didn't want to risk it.

### DIFF
--- a/app/components/Modals/SendModal/SendModal.jsx
+++ b/app/components/Modals/SendModal/SendModal.jsx
@@ -7,7 +7,7 @@ import AddRecipientDisplay from './AddRecipientDisplay'
 import ConfirmDisplay from './ConfirmDisplay'
 import withAddressCheck from './withAddressCheck'
 
-import { validateTransactionBeforeSending, getTokenBalancesMap } from '../../../core/wallet'
+import { validateTransactionBeforeSending, getTokenBalancesMap, isInBlacklist } from '../../../core/wallet'
 import { ASSETS } from '../../../core/constants'
 import { toBigNumber } from '../../../core/math'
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -81,6 +81,7 @@
     "isomorphic-fetch": "2.2.1",
     "lodash": "4.17.4",
     "neon-js": "git+https://github.com/cityofzion/neon-js.git#3.3.3",
+    "node-fetch": "^2.1.1",
```
mhuggins: We already have `axios` in this repo -- we should try to avoid bringing in extra packages to prevent bloat.

app/core/wallet.js
```diff
@@ -1,12 +1,25 @@
...
 
 import { ASSETS } from './constants'
 import { toBigNumber } from './math'
 
 const MIN_PASSPHRASE_LEN = 4
 
+let addressBlacklist: Array<string> | null = null
+export const isInBlacklist = (address: string): Promise<boolean> => {
+  if (addressBlacklist !== null) return Promise.resolve(addressBlacklist.includes(address))
+  return fetch('https://raw.githubusercontent.com/CityOfZion/phishing/master/blockedAddresses.json')
```
mhuggins: We could write this with axios and async/await syntax in a pretty direct manner:

```js
import axios from 'axios'

export const isInBlacklist = async (address: string): Promise<boolean> => {
  if (addressBlacklist === null) {
    const { data } = await axios.get('https://raw.githubusercontent.com/CityOfZion/phishing/master/blockedAddresses.json')
    addressBlacklist = data
  }
  return addressBlacklist.includes(address)
}
```

app/components/Modals/SendModal/SendModal.jsx
```diff
@@ -110,19 +110,25 @@ export default class SendModal extends Component<Props, State> {
...
-    } else {
-      const newBalance = toBigNumber(balances[entry.symbol]).minus(entry.amount).toString()
-
-      this.setState({
-        entries: [...this.state.entries, entry],
-        balances: { ...balances, [entry.symbol]: newBalance },
-        display: DISPLAY_MODES.CONFIRM
-      })
-    }
+    isInBlacklist(entry.address).then(inBlacklist => {
```
mhuggins: FYI using async/await syntax here can remove a layer of nesting and improve the readability.  We've been favoring this syntax over raw promises throughout the codebase.

You would first need to change the method signature to include `async`, i.e.:

```js
handleConfirmAddRecipient = async (entry: SendEntryType) => { ... }
```

Then the skeleton of this function simply becomes:

```js
if (await isInBlacklist(entry.address)) {
  ...
} else {
  ...
}
```

Taconut: Unfortunately, making that function async/await removes access to the "this" variable. I get ``Cannot read property 'showErrorNotification' of undefined`` or ``ReferenceError: _this11 is not defined``. There's a weird workaround to this in withAddressCheck.jsx, but applying it breaks the render function. I'm going to leave it up to you what to do with this function and commit the other changes in the meantime. Let me know what I should do with this.

mhuggins: Ahh, that should be fixable by using `async function(...) { ... }` instead of `async (...) => { ... }`.  Regardless, I can probably merge this as-is and look into it afterwards.

mhuggins:
Thought I'd follow up and mention that the reason making that function couldn't be changed to use `async` was due to [a bug in react-hot-loader](https://github.com/gaearon/react-hot-loader/issues/391#issuecomment-353867009).  I upgraded in #914 to fix it.

PR #895, opened 2018-03-20
Added e2e tests for creating new wallet

drptbl:
Tried to keep it as simple as possible. I think that other checks should be done in separate test cases.
I've also added `-s` (running tests serially) to `test:e2e` script based on https://github.com/avajs/ava/issues/1060 to prevent flakiness and limit electron instances while running tests.

Let me know if there's anything to change/improve.
Feedback is very welcome.

### DIFF
--- a/__e2e__/create-new-wallet.e2e.js
+++ b/__e2e__/create-new-wallet.e2e.js
@@ -1,0 +1,70 @@
+import { Application } from 'spectron'
+import path from 'path'
+import test from 'ava'
+
+let electronPath = path.join(__dirname, '..', 'node_modules', '.bin', 'electron')
+if (process.platform === 'win32') {
+  electronPath += '.cmd'
+}
+const appPath = path.join(__dirname, '..')
+
+const app = new Application({
+  path: electronPath,
+  args: [appPath]
+})
+
+test.before(async () => {
+  await app.start()
+  await app.client.waitUntilWindowLoaded()
+})
+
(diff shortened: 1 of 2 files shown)


mhuggins:
Looks like this is still failing.  Here's a couple thoughts:

* Do we need to change the test setup to only happen once rather than per file?  (Not sure how to go about this, just a thought.
* The [spectron/ava docs](https://github.com/electron/spectron#with-ava) show that they're using `t.context.app` instead of an `app` variable.  Not sure if this might have an impact, but is it worth considering?

drptbl:
Okay, so..
That's another issue which came out on CI environment.
Image which is used for running tests were missing some display packages which are needed for Chrome Driver to run properly.
After adding them everything works as it should now - I've tested it here:
https://circleci.com/gh/drptbl/neon-wallet/11

But why it's failing now? I have no idea to be honest.
```
/root/neon-wallet/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```
I didn't touch coveralls.

mhuggins:
Odd, I can't kick it off to rerun on Circle either since it's under your username. 😕

drptbl:
It looks like too much unrealted issues to pull request title are coming up. I'm going to close this and divide it to several pull requests to make it all clean. @mhuggins Thank you 👍

PR #897, opened 2018-03-20
Add Switcheo token scripthash

RavenXce:
N/A

This adds the Switcheo token which has concluded crowdsale on 17 Mar.

I added the scripthash and token symbol for SWH.

I opened the main window and verified that the script hash pulled up the correct token info and amounts.

No

No

- [ ] Unit tests written?

### DIFF
--- a/app/core/constants.js
+++ b/app/core/constants.js
@@ -77,7 +77,8 @@ export const TOKENS = {
   IAM: '891daf0e1750a1031ebe23030828ad7781d874d6',
   ONT: 'ceab719b8baa2310f232ee0d277c061704541cfb',
   THOR: '67a5086bac196b67d5fd20745b0dc9db4d2930ed',
-  CGE: '34579e4614ac1a7bd295372d3de8621770c76cdc'
+  CGE: '34579e4614ac1a7bd295372d3de8621770c76cdc',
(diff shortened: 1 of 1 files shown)


app/core/constants.js
```diff
@@ -77,7 +77,8 @@ export const TOKENS = {
   IAM: '891daf0e1750a1031ebe23030828ad7781d874d6',
   ONT: 'ceab719b8baa2310f232ee0d277c061704541cfb',
   THOR: '67a5086bac196b67d5fd20745b0dc9db4d2930ed',
-  CGE: '34579e4614ac1a7bd295372d3de8621770c76cdc'
+  CGE: '34579e4614ac1a7bd295372d3de8621770c76cdc',
+  SWH: '78e6d16b914fe15bc16150aeb11d0c2a8e532bdd'
 }
 
 export const ENDED_ICO_TOKENS = ['DBC', 'RPX', 'QLC', 'RHT']
```
mhuggins: Since you mention that Switcheo's crowdsale concluded, it seems that we should probably include it in this array as well.

RavenXce: Yes, we should do that. We meant to merge before the crowdsale. Thanks for the pointing it out!

PR #901, opened 2018-03-20
Add e2e tests for creating new wallet and run tests serially

drptbl:
Tried to keep it as simple as possible. I think that other checks should be done in separate test cases.
I've also added `-s` (running tests serially) to `test:e2e` script based on https://github.com/avajs/ava/issues/1060 to prevent flakiness and limit electron instances while running tests.

Let me know if there's anything to change/improve.
Feedback is very welcome.

### DIFF
--- a/__e2e__/create-new-wallet.e2e.js
+++ b/__e2e__/create-new-wallet.e2e.js
@@ -1,0 +1,70 @@
+import { Application } from 'spectron'
+import path from 'path'
+import test from 'ava'
+
+let electronPath = path.join(__dirname, '..', 'node_modules', '.bin', 'electron')
+if (process.platform === 'win32') {
+  electronPath += '.cmd'
+}
+const appPath = path.join(__dirname, '..')
+
+const app = new Application({
+  path: electronPath,
+  args: [appPath]
+})
+
+test.before(async () => {
+  await app.start()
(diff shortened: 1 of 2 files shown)


__e2e__/create-new-wallet.e2e.js
```diff
@@ -0,0 +1,70 @@
...
+test.serial('should show wallet passphrase', async t => {
+  t.true(await app.client.isVisible('#newWallet .keyListItem:nth-child(1) .key'))
+  const passphraseText = await app.client.getText('#newWallet .keyListItem:nth-child(1) .key');
+  t.truthy(passphraseText.length)
+})
+
+test.serial('should show wallet public address', async t => {
+  t.true(await app.client.isVisible('#newWallet .keyListItem:nth-child(3) .key'))
+  const publicAddressText = await app.client.getText('#newWallet .keyListItem:nth-child(3) .key');
+  t.truthy(publicAddressText.length)
```
mhuggins: We could probably leverage `neon-js` to validate that values like public key, address, etc. are being set correctly.  This is still a big improvement over the current lack of test, so I'm going to merge. 😄

drptbl:
Yeah, I'm going to work on general utils file right now 😄 .

PR #911, opened 2018-03-23
feat(ci): automated builds upon tagging

snowypowers:
This adds the ability to use CI to build upon tagging. I have it setup for linux and windows builds.

## How

1. We will have to tag the build beforehand. Instead of using the release page to tag, we use cmd. Assuming we are on `master` on the commit that we want to release:
```
git tag 0.2.3
git push upstream 0.2.3 --no-verify
```

This will push the tag up and CircleCi should pick this up and trigger the builds.

The builds can be found by going to Workflows -> neon-wallet. The dist folder is uploaded as artifacts. Unfortunately, I do not see anyway to automate the attachement of the builds to the release draft so we will have to do this ourselves.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,16 @@
 version: 2
+
 references:
   container_config: &container_config
     docker:
       - image: electronuserland/builder:wine-chrome
+        environment:
+          TARGET_ARCH: x64
+    working_directory: ~/neon-wallet
+
+  win_config: &win_config
+    docker:
+      - image: electronuserland/builder:wine
         environment:
           TARGET_ARCH: x64
     working_directory: ~/neon-wallet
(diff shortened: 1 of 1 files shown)


mhuggins:
Is the goal/benefit of this so that we can coordinate what commit to build the release from?  Is there anything more that is good to know about it?

mhuggins:
You might need to rebase this against the current `dev` branch to make your CI build pass.

snowypowers:
The goal is to automate the builds so that you can have the windows and linux builds without needing us to help build. CircleCI builds and you grab the exe, deb and appImage from the CircleCI site.

snowypowers:
the test is failing because the trigger is on my branch and not on the coz branch, thus failing at the coveralls step...

drptbl:
@snowypowers I had exactly the same issue with coveralls and I couldn't find a way to fix it. Is there any way?

snowypowers:
@drptbl i just did it. I think setting it to private solves it.

PR #926, opened 2018-03-31
Update input styles to match v2 designs

mhuggins:
N/A

This updates the designs for all inputs (text, password, number, asset, & address) to match the new styles provided by @asalib.

* Organized all inputs under the `components/Inputs` folder.
* Created a base `TextInput` component that all other inputs are derived from.
* Added text highlighting to `SelectInput` for search string matches.
* Updated `AddressInput` to remove save/delete functionality.  (This will be reimplemented as part of the send flow.)

v2-inputs

Since it's mostly visual changes, I just smoke tested.

N/A

N/A

- [x] Unit tests written?


comountainclimber:
👍 nice job @mhuggins will work this into the Home container

mhuggins:
@comountainclimber Hope I didn't step on your toes too much.  Let me know if it's too much trouble, and I can help coordinate on combining what you've done.

Issue #890, opened 2018-03-17
Import Whitescreen + token not showing up in 0.2.2 but in 0.1.4


goldenstar2k:
i would like to post a bug using the latest 0.2.2. I created 10 neo addresses, exported them into a json. I uninstalled my 0.1.4 + i deleted the neon folder remaining and installed the latest 0.2.2. After importing the json into the neon client the screen remained white. I closed the neon wallet reopened and logged into my account with all my tokens. But its not showing up the tokens. On the 0.1.4 the import of the json and showing up of tokens worked flawless.

swolfaardt:
Same problem did you have any tokens? saw someones post saying if it is <0 wont show?

drptbl:
Looks like a duplicate of #851.
Also it was already fixed with latest wallet release (0.2.3).

> Bug Fixes:
> ...
> Fixed wallet recovery resulting in white screen (no longer requires app restart).

Regarding tokens.. it's not a bug. You have to add script hashes for tokens first before you can see them. Go to "Manage Neon settings" => "Manage Tokens" and add them there. After that, login to your wallet and you should see your tokens.

PR #946, opened 2018-04-07
fix array slice bug of accounts array that has lenght 0

Lionardo:
#890
It fixes the white screen issue when loading recovery keys in the settings.

It doesn't slice the accounts array if there are no accounts, this doesn't throw an error.

Inline If with Logical && Operator to check the accounts props length.

tested in production locally on my mac. I didn't test it on windows.

No.

No.

- [x] Unit tests written?

> Test Suites: 23 passed, 23 total
Tests:       89 passed, 89 total
Snapshots:   13 passed, 13 total
Time:        31.52s
Ran all test suites.
✨  Done in 37.43s.

### DIFF
--- a/app/containers/Settings/Settings.jsx
+++ b/app/containers/Settings/Settings.jsx
@@ -214,27 +214,29 @@ export default class Settings extends Component<Props, State> {
               ))}
             </select>
           </div>
-          <div className='settingsItem'>
-            <div className='itemTitle'>Saved Wallet Accounts</div>
-            {map(accounts, account => {
-              return (
-                <div className='walletList' key={`wallet${account.key}`}>
-                  <div className='walletItem'>
-                    <div className='walletName'>{account.key.slice(0, 20)}</div>
-                    <div className='walletKey'>{account.label}</div>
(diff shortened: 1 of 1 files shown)


mhuggins:
I'm not sure I'm following the change in logic here.  You're adding a check for `accounts.length > 0`, but it seems like the `map` call will never enter the function if `length === 0` in the first place.  What does the `accounts` data structure look like when you're running into this issue?

Lionardo:
accounts data structure is an empty array []. Apparently the map function is entered even if you have an empty accounts array. 
screen shot 2018-04-07 at 09 42 06

drptbl:
> It fixes the white screen issue when loading recovery keys in the settings.
This issue is already fixed with latest wallet release (0.2.3) - just re-tested it for second time.
import_whitescreen

mhuggins:
Thanks for verifying @drptbl.  Going to mark as closed/fixed.

Lionardo:
alright, the only change I would make is to remove the phrase "saved wallets accounts", if there are none saved.

PR #955, opened 2018-04-10
Add prettier for beautiful code

evgenyboxer:
None

Adds prettier (https://prettier.io/), a very popular opinionated code formatter, that turns the code to beautiful and readable.

I added a `format` script that will format the files based on the rules I put it .prettierrc.
This is also integrated with eslint, so you get the lint errors.

App runs, tests/lint/flow passes.


mhuggins:
@evgenyboxer Thanks for this!  I'm thinking that it might be better to do this as part of the `design-v2` branch though, because if I merge this into `dev`, I'm going to have a hell of a time reconciling the differences getting those changes in.  Does that make sense/seem reasonable?  (I'm assuming you just ran a command to automatically fix everything, and that you didn't actually go through and make all these changes by hand!)

evgenyboxer:
Sure thing! np. Will make a PR for `design-v2` branch.
Yes, I just ran the command that auto formats this.

comountainclimber:
@mhuggins @evgenyboxer I already created a `.prettierrc` file in the `design-v2` branch that matches the eslint spec. Please be aware the eslint config in neon wallet currently has an undesirable rule that is non configurable with prettier and that is our use of single quotes in jsx

Issue #731, opened 2018-02-18
Make version more visible


michaelf318is:
It is hard to determine what the version of neowallet is without logging into a wallet.

Consider putting Version also in "Manage Neon Settings"  or on the "Home" screen at the bottom.

mhuggins:
Good suggestion.  @asalib What do you think about including the version number on the new login screens before getting to the dashboard?

comountainclimber:
Fixed in `design-v2` branch

PR #957, opened 2018-04-10
Made wallet version more visible

digilifecz123:
[#731](https://github.com/CityOfZion/neon-wallet/issues/731)

This adds wallet version to the top right corner of the app

I added the wallet component to the header

I have tested locally on my macbook.

nope

nope

- [ ] Unit tests written?


comountainclimber:
Do we really need this change? Also how exactly does it "make the wallet version more visible"

mhuggins:
This doesn't fit in with the work we've been doing towards implementing the V2 designs. I appreciate the contribution though!

PR #966, opened 2018-04-15
Cleaned up button styles

mhuggins:
N/A

This cleans up a few things around the new v2 button designs.

* I removed the `primary` prop type since it was redundant (a button will either be primary or secondary, so no need for both).
* The home button had a link-style underline because it was wrapped in a `NavLink`.  I changed the behavior so it uses `history.push` instead, making it a true button.
* I removed the default 15px margin and allowed a custom `className` prop to be passed in.  This will allow the margin to be dynamic via a custom class in the places where it's needed, so we don't force it in all scenarios.
* I added a disabled state (waiting for more designs around this), so that it's clear a button can't be clicked in cases where it's not available.
* Cleaned up styles so we don't have to duplicate definitions and instead leverage SCSS hierarchy.

Updated the button JSX and SCSS files.

Smoke tested the login screens.

N/A

N/A

- [x] Unit tests written?


mhuggins:
@comountainclimber I'm going to merge this to fix some issues found during QA by @drptbl.  Please leave me any retro feedback if you have any, and I'm happy to circle back!

comountainclimber:
@mhuggins this looks great thanks for cleaning this up and simplifying the new button API

PR #1009, opened 2018-04-23
Improved offline support

### LINKED ISSUE #918: Improve offline behavior
I'm guessing the 0.2.0 release is what changed this, but it looks like there has been a regression in offline behavior in Neon as compared to 0.1.4.

Neon 0.2.2 won't load at all without an internet connection now. It just gives a generic text error on a white background:

> Failed to load.

If you happened to have opened up the app with a connection to get to the main screen, then you can actually try to open a wallet.

If you are offline when you try to open a wallet (saved wallet or entering an encrypted/private key), however, then you just end up on a blank white screen and you'll have to restart the app if you want to actually do anything.

Neon 0.1.4 work perfectly in these scenarios. The main app could load while offline and you could also load a wallet while offline. The balances just wouldn't display until an internet connection was established.

For the risk averse, it's reasonable that they may want to operate the wallet in an offline state, particularly for entering of private keys/passphrases. The new requirement of an internet connection in order to use the app at all prevents users from loading and interacting with wallets while offline.

mhuggins:
Fixes #918.

Today, when a user pulls up Neon Wallet with a bad connection, they just see a blank white screen with a "Failed" message.  The user should ideally be able to authenticate, manage the app, and otherwise navigate, even if data cannot be fetched.

This pushes progress loading further down the component stack.  Rather than failing at the top, we're now failing in individual places where we render.  Anything that fails only fails because of the data that it depends on, and is not impacted by other unrelated failed calls.

I smoke tested the components that were related to this change, once with my wifi turned on, and again with it turned off.

N/A

N/A

- [x] Unit tests written?


mhuggins:
Improved the offline support by adding loading/failed states to each panel.

**Loading panels:**
loading-panels

**Failed panels:**
failed-panels

PR #1018, opened 2018-04-28
Login redo

comountainclimber:
authentication v2 🔑 

- New home container
- home container UI and state management
- fixes a bug introduced in the styling of Button.jsx
- new  loader styling
screen shot 2018-04-27 at 6 58 47 pm

authentication in its current state on this branch includes:
- private key auth
screen shot 2018-04-28 at 6 59 52 pm

- encrypted key auth
screen shot 2018-04-28 at 6 53 45 pm

- ledger auth
screen shot 2018-04-28 at 6 51 08 pm

- saved wallet local storage auth
screen shot 2018-04-28 at 6 43 41 pm

Additional changes:
- officially fixes the jsx being single quoted and adds the linting rule...
https://github.com/airbnb/javascript/tree/master/react#quotes

 now neon-wallet is fully compatible with prettier 🔥 💅 
https://github.com/prettier/prettier

- [ ] Unit tests written?


app/containers/Home/Home.jsx
```diff
@@ -1,50 +1,143 @@
...
+          />
+
+          <TextInput
+            className={styles.input}
+            onChange={(e: Object) =>
+              this.setState({ password: e.target.value })
+            }
+            placeholder="Password"
+            type="password"
+          />
```
mhuggins: I would recommend moving the contents of the input container into its own component.  That way, based upon the value of `this.state.option`, you can just render a single component.  You could expose an `onLogin` prop for when the user submits there password (for passphrase login) or any other data (for other login methods).

app/containers/Home/Home.jsx
```diff
@@ -1,50 +1,143 @@
...
+//       <div className={classNames('linkBox', styles.linkBox, styles.linkBoxAlt)}>
+//         Encrypt an existing key
+//       </div>
+//     </Link>
+//     <Link to={ROUTES.SETTINGS}>
+//       <div className={classNames('linkBox', styles.linkBox, styles.linkBoxAlt)}>
+//         Manage Neon settings
+//       </div>
+//     </Link>
+//   </div>
```
mhuggins: It looks like these routes need to go somewhere still.  Will merging this break logging in and seeing the rest of the app?  I think we should at least ensure routing works before merging.  (I can help if needed.)

comountainclimber:
@mhuggins implemented your suggestions thinks its much cleaner and more maintainable that way... unfortunately my particular implementation introduced some type checking fuckery but hopefully it seems reasonable to you

comountainclimber:
@mhuggins failed login attempt modal still appears 👍

comountainclimber:
screen shot 2018-04-28 at 4 05 36 pm

mhuggins:
e2e tests pass locally, best I can tell this is an intermittent failure, possibly due to API/RPC issues at the moment.

Issue #1021, opened 2018-04-28
Disable ability to spam "Refresh" buttons on wallet dashboard


drptbl:
**Wallet version or commit SHA:**
ef95d25

**Environment:**
Win 10 Home (64 bit) Version 1709 Build 16299.371

**Prerequisites:**
At least one wallet imported to the app

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Select your wallet, enter your passphrase and click on "Login" button
4. Click on "Refresh" button several times in a row

Currently user is able to spam "Refresh" buttons causing requests to queue. "Refresh" buttons shouldn't be able to be clicked until requests have finished loading.

**Screenshots/Video:**
![ref_but_spam](https://user-images.githubusercontent.com/8177587/39396656-7395c3ac-4af2-11e8-82d1-7b2fad3fb6e7.gif)

comountainclimber:
should be a trivial fix I can take care of it 👍

comountainclimber:
https://github.com/CityOfZion/neon-wallet/pull/1023

mhuggins:
Code complete, just pending QA.

drptbl:
After re-testing - this issue is still valid.

Changes from linked PR (#1023) applies only to `AssetBalancesPanel.jsx`, but `TokenBalancesPanel.jsx` has been forgotten. Because of that this issue is still valid for "Refresh" button in "Token Balances" section (it's spammable).

I wasn't able to test these changes on "Refresh" button in "Balances" section because it's broken currently (#1079). Will re-test later, after it's fixed.

comountainclimber:
@drptbl good call... I can address all of the above

mhuggins:
Fixed by #1159, pending QA.

drptbl:
Looks good, thanks!

PR #1023, opened 2018-04-29
controlling rate of the refresh button

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1021

Requests can no be made until the current request is complete

Testing locally

- [ ] Unit tests written?


comountainclimber:
that solution is much better thanks 👍

mhuggins:
@comountainclimber Looks like this needs to have single quotes changed to double quotes now that we merged the other PR.

comountainclimber:
@mhuggins updated... still cant get `__tests__/components/Dashboard/AssetBalancesPanel.test.js` to pass either

drptbl:
Issues found which are probably related to this PR:
#1079.

Also re-opened old issue:
#1021. 

It seems that this PR missed `TokenBalancesPanel.jsx` which is related to "Token Balances" section.

PR #1070, opened 2018-05-06
Feature/create wallet

### LINKED ISSUE #976: [v2] Create Wallet
Placeholder story for create wallet screen.

### LINKED ISSUE #977: [v2] Import Wallet
Placeholder story for import wallet screen.

comountainclimber:
Fixes #976.
Fixes #977.

Streamlines the UX for creating or importing a wallet into a single view.

screen shot 2018-05-06 at 3 06 42 pm
screen shot 2018-05-06 at 3 06 48 pm

Manually tested both scenarios numerous times

Wallets are always saved to local storage on creation now
screen shot 2018-05-12 at 6 50 43 pm

- [ ] Unit tests written?


app/components/BackButton/BackButton.jsx
```diff
@@ -0,0 +1,26 @@
...
+  routeTo: string,
+  history: {
+    push: Function
+  }
+}
+
+export default class BackButton extends React.Component<Props> {
+  render = () => {
+    return (
+      <span className={this.props.className} onClick={this.handleClick}>
```
mhuggins: Seems like we can/should just make this a link, e.g.:

```jsx
import { Link } from 'react-router-dom'

render = () => {
  return (
    <Link to={this.props.routeTo}>
      <Arrow className={styles.icon} />
    </Link>
  )
}
```

That will also make it so we can remove the `history` propType.

app/components/Root/Routes.jsx
```diff
@@ -12,19 +12,14 @@ import Contacts from '../../containers/Contacts'
 import AddContact from '../../containers/AddContact'
 import EditContact from '../../containers/EditContact'
 import Settings from '../../containers/Settings'
-import DisplayWalletAccounts from '../../containers/DisplayWalletAccounts'
```
mhuggins: Ahh, I think we may still need to show the account details somehow when the user is creating a wallet, but not importing a wallet.  I don't believe that experience exists in the designs though.  I can help figure out how this should work since Alf's not been available on that front.

app/containers/Home/WithHomeMarkUp.jsx
```diff
@@ -0,0 +1,29 @@
...
+import React, { Component } from 'react'
+import styles from './Home.scss'
+import neonLogo from '../../images/neon-logo-redesign.png'
+
+type Props = {
+  children: React$Node,
+  renderNavigation?: Function
+}
+
+export default class WithHomeMarkUp extends Component<Props> {
```
mhuggins: What do you think about renaming this to something like `HomeLayout` instead?

app/containers/CreateWallet/CreateWallet.jsx
```diff
@@ -15,73 +19,120 @@ type State = {
...
+              </Button>
+            </form>
+          </div>
+        </div>
+      </WithHomeMarkUp>
     )
   }
 }
+
+const OptionButton = ({ active, children, handleClick }: Object) => {
```
mhuggins: 💅 This should go into its own file and then be imported.

app/modules/generateWallet.js
```diff
@@ -162,14 +163,16 @@ export const generateNewWalletAccount = (passphrase: string, passphrase2: string
...
+
         dispatch(hideNotification(infoNotificationId))
         dispatch(newWalletAccount({
           wif: WIF,
           address,
           passphrase,
           encryptedWIF
         }))
-        history.push(ROUTES.DISPLAY_WALLET_KEYS)
+        history.push(ROUTES.HOME)
```
mhuggins: Sorry again for the confusion.  My original comment was just about importing wallets, and we don't need to show the data to the user in that use case.

For this use case, where the user is creating a new account, we do need to show the data for newly created accounts.  We don't necessarily have to redirect to a route to show this info off.  I'm happy to help figure this out.  We might need to split `saveAccountActions` into two separate actions -- one for creating the account, and one for saving it to local storage.

comountainclimber:
@mhuggins Do you think that wallet creation should have "wallet name" as optional? For now what I am thinking is that we will keep the `DisplayWalletAccounts` view during import and during creation (with wallet name being optional)

If that sounds ok to you I can quickly prep some mocks for what I think a V2 `DisplayWalletAccounts` component will look like.

app/components/BackButton/BackButton.jsx
comountainclimber: 👍

app/containers/Home/WithHomeMarkUp.jsx
comountainclimber: Much better... The `with` naming convention is supposed to be reserved for denoting HOC.

app/components/Root/Routes.jsx
mhuggins: Resolved. 👍

app/containers/CreateWallet/CreateWallet.jsx
```diff
@@ -15,73 +21,108 @@ type State = {
   passphrase: string,
   passphrase2: string,
   wif: string,
+  option: string,
```
mhuggins: Maybe change this to `oneOf(['CREATE', 'IMPORT'])` to be more explicit?

app/containers/CreateWallet/OptionButton.jsx
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import React from 'react'
+import styles from '../Home/Home.scss'
+
+const OptionButton = ({ active, children, handleClick }: Object) => {
+  return (
+    <div
+      onClick={handleClick}
+      className={active ? styles.activeOptionButton : styles.optionButton}
```
mhuggins: We should create a separate `OptionButton.scss` file that these styles go into.

app/containers/CreateWallet/CreateWallet.jsx
```diff
@@ -15,73 +21,108 @@ type State = {
...
+      walletName
+    )
   }
 
-  render () {
-    const { encryptWIF } = this.props
-    const { passphrase, passphrase2, wif } = this.state
+  render = () => {
+    const { passphrase, passphrase2, wif, option, walletName } = this.state
     let disabledButton
```
mhuggins: This isn't being set anymore.  I think we could replace this local var with an `isValid` or `isDisabled` function based upon `passphrase`, `passphrase2`, etc.

comountainclimber: 👍

app/containers/CreateWallet/OptionButton.jsx
comountainclimber: 👍

app/containers/CreateWallet/CreateWallet.jsx
comountainclimber: 👍

app/containers/CreateWallet/OptionButton.jsx
comountainclimber: ☝️ this has been resolved with https://github.com/CityOfZion/neon-wallet/pull/1070/commits/bef1f8aacf6aa62dd016a1c77fcff876385caff9

app/containers/CreateWallet/CreateWallet.jsx
```diff
@@ -17,11 +17,13 @@ type Props = {
   history: Object
 }
 
+type Option = 'CREATE' | 'IMPORT'
```
mhuggins: Haha, right, we're using flow.  Too much context switching.  😅

PR #1071, opened 2018-05-07
V2/retry

mhuggins:
N/A

This adds the ability to retry failed requests (e.g.: when data fails to load from an API or RPC server).

retry-failed

I created an `onRetry` function prop that is passed to the `FailedPanel` component.  When that component renders, the user can click on a link to call that function.

I forced an error to be raised on the first instance that an action is called, but to continue processing normally on subsequent calls.

N/A

@snowypowers I initially tried to test this by turning off my wifi, getting failed requests, then turning my wifi back on and retrying, but it did not work as expected.  After the first failed set of requests, the app continued to hit endpoints like http://localhost:3000/undefined/v2/network/nodes and http://localhost:3000/undefined/v1/get_all_nodes.  Manually switching between TestNet and MainNet resolved the issue.  I'm wondering if neon-js might be caching something?  If not, I'll dig in more on my end as a follow-up.

- [ ] Unit tests written?

### DIFF
--- a/__tests__/components/__snapshots__/TransactionHistoryPanel.test.js.snap
+++ b/__tests__/components/__snapshots__/TransactionHistoryPanel.test.js.snap
@@ -1,9 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`TransactionHistoryPanel renders without crashing 1`] = `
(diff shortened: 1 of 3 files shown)


snowypowers:
those urls look like neoscan/neonDB urls. Those should be hardcoded as opposed to caching? I don't actually understand the problem here.

PR #1084, opened 2018-05-11
added Guardium hash

HayesData:
n/a

Adds the Guardium (GDM) token hash to the wallet.

Added the hash to the constants file.

ran/tested locally

nope

nope

- [ ] Unit tests written?
nope

### DIFF
--- a/app/core/constants.js
+++ b/app/core/constants.js
@@ -68,6 +68,7 @@ export const TOKENS_TEST = {
 
 // MainNet
 export const TOKENS = {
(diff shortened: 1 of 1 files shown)


HayesData:
build appears to be failing due to a dependency outside of the scope of work I did on this branch

mhuggins:
@HayesData Are you guys having a token sale for this?  Has it already ended?  If it has, we should add this to the `ENDED_ICO_TOKENS` constant as well.

Issue #1029, opened 2018-04-29
Use paginated endpoint for getting transaction history


drptbl:
Currently wallet uses non-paginated endpoint while getting transaction history.

**Url example:**
`https://api.neoscan.io/api/main_net/v1/get_address_neon/AK6SD7ZuZZkNPDi5SX9WkeELReob6tPBXQ`

**Code:**
https://github.com/CityOfZion/neon-wallet/blob/0076a20d97ae04b25a34251a341d6d19cbde7a7e/app/actions/transactionHistoryActions.js#L17

Two main drawbacks of doing this is that backend takes higher load and users are waiting longer on "Send" page after adding new recipient to the list. It should use paginated endpoint instead.

drptbl:
Fixed with #1117. Re-tested.

PR #1117, opened 2018-05-17
Updated transaction API fetches to use paginated neoscan endpoint

mhuggins:
N/A

This changes the transaction history fetch to get the latest page of transactions from neoscan rather than the full history.  It also updates the transaction history sidebar to gracefully fail if the fetch does not succeed, preventing the white screen of death.

I replaced the neon-js call with a raw axios call.

I smoke tested several addresses and compared the results to neoscan.

N/A

@snowypowers In case you want to mimic this behavior in neon-js, this is what I did.

- [ ] Unit tests written?

### DIFF
--- a/__tests__/components/__snapshots__/TransactionHistory.test.js.snap
+++ b/__tests__/components/__snapshots__/TransactionHistory.test.js.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`TransactionHistory renders without crashing 1`] = `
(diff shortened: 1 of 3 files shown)


mhuggins:
Had the vin/vouts backwards, fixed this with one additional commit: f1f688e0c3774ddf3b0a318510b0e061644269a7

drptbl:
There was an issue for this - #1029. I've linked it, thank you 👍.

drptbl:
Commit f1f688e also fixed #1134.

PR #1118, opened 2018-05-17
Add more ledger error msgs

snowypowers:
- Message for transaction denied
- Message for parsing error. This is mostly likely due to the NEO app being outdated.

Side effect of getting a ton of lint/prettier changes


mhuggins:
Circle failures are e2e tests that won't work on TestNet right now because neoscan is resyncing.  Merging.

mhuggins:
Actually I'll merged as soon as #1119 is merged.

PR #1133, opened 2018-05-20
Update README.md

lindsayemarc:
standalone apps are available now. Updating readme to reflect that


lindsayemarc:
looks like integration tests failed because of timeouts. That is not an issue with this PR, as this simply changes the readme.

Issue #1122, opened 2018-05-18
Wallet ends in blank screen with "Failed to load." message if coinmarketcap request failed


drptbl:
**Wallet version or commit SHA:**
0.2.5

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.48

**Prerequisites:**
Any app for intercepting of HTTP requests

**Requested url:**
`api.coinmarketcap.com/v1/ticker/?limit=0&convert=usd`

**Reproduction steps:**
1. Start intercepting
2. Start Neon Wallet
3. After url mentioned above is intercepted, drop it (or change response to `null`)

Wallet ends in a blank screen with "Failed to load." message. We shouldn't rely on coinmarketcap API to run the wallet properly.

**Screenshots/Video:**
![cmc failed](https://user-images.githubusercontent.com/8177587/40238268-0834ee1e-5ab3-11e8-89f6-0b3f6912e2ed.PNG)

drptbl:
@evgenyboxer Just re-tested `dev` branch with latest fixes and I'm still not able to load the wallet if `api.coinmarketcap.com/v1/ticker/?limit=0&convert=usd` was dropped, changed to `null` or to empty array.

Same reproduction steps as above. Start intercepting, start the wallet, change the response and wallet doesn't load with "Failed to load." message.

**Screenshots/Video:**
![neoncmc](https://user-images.githubusercontent.com/8177587/40589828-0b4624f8-61f5-11e8-8902-8c92d0fd0bd6.gif)

drptbl:
Re-tested, fixed. Thanks.

PR #1135, opened 2018-05-22
fix(dashboard): Allow dashboard to load despite having bad cmc response

evgenyboxer:
https://github.com/CityOfZion/neon-wallet/issues/1122

When CMC is down the dashboard will turn blank.

Using lodash `get` when getting the prices, and also made a new `Prices` type, and marked all prices as optionals.
In addition to that, I also replaced the `neoValue/gasValue` calculations to use BigNumber.

Tested it with cmc with null/undefined/bad values

No.

No.

- [ ] Unit tests written?
No.

### DIFF
--- a/app/containers/WalletInfo/index.js
+++ b/app/containers/WalletInfo/index.js
@@ -2,7 +2,7 @@
 import { connect, type MapStateToProps } from 'react-redux'
 import { bindActionCreators } from 'redux'
 import { compose } from 'recompose'
-import { values, omit, filter } from 'lodash'
+import { values, omit, filter, get } from 'lodash'
 import { withData, withActions } from 'spunky'
 
 import accountActions from '../../actions/accountActions'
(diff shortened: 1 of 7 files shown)


app/containers/WalletInfo/index.js
```diff
@@ -41,9 +41,12 @@ const mapBalanceDataToProps = (balances) => ({
...
 
-const mapPricesDataToProps = ({ NEO, GAS }) => ({
-  neoPrice: NEO,
-  gasPrice: GAS
+const mapPricesDataToProps = (prices: ?Prices): {
+  neoPrice: ?number,
+  gasPrice: ?number
+} => ({
+  neoPrice: get(prices, 'NEO'),
+  gasPrice: get(prices, 'GAS')
```
comountainclimber: Nice 👍

PR #1139, opened 2018-05-26
Add some warnings for deprecated tokens

miguelzf:

### DIFF
--- a/app/containers/WalletInfo/TokenBalances.jsx
+++ b/app/containers/WalletInfo/TokenBalances.jsx
@@ -40,7 +40,7 @@ const TokenBalances = ({ tokenBalances, showModal }: Props) => (
               <Tooltip
                 title={formattedBalance}
                 disabled={isZero(balance)}>
-                {formattedBalanceDisplay} {deprecated && <span style={{ color: 'red' }}>(deprecated)</span>}
+                {formattedBalanceDisplay} {deprecated && <span className={styles.deprecated}>(deprecated)</span>}
               </Tooltip>
             </td>
           </tr>
--- a/app/containers/WalletInfo/TokenBalances.jsx
+++ b/app/containers/WalletInfo/TokenBalances.jsx
@@ -4,7 +4,7 @@ import React from 'react'
 import Table from '../../components/Table'
 import Tooltip from '../../components/Tooltip'
 
-import { MODAL_TYPES } from '../../core/constants'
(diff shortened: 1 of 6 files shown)


app/core/constants.js
```diff
@@ -90,6 +90,16 @@ export const ENDED_ICO_TOKENS = [
...
 ]
 
+export const DEPRECATED_TOKENS = [
+  '4b4f63919b9ecfd2483f0c72ff46ed31b5bbb7a4', // Phantasma SOUL
+  'a0b328c01eac8b12b0f8a4fe93645d18fb3f1f0a', // NKN
+  '7ac4a2bb052a047506f2f2d3d1528b89cc38e8d4', // quarteria
+  '23501e5fef0f67ec476406c556e91992323a0357', // Orbis
+  '442e7964f6486005235e87e082f56cd52aa663b8', // Ontology
+  '2e25d2127e0240c6deaf35394702feb236d4d7fc', // Narrative
+  '6d36b38af912ca107f55a5daedc650054f7e4f75' // Aphelion
```
mhuggins: How do you suggest I go about verifying these are the correct deprecated script hashes?

app/containers/WalletInfo/TokenBalances.jsx
```diff
@@ -39,7 +40,7 @@ const TokenBalances = ({ tokenBalances, showModal }: Props) => (
               <Tooltip
                 title={formattedBalance}
                 disabled={isZero(balance)}>
-                {formattedBalanceDisplay}
+                {formattedBalanceDisplay} {deprecated && <span style={{ color: 'red' }}>(deprecated)</span>}
```
mhuggins: Would prefer that we use a className (via `TokenBalances.scss`) instead of inline styles here.

app/core/constants.js
miguelzf: I looked at duplicates mainnet contracts where the oldest 1 isnt on the list here

But yeah, probably better to contact each owner to ask if they want to list it as deprecated

I can edit the PR to leave only our contract now (SOUL). Also NRV and Orbis i know have been deprecated

evgenyboxer:
@miguelzf any update on this? (regarding Matt's comments)

miguelzf:
Sorry for the delay guys, pushed a commit to move the stylings to its proper place

drptbl:
Tested, looks good 👍 .

PR #1156, opened 2018-05-28
Feature/update placeholder text

comountainclimber:
There is no issue currently open but @drptbl noticed that our placeholders dont match the mocks and I agree they are verbose...This updates placeholders to consistently just use`Password`.

- [ ] Unit tests written?

### DIFF
--- a/app/containers/LoginNep2/LoginNep2.jsx
+++ b/app/containers/LoginNep2/LoginNep2.jsx
@@ -39,7 +39,7 @@ export default class LoginNep2 extends Component<Props, State> {
             />
           </div>
           <PasswordInput
-            placeholder="Enter your passphrase here"
+            placeholder="Password"
             value={passphrase}
             disabled={loading}
             onChange={e => this.setState({ passphrase: e.target.value })}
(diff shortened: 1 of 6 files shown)


app/containers/LoginNep2/LoginNep2.jsx
```diff
@@ -39,7 +39,7 @@ export default class LoginNep2 extends Component<Props, State> {
             />
           </div>
           <PasswordInput
-            placeholder="Enter your passphrase here"
+            placeholder="Password"
```
mhuggins: Up on line 34 is a verbose placeholder "Enter your encrypted key here".  Should we update that as well, perhaps to say "Encrypted Key"?

comountainclimber: 👍 sounds good @mhuggins

mhuggins:
@comountainclimber This broke a test, you'll need to update `LoginNep2.test.js` to change "Enter your encrypted key here" to "Encrypted Key".

app/containers/LoginNep2/LoginNep2.jsx
```diff
@@ -31,15 +31,15 @@ export default class LoginNep2 extends Component<Props, State> {
         <form onSubmit={this.handleSubmit}>
```
affendy93: 😁

comountainclimber: ???

PR #1158, opened 2018-05-28
Updated to fetch block height from RPC

### LINKED ISSUE #1127: Get block height from RPC instead of neoscan
As discussed on Discord, block height shouldn't be taken from explorer, but RPC nodes instead.

mhuggins:
Fixes #1127.

This replaces the API call for the current block height with an RPC call.

I update the `blockHeightActions` file to use RPC for getting the current block height.

I logged into an account and ensured the block height displayed in the header.

N/A

N/A

- [x] Unit tests written?

### DIFF
--- a/__tests__/actions/blockHeightActions.test.js
+++ b/__tests__/actions/blockHeightActions.test.js
@@ -1,4 +1,5 @@
 import { api } from 'neon-js'
+import nock from 'nock'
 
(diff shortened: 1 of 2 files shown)


drptbl:
Raised some questions about this one, but it's tested for now.

drptbl:
This PR introduced: #1229

PR #1159, opened 2018-05-28
TokenBalancesPanel.jsx is non spammable

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1021

- [ ] Unit tests written?


mhuggins:
Thanks!  Will merge after tests run.

PR #1160, opened 2018-05-28
Fixed e2e test for transaction history

mhuggins:
N/A

The end-to-end test for checking transaction history has not been working as expected.  This was unknown because neoscan TestNet had been down for awhile.  Now that it's working, we can see that this test is failing.

I updated the test to not care about transaction ID ordering.

I ran the test suite.

N/A

N/A

- [x] Unit tests written?

### DIFF
--- a/__e2e__/view-balance.e2e.js
+++ b/__e2e__/view-balance.e2e.js
@@ -52,10 +52,17 @@ test.serial('should show correct balance', async t => {
 })
 
 test.serial('should show correct transaction list', async t => {
-  await app.client.waitUntilTextExists('#transactionList li:first-child .txid', '57da6b7a1074c8508796549c19fdb2a8', 60000)
+  const txids = [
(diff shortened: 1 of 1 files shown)


__e2e__/view-balance.e2e.js
```diff
@@ -52,10 +52,17 @@ test.serial('should show correct balance', async t => {
...
 
 test.serial('should show correct transaction list', async t => {
-  await app.client.waitUntilTextExists('#transactionList li:first-child .txid', '57da6b7a1074c8508796549c19fdb2a8', 60000)
+  const txids = [
+    '57da6b7a1074c8508796549c19fdb2a8',
+    '4bb9b6e0a6ef46c42dd6a1f11326fb0c'
+  ]
+
+  await app.client.waitUntilTextExists('#transactionList li .txid', txids[0], 60000)
   const transactions = await app.client.getText('#transactionList li .txid')
```
affendy93: 😁

Issue #1129, opened 2018-05-18
Wallet ends in blank page if neoscan 'get_unclaimed' request returns null


drptbl:
**Wallet version or commit SHA:**
0.2.5

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.48

**Prerequisites:**
- Any app for intercepting of HTTP requests

**Requested url:**
`api.neoscan.io/api/main_net/v1/get_unclaimed`

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Select your wallet, enter your passphrase and click on "Login" button
4. Start intercepting
5. Click on "X.XXX~ GAS"
3. After url mentioned above is intercepted, change it's response to `null`

Wallet ends in a blank screen after getting `null` response from request mentioned above with error in browser console.

**Screenshots/Video:**
![unclaimed](https://user-images.githubusercontent.com/8177587/40245194-35f01aec-5ac5-11e8-8067-c804ce95d005.PNG)

drptbl:
Tested, looks good. Thanks.

Issue #1124, opened 2018-05-18
Wallet ends in blank screen if neoscan 'get_last_transactions_by_address' request failed


drptbl:
**Wallet version or commit SHA:**
0.2.5

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.48

**Prerequisites:**
- Any app for intercepting of HTTP requests
- At least one wallet imported to the app

**Requested url:**
`api.neoscan.io/api/main_net/v1/get_last_transactions_by_address/{wallet_addr}`

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Start intercepting
4. Select your wallet, enter your passphrase and click on "Login" button
5. After url mentioned above is intercepted, drop it (or change response to `null`)

Wallet ends in a blank screen after dropping response from request mentioned above with error in browser console.

**Screenshots/Video:**
![lasttracastionsbyaddresserror](https://user-images.githubusercontent.com/8177587/40238846-ae106010-5ab4-11e8-9275-aec9a560f6ba.PNG)

drptbl:
Tested, works good. Thanks.

Issue #1123, opened 2018-05-18
Wallet ends in blank screen message if neoscan 'get_all_nodes' request failed


drptbl:
**Wallet version or commit SHA:**
0.2.5

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.48

**Prerequisites:**
- Any app for intercepting of HTTP requests
- At least one wallet imported to the app

**Requested url:**
`api.neoscan.io/api/main_net/v1/get_all_nodes`

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Start intercepting
4. Select your wallet, enter your passphrase and click on "Login" button
5. After url mentioned above is intercepted, drop it (or change response to `null`)

Wallet ends in a blank screen after dropping response from request mentioned above with error in browser console.

**Screenshots/Video:**
![get_all_nodes_error](https://user-images.githubusercontent.com/8177587/40238714-4aa7fc5e-5ab4-11e8-8c07-8907be42346c.PNG)

drptbl:
Tested, looks good. Thanks.

PR #1166, opened 2018-05-30
Fix connectivity issues

evgenyboxer:
https://github.com/CityOfZion/neon-wallet/issues/1129
https://github.com/CityOfZion/neon-wallet/issues/1124
https://github.com/CityOfZion/neon-wallet/issues/1123
https://github.com/CityOfZion/neon-wallet/issues/1125

This solves the blank screen we get when tx history / balances / claims are not working correctly (either a bad response or nodes are down)

For balances - falling back to `null` and showing a `-` (dash) in the NEO/GAS values, also the Send button will be disabled (with a Tooltip explaining why is it disabled), because it would require a lot of work to deal with a null value in Send modal.

For claims - falling back to `null` and showing "GAS Claim unavailable" with a tooltip explaining that the network is down.

For tx history - falling back to an empty array, and I replaced the failed message with a better message + styled.

I have tested bad responses for all the issues (along with 404's), and verified that it all works.

Nope.

Nope.

- [ ] Unit tests written?


evgenyboxer:
GAS claim error - 
screen shot 2018-05-30 at 8 24 32 pm

Balances error (with Tooltip) - 
screen shot 2018-05-30 at 8 25 15 pm

Balances error (without Tooltip) - 
screen shot 2018-05-30 at 8 25 10 pm

Transaction history error -
screen shot 2018-05-30 at 8 26 33 pm

evgenyboxer:
@drptbl

drptbl:
Issues found:
#1179,  #1181, #1183
Enhancement proposals:
#1178, #1180, #1182,

PR #1173, opened 2018-06-04
Feature/wallet manager v2

comountainclimber:
**What current issue(s) from Trello/Github does this address?**
https://github.com/CityOfZion/neon-wallet/issues/978

screen shot 2018-06-03 at 10 56 48 pm

screen shot 2018-06-04 at 12 55 27 am

screen shot 2018-06-04 at 12 55 42 am

**What problem does this PR solve?**
This PR creates the new UX for managing, editing, importing/exporting wallets saved to local storage.

**How did you solve this problem?**

**How did you make sure your solution works?**
Manual testing

**Are there any special changes in the code that we should be aware of?**
I added a new action to `accountsActions.js` and need help confirming what I did makes sense... 

**Is there anything else we should know?**

- [ ] Unit tests written?

PR #1174, opened 2018-06-04
Add ability to recall saved wallet address credentials

nickfujita:
Currently there is no way for users who have created their wallets with neon-wallet to recall the credentials screen that was displayed when they first created the wallet. While it could be argued that it is for their own security that they not be able to pull up this information again, and that it's their responsibility to save these details upon creation. However, the sad reality is that there are many people who save their wallet, and blow past this screen without a second thought. While users are able to export their wallet recovery JSON file with their encrypted key, many less technical people have no idea what this file is or how to make sense of it.

In order to remedy this, an option was added to the wallet management screen to allow users to recall the credentials screen in order to reference or print their address details after the creation of the wallet. The reason that this was added to the wallet management screen rather than the main wallet screen, to create a user flow that will allow them to only pull up this information if it's what they are specifically looking for. Rather than having it on the main wallet page, where they may leave that open on accident, or press on the icon to bring it up when the didn't intend to.

The addition of this feature improves the portability of addresses created with neon-wallet, and provides user with the option to recall and print their credentials at a later point in time.

In order to test this:
- Navigate to the "Manage Neon settings" screen from the home screen. 
- Click on the small lock icon to the right of the address that you want to pull up the details for
- Enter the passphrase for the address to unlock this information
- Upon entry of the correct passphrase, you will be brought to the credentials screen

- If an incorrect passphrase in entered, you will remain on the settings page with the modal present
- An error message will be shown at the bottom of the modal

As a side affect of this change, when saving the address upon wallet creation, the field for saving will disappear once saved. Unlike it's current state, where no indication that it was saved or not, which sometimes lead to people clicking on it multiple times and saving it multiple times.

**Design V2 Plans**
Will add this to design-v2 once https://github.com/CityOfZion/neon-wallet/pull/1173 is merged

### DIFF
--- a/app/actions/authActions.js
+++ b/app/actions/authActions.js
@@ -61,7 +61,7 @@ export const nep2LoginActions = createActions(ID, ({ passphrase, encryptedWIF }:
   return { wif: account.WIF, address: account.address, isHardwareLogin: false }
 })
 
-export const nep2DetailsLoginActions = (passphrase: string,  encryptedWIF: string, history: Object) => (dispatch: DispatchType) => {
+export const nep2DetailsLoginActions = (passphrase: string,  encryptedWIF: string, history: Object) => (dispatch: DispatchType): Promise<*> => {
   return new Promise((resolve, reject) => {
     if (!validatePassphraseLength(passphrase)) {
       reject('Passphrase too short')
--- a/app/actions/authActions.js
+++ b/app/actions/authActions.js
@@ -62,26 +62,32 @@ export const nep2LoginActions = createActions(ID, ({ passphrase, encryptedWIF }:
 })
 
 export const nep2DetailsLoginActions = (passphrase: string,  encryptedWIF: string, history: Object) => (dispatch: DispatchType) => {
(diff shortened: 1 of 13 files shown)


nickfujita:
@drptbl fixed

drptbl:
@nickfujita Thank you, just noticed 3 more things:
1. Could we change "Login" button to some other label? In the reality we are not logging in anywhere. Maybe "Recover" or something similar?
2. Can we fix the input field for password to end in the same place as buttons below? Look at this as reference:
align
currently
3. Can we make error text horizontally centered with buttons on the right? Look at this example:
add-padding

nickfujita:
@drptbl updated

drptbl:
@nickfujita Looks like after updating styles for password input field, it broke in two places:
1. Create new wallet:
create


2. Encrypt an existing key:
encrypt

These issues are not present in current production build.

evgenyboxer:
@drptbl could you retest and confirm?

drptbl:
@evgenyboxer Re-tested, looks good.

PR #1175, opened 2018-06-04
Add option to save account NEP2 login screen

nickfujita:
Users who have have a encrypted key that was not created in neon-wallet or imported via JSON, need to have an easy option to allow them to save their account when logging in.

Added an options to set a account name, and save the account upon logging in.

- Naivigate to "Login using an encrypted key"
- Enter passphrase and encrypted key
- Check the option for save account
- Enter a account label
- Login
- If successful, you will be taken to the normal account dashbaord

Corner cases
- If the user already saved this encrypted key, it will simply ignore the new label, and log in the user
- If the user enters a label that is conflicting with one that is already saved, it will show them an error notification

### DIFF
--- a/__tests__/components/__snapshots__/LoginNep2.test.js.snap
+++ b/__tests__/components/__snapshots__/LoginNep2.test.js.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`LoginNep2 renders without crashing 1`] = `
-<withActions(Connect(Connect(withError(Connect(withProgress(ErrorNotifier))))))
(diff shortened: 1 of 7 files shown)


drptbl:
@nickfujita Thanks.

Looks like it doesn't save new account for me. Tested on Windows 10.

save_key

nickfujita:
@drptbl fixed

drptbl:
@nickfujita Everything works great here 👍.

Would it be hard to add this message if account has been saved with success?
acc_saved

This message show up if account was saved with success after saving it on new wallet creation.

nickfujita:
Added `Account saved!` notification.

drptbl:
Re-tested, works fine, thank you.

Ashleygross:
Needs little bit more

PR #1187, opened 2018-06-05
Fixed spacing on portfolio pie chart to match design notes (fixes #1006)

### LINKED ISSUE #1006: Too much white space between pie chart and "Portfolio" section border on wallet dashboard
71bee7e

**Environment:**
Win 10 Home (64 bit) Version 1709 Build 16299.371

**Prerequisites:**

**Reproduction steps:**
1. Start Neon Wallet

I find it important as fixing this could give us more space for values in "Portfolio" section and do less truncating.

For reference check `2.0-Home-Light` designs.

**Screenshots/Video:**

*Currently*:
curr

*Expected*:
exp

alexhulbert:
Issue #1006 

The portfolio pie chart had padding of its own. When combined with the padding of its parent container, there was too much whitespace between the edge of the portfolio panel and the chart.

Cutting down on the whitespace between the edge of the portfolio panel and pie chart to match the design documents as well as other elements on the page.

See for yourself!

Before:
Before

After:
After

This is probably nothing, but I used a negative margin value (margin-top: -25px), which is kind of hokey, but unfortunately, it was the only elegant solution since the chart's padding is baked into the SVG and I didn't want to touch that. I figured it would be okay since there are other instances of people using negative margin values elsewhere in the codebase.

### DIFF
--- a/app/components/Dashboard/PortfolioPanel/PortfolioBreakdownChart.jsx
+++ b/app/components/Dashboard/PortfolioPanel/PortfolioBreakdownChart.jsx
@@ -29,8 +29,8 @@ export default class PortfolioBreakdownChart extends React.Component<Props> {
     const data = this.getData()
 
     return (
-      <ResponsiveContainer width={220} className={classNames(styles.priceHistoryChart, className)}>
-        <PieChart width={220} height={180}>
+      <ResponsiveContainer width={200} className={classNames(styles.priceHistoryChart, className)}>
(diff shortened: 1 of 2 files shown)


alexhulbert:
Oh I just realized I accidentally set the pie chart's width to 200 instead of 220. I can fix it if that's an issue. (although looking at that screenshot it actually seems a bit closer to the design documents)

PR #1189, opened 2018-06-05
Added percent change to assets panel (fixes #992)

### LINKED ISSUE #992: Percentage changes are missing in "Balances" section (for NEO and GAS) on wallet dashboard
be16b42

**Prerequisites:**

**Reproduction steps:**
1. Start Neon Wallet

Percentage changes are missing in "Balances" section (for NEO and GAS). 

For reference check `2.0-Home-Light` designs.

**Screenshots/Video:**
balances_current

alexhulbert:
Issue #992 

There were no price change chips in the asset balances panel

I used the priceHistoryActions to retrieve Yesterday's NEO and GAS prices and computed the percent change. I then added this to the right of the formatted fiat values in the assetBalances panel.

I checked the price changes data against the data in the price history graph. It matched.

- [X] Unit tests written?


app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -99,11 +107,35 @@ export default class AssetBalancesPanel extends React.Component<Props> {
     return neoPrice && NEO !== '0' ? neoPrice * toNumber(NEO) : 0
   }
 
+  getNEOFormattedPriceChange = (): string => {
+    return (this.props.neoPriceChange >= 0 ? '+' : '') + (this.props.neoPriceChange * 100).toFixed(2) + '%'
+  }
+
+  getNEOPriceChangeDirection = (): string => {
```
evgenyboxer: I think you can specify the type here

`type PriceDirection = 'increase' | 'decrease'`

`getNEOPriceChangeDirection = (): PriceDirection`

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -99,11 +107,35 @@ export default class AssetBalancesPanel extends React.Component<Props> {
     return neoPrice && NEO !== '0' ? neoPrice * toNumber(NEO) : 0
   }
 
+  getNEOFormattedPriceChange = (): string => {
+    return (this.props.neoPriceChange >= 0 ? '+' : '') + (this.props.neoPriceChange * 100).toFixed(2) + '%'
```
evgenyboxer: Its a bit hard to read, how about:
`const { neoPriceChange } = this.props` 
and then using ES6 template for the contact?

Also, better use `bignumber.js` here.

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -99,11 +107,35 @@ export default class AssetBalancesPanel extends React.Component<Props> {
...
+    }
+  }
+
   getGASValue = (): number => {
     const { GAS, gasPrice } = this.props
     return gasPrice && GAS !== '0' ? gasPrice * toNumber(GAS) : 0
   }
 
+  getGASFormattedPriceChange = (): string => {
+    return (this.props.gasPriceChange >= 0 ? '+' : '') + (this.props.gasPriceChange * 100).toFixed(2) + '%'
```
evgenyboxer: Better extract that logic to a function and use it in both `getGASFormattedPriceChange` and `getNEOFormattedPriceChange`.

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -99,11 +107,35 @@ export default class AssetBalancesPanel extends React.Component<Props> {
...
+  }
+
+  getGASPriceChangeDirection = (): string => {
+    if (this.props.gasPriceChange < 0) {
```
evgenyboxer: Better extract that logic to a function and use it in both `getGASPriceChangeDirection` and `getNEOFormattedPriceChange`

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -10,13 +10,18 @@ import { toNumber, toBigNumber } from '../../../core/math'
...
+
+type PriceDirection = 'increase' | 'decrease'
 
 type Props = {
   className: ?string,
   NEO: string,
   GAS: string,
   neoPrice: number,
   gasPrice: number,
+  neoPriceChange: number,
```
evgenyboxer: This needs to be of type `BigNumber`

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -10,13 +10,18 @@ import { toNumber, toBigNumber } from '../../../core/math'
...
+  gasPriceChange: number,
```
evgenyboxer: same here ^

app/components/Dashboard/AssetBalancesPanel/index.js
```diff
@@ -29,6 +32,18 @@ const mapPricesDataToProps = ({ NEO, GAS }) => ({
   gasPrice: GAS
 })
 
+const mapPriceChangeDataToProps = (prices, props) => {
```
evgenyboxer: is props being used? can it be dropped?
also, could you describe the shape of `prices`?

finally, I think its good to return to type too

```
const mapPriceChangeDataToProps = (prices, props): { neoPriceChange: BigNumber, gasPriceChange: BigNumber }  => {
}
```

evgenyboxer:
Awesome job, few more comments

app/components/Dashboard/AssetBalancesPanel/index.js
alexhulbert: Props isn't being used so I can remove that, but I don't think return types will do us any good. There are dozens of these mapXDataToProps functions in the code and not one of them has a return type. I think this is because it's just injecting the resulting object into the properties.

Issue #1195, opened 2018-06-06
Improve testing of Address component


austinhinderer:
**Wallet version or commit SHA:**
e.g.: 8acf86b02aafc262417565eabfe69080da323723

**Environment:**
e.g.: OSX

**Prerequisites:**
None

**Reproduction steps:**
e.g.:
1. Start Neon Wallet
2. Select a wallet and login
3. Click on "Receive" button
4. Click on "Address" button

PR #1196, opened 2018-06-06
Add address tests

austinhinderer:
https://github.com/CityOfZion/neon-wallet/issues/1195

Improves functional testing of Address component

- Re-worked component code to properly export functions for mocking

- Built wallet, confirmed that the address button is still clickable

N/A

- [x] Unit tests written?


evgenyboxer:
Could you turn that component to stateless?

Issue #1197, opened 2018-06-06
Improve testing of Transaction component


austinhinderer:
**Wallet version or commit SHA:**
8acf86b02aafc262417565eabfe69080da323723

**Environment:**
OSX

**Prerequisites:**
N/A

**Reproduction steps:**
1. Start Neon Wallet
2. Login to a wallet
3. Click on transaction hash to open transaction info in new window

PR #1198, opened 2018-06-06
Updated Transaction functions for easier mocking, added tests for Tra…

austinhinderer:
https://github.com/CityOfZion/neon-wallet/issues/1197

Adds missing tests for Transaction component

Added tests

Clicked on Transaction component to open new transaction window

N/A

N/A

- [x] Unit tests written?

### DIFF
--- a/__tests__/components/Blockchain/Transaction/Transaction.test.js
+++ b/__tests__/components/Blockchain/Transaction/Transaction.test.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import Transaction from '../../../../app/components/Blockchain/Transaction/Transaction'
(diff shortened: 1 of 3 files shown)


evgenyboxer:
Hey, @austinhinderer thanks for your contribution.

I think turning the component to a stateless component is better than the current implementation.
Also, I'm not a fan of binding in the constructor, as I prefer to use the arrow function instead.

austinhinderer:
@evgenyboxer I'm in favor of stateless functional components as well. With this change, I was only hoping to make the code testable. We can go functional as well if that's required for the change.

The reason for binding in the constructor can be found in this article: https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1#a208

I prefer the shorter syntax as well but it makes it harder to test effectively.

Issue #1202, opened 2018-06-07
Ability to prioritize transactions in send


comountainclimber:
http://cityofzion.io/neon-js/docs/en/api-transactions.html#fees

We should leverage the ability to prioritize transactions (currently built into neon-js). Add this to send window UI "priority fee"

PR #1209, opened 2018-06-09
Feature/prioritize transactions

comountainclimber:
screen shot 2018-06-09 at 2 38 05 pm

https://github.com/CityOfZion/neon-wallet/issues/1202

This allows for users to prioritize transactions using existing parts of the neon-js API. We can port this over to v2 after merging... I know this has been a priority (no pun intended) for a while now

Tested locally and verified via https://neotracker.io/tx/7de63fb646827b9cd261b0b0b253b1035882a97473fd22887c73979090852daa

If more gas is added to the fee than is available in balance an alert is popped.
screen shot 2018-06-09 at 2 39 34 pm

This upgrades neon-js dependency to 3.8.0.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/SendModal/ConfirmDisplay.jsx
+++ b/app/components/Modals/SendModal/ConfirmDisplay.jsx
@@ -28,7 +28,8 @@ export default class ConfirmDisplay extends React.Component<Props, State> {
     agree: false
   }
 
-  render () {
+  render() {
+    console.log({ state: this.state, props: this.props })
     const { onConfirm, onCancel, entries, address, message } = this.props
     const { agree } = this.state
 
@@ -39,6 +40,7 @@ export default class ConfirmDisplay extends React.Component<Props, State> {
             <thead>
               <tr>
                 <th>Asset</th>
+                <th>Priority Fee</th>
                 <th>Recipient</th>
                 <th>Actions</th>
               </tr>
@@ -47,6 +49,7 @@ export default class ConfirmDisplay extends React.Component<Props, State> {
               {entries.map((entry, i) => (
                 <tr key={`entry-${i}`}>
                   <td>{formatBalance(entry.symbol, entry.amount)} {entry.symbol}</td>
+                  <td>{entry.priorityFee} GAS</td>
                   <td><Address address={entry.address} /></td>
                   <td>
                     <Delete className={styles.entryAction} onClick={this.handleDelete(entry)} />
@@ -68,7 +71,7 @@ export default class ConfirmDisplay extends React.Component<Props, State> {
         <div className={styles.agree}>
           <input id='agree' type='checkbox' checked={agree} onChange={() => this.setState({ agree: !agree })} />
           <label htmlFor='agree'>
-            I agree to transfer the above assets & tokens from{' '}
+            I agree to transfer the above assets & tokens from{' '} with a priority fee of
             <Address address={address} />.
           </label>
         </div>
--- a/app/components/Modals/SendModal/ConfirmDisplay.jsx
+++ b/app/components/Modals/SendModal/ConfirmDisplay.jsx
@@ -27,24 +27,34 @@ type Props = {
   onCancel: Function,
   priorityFee: string,
   onUpdatePriorityFee: Function,
-  toggleCollapsedFeeInfo: Function
+  onToggleFeeInfo: Function,
+  priorityFeeCollapsed: boolean
 }
 
 type State = {
   agree: boolean,
-  priorityFeeVisible: boolean
+
 }
 
 export default class ConfirmDisplay extends React.Component<Props, State> {
   state = {
-    agree: false,
-    priorityFee: '',
-    priorityFeeVisible: false
+    agree: false
   }
 
   render () {
-    const { onConfirm, onCancel, entries, address, message, balances, priorityFee, onUpdatePriorityFee } = this.props
-    const { agree, priorityFeeVisible } = this.state
+    const {
+      onConfirm,
+      onCancel,
+      entries,
+      address,
+      message,
+      balances,
+      priorityFee,
+      onUpdatePriorityFee,
+      onToggleFeeInfo,
+      priorityFeeCollapsed
+    } = this.props
+    const { agree } = this.state
 
     return (
       <div className={styles.confirmDisplay}>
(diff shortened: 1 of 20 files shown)


app/modules/transactions.js
```diff
@@ -77,14 +77,20 @@ const makeRequest = (sendEntries: Array<SendEntryType>, config: Object) => {
...
+
   if (script === '') {
     return api.sendAsset({ ...config, intents: buildIntents(sendEntries) }, api.neoscan)
   } else {
     return api.doInvoke({
       ...config,
       intents: buildIntents(sendEntries),
       script,
-      gas: 0
+      gas: priorityFees
```
comountainclimber: @snowypowers if you could take a look specifically here (Im pretty sure this is incorrect as it is not working currently) and explain where I need to attach priority fees this PR will be 👍

snowypowers:
```

Comment (comountainclimber) [reply app/modules/transactions.js]: Thanks!

**Comment by comountainclimber** [comment]
Ahhh ok thank you very much @snowypowers I will go ahead and refactor the UI

**Comment by comountainclimber** [comment]
The UI has been updated and manual testing is complete

**Comment by comountainclimber** [comment]
@snowypowers I addressed your comment about UI/UX and refactored PR... is there anything else you think I need to change??

File: app/components/Modals/SendModal/ConfirmDisplay.jsx [hunk 2]
```diff
@@ -55,9 +61,28 @@ export default class ConfirmDisplay extends React.Component<Props, State> {
...
+          <div className={addRecipientDisplayStyles.row}>
+            <div id='sendAmount' className={addRecipientDisplayStyles.column}>
+              <label className={addRecipientDisplayStyles.label}>Priority Fee:</label>
+              <NumberInput
+                value={priorityFee}
+                placeholder='Priority Fee'
+                options={{ numeralDecimalScale: COIN_DECIMAL_LENGTH }}
+                onChange={onUpdatePriorityFee} />
+              <label className={addRecipientDisplayStyles.label}>
+                ({balances['GAS']} GAS Available)
```
evgenyboxer: Use `balances[ASSETS.GAS]`

app/components/Modals/SendModal/ConfirmDisplay.jsx
```diff
@@ -3,33 +3,39 @@ import React from 'react'
...
 import { Address } from '../../Blockchain'
-import { formatBalance } from '../../../core/formatters'
+import { formatBalance, COIN_DECIMAL_LENGTH } from '../../../core/formatters'
 
 import styles from './ConfirmDisplay.scss'
+import addRecipientDisplayStyles from './AddRecipientDisplay.scss'
 
 type Props = {
   address: string,
+  balances: Object,
```
evgenyboxer: Use `BalancesType`

app/components/Modals/SendModal/SendModal.jsx
```diff
@@ -143,9 +149,14 @@ export default class SendModal extends Component<Props, State> {
   }
 
   handleConfirmTransaction = () => {
-    const { sendTransaction } = this.props
-    const { entries } = this.state
-    sendTransaction(entries).then(this.close)
+    const { showErrorNotification, sendTransaction } = this.props
+    const { priorityFee, balances, entries } = this.state
+
+    if (priorityFee && toBigNumber(priorityFee).gt(balances.GAS)) {
```
evgenyboxer: `balances[ASSETS.GAS]`

app/modules/transactions.js
```diff
@@ -132,6 +131,7 @@ export const sendTransaction = (sendEntries: Array<SendEntryType>) => async (
       tokensBalanceMap,
       address: fromAddress,
       publicKey,
+      fees: Number(priorityFee),
```
evgenyboxer: Use `toNumber` from `core/math.js`

evgenyboxer: Also, where is `fees` used?

evgenyboxer:
UI Wise, what happens if they got no GAS at all?

app/modules/transactions.js
snowypowers: `fees` is passed as part of the config object into `sendAsset` / `doInvoke` which will handle the rest.

snowypowers:
My approval is for interactions with the backend. I will let evgeny handle the rest of the front end review

comountainclimber:
@evgenyboxer If they have no gas the alert is shown. See my description of PR

evgenyboxer:
@comountainclimber could you add the fee priority as a toggle?
"Add priority fee" - that would trigger the label+input...  (instead of a blank text input) as this isn't something that everyone would use all the time.

comountainclimber:
@evgenyboxer @drptbl Thank you guys for the excellent feedback and for testing! I will go ahead and implement your suggestions. In terms of @drptbl comments for item #3 was planning on implementing the prompt if priority fee is greater than 0.1 GAS. If someone thinks the threshold should be greater or lower let me know

comountainclimber:
The UI/UX has been updated as per the suggestions... I also polished up the modal sizes:
screen shot 2018-06-11 at 10 21 39 pm
screen shot 2018-06-11 at 10 21 44 pm

Theres still no alert if you enter in more than .1 GAS let me know if you think that is still necessary @evgenyboxer

app/components/Modals/SendModal/SendModal.jsx
```diff
@@ -85,15 +89,27 @@ export default class SendModal extends Component<Props, State> {
...
   handleAddRecipient = () => {
     this.setState({ display: DISPLAY_MODES.ADD_RECIPIENT })
   }
 
+  determineModalHeight = () => {
+    const { display, priorityFeeCollapsed } = this.state
+    if (display === DISPLAY_MODES.ADD_RECIPIENT) {
+      return '350px'
+    } else if (priorityFeeCollapsed) return '350px'
+    return '485px'
```
evgenyboxer: shouldn't this be: 

`return display === DISPLAY_MODES.ADD_RECIPIENT || priorityFeeCollapsed ? '350px' : '485px'` ? (you return the same value in 2 conditions)

also, would be best if these are constants

comountainclimber:
@evgenyboxer added the constants to core/constants and fixed the weird function... let me know if there is anything else and thanks for the review 👍

evgenyboxer:
@drptbl you can QA this final piece, and we can finally merge 👍

drptbl:
@comountainclimber @evgenyboxer 
There is a new bug which has to be fixed before this is merged. After adding a fee, changing it to any amount, then removing and sending transaction - fee is still added.

Reproduction steps:
1. Login to the wallet
2. Click on "Send" button
3. Add any amount of any token to transaction
4. Click on "Next" button
5. Click on "Add Priority Fee" button
6. Fill any amount of priority fee (or click on "Add smallest fee possible")
7. Click on "Remove Priority Fee" button
8. Click on "I agree~" checkbox
9. Click on "Send Assets" button

After transaction is sent, previous fee which was removed by "Remove Priority Fee" button is still added to the transaction.

comountainclimber:
Weird I thought I manually tested this scenario... Anyway excellent catch @drptbl 🙏 I will go ahead and address this later today.

comountainclimber:
@drptbl should be good to go now

drptbl:
@evgenyboxer @comountainclimber Looks good now 👍.

PR #1212, opened 2018-06-10
Feature/prioritize transactions v2

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1202

see => https://github.com/CityOfZion/neon-wallet/pull/1209

https://neotracker.io/tx/48dcedf98a14755af15ca34e538d5dc349a111cc53d363254a04ae9dbbd32aa8

This upgrades neon-js dependency to 3.8.0

- [ ] Unit tests written?


evgenyboxer:
@comountainclimber how does it look? we would need to let Ben design it

comountainclimber:
Seeing as none of the v2 send work is complete anyway I figured it would be nice to have this code already merged in before the redesign work is done rather than after... Currently it looks exactly like it does in v1 but with the new shared components

PR #1218, opened 2018-06-12
Cleared yarn.lock and exposed outdated dependency. Updated babel conf…

austinhinderer:
…iguration to properly transform code for jest tests

None

Clearing yarn.lock exposes that we're using an outdated interaction between Jest and Babel.

Explicitly included babel-jest to help with the transformation of code to commonjs for Jest.

Deleted `yarn.lock`, ran `yarn install` then `yarn prepush` to validate that tests weren't passing due to import not being available. Made changes and confirmed `yarn prepush` now passed.

N/A

N/A

- [ ] Unit tests written?


package.json
```diff
@@ -115,9 +115,11 @@
     "babel-cli": "6.26.0",
     "babel-core": "6.26.0",
     "babel-eslint": "8.0.2",
+    "babel-jest": "^23.0.1",
```
evgenyboxer: Could you use tight versioning? just "23.0.1"

package.json
```diff
@@ -115,9 +115,11 @@
...
     "babel-loader": "7.1.2",
     "babel-plugin-lodash": "3.2.11",
     "babel-plugin-transform-async-to-generator": "6.24.1",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
```
evgenyboxer: same here ^

and rerun yarn install

austinhinderer: Will do.

PR #1219, opened 2018-06-12
Fix explorer links

austinhinderer:
N/A

The URL format for antchain explorer has changed, which means links to transactions are broken in the wallet.

Adjusted the URL format, and reworked the code for handling URLs to be more generic. I will get explorer.js to 100% coverage in another PR.

Tested that updated links worked now.

There was a general refactor of how the explorer links are generated and how the data is stored.

N/A

- [x] Unit tests written?

### DIFF
--- a/app/core/explorer.js
+++ b/app/core/explorer.js
@@ -1,77 +1,41 @@
 // @flow
 import { openExternal } from './electron'
-import { EXPLORERS } from '../core/constants'
-import { isMainNetwork } from '../core/networks'
+import explorerConfig from './explorerConfig'
+import { isMainNetwork } from './networks'
+
+export const buildExplorerLink = (
+  networkId: string,
+  explorer: ExplorerType,
+  structure: string,
+  suffix: string
+) => {
+  if (Object.prototype.hasOwnProperty.call(explorerConfig, explorer)) {
+    const explorerInfo = explorerConfig[explorer]
+
+    return getExplorerBaseURL(networkId, explorer) + explorerInfo[structure] + suffix
+  }
+
+  return ''
+}
 
 export const getExplorerBaseURL = (networkId: string, explorer: ExplorerType) => {
-  let baseURL
-  if (explorer === EXPLORERS.NEO_TRACKER) {
-    if (isMainNetwork(networkId)) {
-      baseURL = 'https://neotracker.io'
-    } else {
-      baseURL = 'https://testnet.neotracker.io'
-    }
-  } else if (explorer === EXPLORERS.NEO_SCAN) {
-    if (isMainNetwork(networkId)) {
-      baseURL = 'https://neoscan.io'
-    } else {
-      baseURL = 'https://neoscan-testnet.io'
-    }
-  } else {
-    if (isMainNetwork(networkId)) {
-      baseURL = 'http://antcha.in'
-    } else {
-      baseURL = 'http://testnet.antcha.in'
-    }
-  }
-  return baseURL
-}
+  if (Object.prototype.hasOwnProperty.call(explorerConfig, explorer)) {
+    const selectedExplorer = explorerConfig[explorer]
 
-export const getExplorerTxLink = (networkId: string, explorer: ExplorerType, txId: string) => {
-  const baseURL = getExplorerBaseURL(networkId, explorer)
-
-  if (explorer === EXPLORERS.NEO_TRACKER) {
-    return `${baseURL}/tx/${txId}`
-  } else if (explorer === EXPLORERS.NEO_SCAN) {
-    return `${baseURL}/transaction/${txId}`
-  } else {
-    return `${baseURL}/tx/hash/${txId}`
+    return isMainNetwork(networkId)
+      ? selectedExplorer.mainNetwork
+      : selectedExplorer.testNetwork
   }
 }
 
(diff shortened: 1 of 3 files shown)


app/core/explorer.js
```diff
@@ -1,77 +1,41 @@
...
+import explorerConfig from './explorerConfig'
+import { isMainNetwork } from './networks'
+
+export const buildExplorerLink = (
+  networkId: string,
+  explorer: ExplorerType,
+  structure: string,
+  suffix: string
+) => {
+  if (Object.prototype.hasOwnProperty.call(explorerConfig, explorer)) {
```
evgenyboxer: Why is this needed? won't `if (!!explorerConfig[explorer]) { ... }` suffice?

app/core/explorer.js
```diff
@@ -1,77 +1,41 @@
...
-  const baseURL = getExplorerBaseURL(networkId, explorer)
-
-  if (explorer === EXPLORERS.NEO_TRACKER) {
-    return `${baseURL}/tx/${txId}`
-  } else if (explorer === EXPLORERS.NEO_SCAN) {
-    return `${baseURL}/transaction/${txId}`
-  } else {
-    return `${baseURL}/tx/hash/${txId}`
-  }
+  return ''
```
evgenyboxer: better throw an error... we want to know about this

app/core/explorerConfig.js
```diff
@@ -0,0 +1,25 @@
...
+    mainNetwork: 'https://neoscan.io/',
+    testNetwork: 'https://neoscan-testnet.io/',
+    trxLinkStructure: 'transaction/'
+  },
+  [EXPLORERS.ANT_CHAIN]: {
+    addressLinkStructure: 'address/info/',
+    assetLinkStructure: 'asset/hash/',
+    mainNetwork: 'http://antcha.in/',
+    testNetwork: 'http://testnet.antcha.in/',
+    trxLinkStructure: 'tx/hash/0x'
```
evgenyboxer: sweet!

app/core/explorer.js
austinhinderer: My personal feeling on !! is that's it's an anti-pattern. They do a similar thing in the end but I feel that the obj.hasOwnProperty code is more clear. The more verbose version is required because this is React.

app/core/explorer.js
austinhinderer: Is there an example of an error you can provide?

app/core/explorer.js
evgenyboxer: How about `explorerConfig.hasOwnProperty(explorer)` or even `(explorer in explorerConfig)`?

app/core/explorer.js
evgenyboxer: maybe something like this:

`throw new Error(`${explorer} is not a valid explorer type. Valid explorer types are: ${Object.keys(explorerConfig).join(', ')}`)`

app/core/explorer.js
austinhinderer: `explorerConfig.hasOwnProperty(explorer)` is not React compatible, which is why I used the longer syntax.

`in` will also return true if key gets found somewhere in the prototype chain, whereas Object.hasOwnProperty (like the name already tells us), will only return true if key is available on that object directly (its "owns" the property). [Citation](https://stackoverflow.com/questions/13632999/if-key-in-object-or-ifobject-hasownpropertykey)

evgenyboxer: What do you mean its not React compatible?

If you are doing a serious runtime validation, then your check is not sufficient. What would it help to know if the key just exists there? It can still fail if it doesn't adhere to a schema.
I recommend to use a schema validation, such as https://github.com/hapijs/joi. You can also write a flow type for `explorerConfig` to describe the shape.

austinhinderer: I had issues using obj.hasOwnProperty directly in React 15. They must have changed something about how objects are built for 16, but that means obj.hasOwnProperty works now.

evgenyboxer:
Waiting for QA @drptbl

drptbl:
@evgenyboxer 

Neotracker on testnet doesn't work for any transaction because it's not maintained anymore.

Antshares on mainnet last block update was 8 days ago which makes all my transactions lead to 404 pages (last transaction shown in my transaction history is from 5 days ago), example:
https://neoscan.io/transaction/706f8d22aaa1acc7be8fe64b92e453b0e88de1d4be4d27dac653bf128ecfb111
http://antcha.in/tx/hash/0x706f8d22aaa1acc7be8fe64b92e453b0e88de1d4be4d27dac653bf128ecfb111

Summary:
1. Neotracker works fine for mainnet but not for testnet
2. Antshares works fine for testnet but not for mainnet

austinhinderer:
None of those problems are the result of this change, however. There are no regressions caused by this code, so should that block the release?

drptbl:
@austinhinderer It's not blocking the release in any way. This pull request is named "Fix explorer links", so maybe it would be nice to sort out other issues with explorer links which are not working. That's why I've mentioned about other issues which are still there, so we can decide what to do next - fix them in this PR or fix them later.

evgenyboxer:
Lets merge it and open tickets for the issues @drptbl found.

PR #1260, opened 2018-06-26
fixes vulnerability in v2

comountainclimber:

https://discuss.atom.io/t/prevent-window-navigation-when-dropping-a-link/24365

- [ ] Unit tests written?


main.js
```diff
@@ -119,7 +124,7 @@ app.on('ready', () => {
         })
       )
     }
-    mainWindow.on('closed', function() {
+    mainWindow.on('closed', function () {
```
webda2l: mainWindow.on('closed', () => {

package.json
```diff
@@ -72,6 +72,8 @@
     "cleave.js": "1.0.7",
     "compare-versions": "^3.1.0",
     "coveralls": "3.0.0",
+    "electron": "1.8.4",
```
evgenyboxer: Could you move `electron` and `electron-devtools-installer` to devDependencies?

Issue #1229, opened 2018-06-17
"get_all_nodes" endpoint is triggered twice and "getblockcount" is triggered three times on wallet login


drptbl:
**Wallet version or commit SHA:**
 `dev` branch (c8585ba)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Prerequisites:**
- At least one wallet imported to the app

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Open Chrome developer tools and switch to "Network" tab
4. Select your wallet, enter your passphrase and click on "Login" button

After logging in for the first time in the wallet you can see that `get_all_nodes` endpoint is triggered twice in a row and `getblockcount` is invoked three times in a row. This doesn't happen on current production build.

**Screenshots/Video:**
![weirdo](https://user-images.githubusercontent.com/8177587/41507489-7abf5d74-7233-11e8-957f-f0460e9b5723.gif)

drptbl:
Whole conversation regarding this issue is continued in related PR #1264.

drptbl:
There is no way to fix double invoke for `getblockcount` without modyfing `neon-js` code, so we've agreed to leave this for v2.

PR #1264, opened 2018-06-27
Bug/remove dupe requests

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1229

By turning Routes.jsx into a component

Manual testing locally

- [ ] Unit tests written?


drptbl:
Multiple requests to `get_all_nodes` are fixed 👍.

However I've found two issues so far:
1. `getblockcount` is still invoked twice in a row (it was invoked three times in a row previously) - is there any specific reason for that?
2. "Send" button is always disabled because of "network issues", however it works fine with current `dev`. These changes were introduced with #1166.
https://github.com/CityOfZion/neon-wallet/blob/20da4ff24d4d7ff196179867fafa31d7863820f8/app/containers/Dashboard/Dashboard.jsx#L61

comountainclimber:
While I was able to fix the issues that you pointed our @drptbl unfortunately I broke the WalletInfo tests that are relying on the duplicate requests and the previous state management being present 🤦‍♂️

drptbl:
@comountainclimber Second issue with "Send" button is fixed 👍.

First issue with `getblockcount` called twice in a row is still there.

app/Routes.jsx
```diff
@@ -12,28 +12,53 @@ import Settings from './containers/Settings'
...
+    return (
+      <App {...this.state}>
+        <Switch>
+          <Route
+            path={ROUTES.DASHBOARD}
+            render={() => (
+              <Dashboard
+                handlePassBlockHeight={blockHeight => {
+                  if (blockHeight !== this.state.blockHeight) {
+                    console.log('setting block height')
```
evgenyboxer: sneaky console.log

.gitignore
```diff
@@ -13,4 +13,5 @@ coverage/*
 /yarn-error.log
 /coverage/
 /.jshintrc
-.settings
\ No newline at end of file
+.settings
+.prettierrc
```
evgenyboxer: we don't really have a .prettierrc on dev

app/Routes.jsx
evgenyboxer: I don't really like this method... `handlePassBlockHeight` feels like a hacky patch for this problem.

app/Routes.jsx
```diff
@@ -12,28 +12,53 @@ import Settings from './containers/Settings'
...
-        path={ROUTES.DISPLAY_WALLET_KEYS}
-        component={DisplayWalletAccounts}
-      />
-      <Route path={ROUTES.HOME} component={Home} />
-    </Switch>
-  </App>
-)
+export default class Routes extends Component<Props> {
+  state = {
+    blockHeight: 0
```
evgenyboxer: I really don't think its needed

app/actions/balancesActions.js
```diff
@@ -18,13 +18,18 @@ export const ID = 'BALANCES'
 async function getBalances ({ net, address, tokens }: Props) {
   const endpoint = await api.getRPCEndpointFrom({ net }, api.neoscan)
 
+  const client = new rpc.RPCClient(endpoint)
+  const blockHeight = await client.getBlockCount()
```
evgenyboxer: great idea!

app/containers/App/Header/WalletBlockHeight/index.js
```diff
@@ -1,20 +1 @@
...
-  blockHeight
-})
-
-export default compose(
-  withNetworkData(),
-  withCall(blockHeightActions),
-  withData(blockHeightActions, mapBlockHeightDataToProps),
-  withRecall(blockHeightActions, ['networkId'])
-)(WalletBlockHeight)
+export { default } from './WalletBlockHeight'
```
evgenyboxer: Instead of getting the data from `blockHeightActions` you just need to get it from `balanceActions` (since balanceActions now combines block height + balances)

comountainclimber:
Will make these changes thanks!

app/containers/App/Header/WalletBlockHeight/index.js
comountainclimber: Is there a way to get that data without actually invoking the action twice?

evgenyboxer: Try `withData(balancesActions, mapBalanceDataToProps)`

comountainclimber:
@evgenyboxer addressed your concerns but think I need your help bringing the WalletInfo component tests back from the dead

Issue #1284, opened 2018-07-03
Unable to participate in token sale, getting console error


drptbl:
**Wallet version or commit SHA:**
 `dev` branch (e109f2d)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Prerequisites:**
- At least one wallet imported to the app
- Any amount of any currency in the wallet

Here are some TestNet script hashes to test with:
https://github.com/CityOfZion/neon-js/blob/6d357964b20fa4843d3f2bb4d4e203b905e273f9/src/consts.js#L16-L22

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login using a saved wallet" button
3. Select your wallet, enter your passphrase and click on "Login" button
4. Click on "Participate in a token sale" button
5. Select any token to participate in sale
6. Fill in any amount of chosen currency (for example "1 GAS")
7. Tick all checkboxes
8. Click on "Continue" button
9. Click on "Confirm Purchase" button

After doing the last step in reproduction steps, "Confirm Purchase" button will be dimmed and nothing happens, user is not able to participate in token sale. In addition to that, you can see an error in the console.

**Screenshots/Video:**
![parttt](https://user-images.githubusercontent.com/8177587/42192102-f0614b02-7e66-11e8-99a3-7a32eff0000a.PNG)
![partt](https://user-images.githubusercontent.com/8177587/42192104-f1bc62de-7e66-11e8-9c2b-7583c5788c11.gif)

PR #1285, opened 2018-07-03
Fix sending tx/sale participation

evgenyboxer:
#1284

This PR fixes the `getBalance` selector to get the balances from the correct object path. This means that sending transactions did not work too.

`balanceActions` is now renamed to `dashboardActions` as it is only used in Dashboard component and has more than one action. I also removed `blockHeightActions` as it is not used anymore.


evgenyboxer:
@drptbl

app/actions/balancesAndBlockHeightActions.js
```diff
@@ -13,7 +13,7 @@ type Props = {
   tokens: Array<TokenItemType>
 }
 
-export const ID = 'BALANCES'
+export const ID = 'BALANCES_AND_BLOCKHEIGHT'
```
comountainclimber: 👍

comountainclimber:
Nice job fixing this @evgenyboxer looks good to me... Im going to take it on myself to increase coverage in testing specifically related to the block height stuff.

drptbl:
@evgenyboxer Tested, looks good. Thanks!

PR #1288, opened 2018-07-03
switch to statetx branch for neon-js

evgenyboxer:
#1286 

Switch to `statetx` branch so we can use the voting methods.


drptbl:
@evgenyboxer Tested, no regressions found.

Issue #910, opened 2018-03-22
Get token hashes and open ICOs dynamically from a JSON file hosted at GitHub


metachris:
Remove hardcoded token hashes from neon-wallet, and instead have neon-wallet dynamically load token hashes (and open ICOs) dynamically from a JSON file hosted at GitHub pages. This could simply be a JSON file in a `/docs` folder that is published to GitHub pages, or a separate repo (`coz/neon-wallet-tokens`?).

That way projects could add or remove their tokens and we wouldn't need to release a new version of neon-wallet for every change.

cc/ @mhuggins @snowypowers @Ejhfast @dvdschwrtz

vidov1:
Tokens are hardcoded ? Is that the reason for which I can't see a token even after entering its hash in the Neon token management menu ?

snowypowers:
If we can get the token sale format approved as an NEP, we can check the blockchain itself for the token sale start instead of manually editing the token sale list. It also comes with added benefit of checking if the current address is whitelisted.

My preference is to remove the entire hardcoded token list and have users maintain their own token list as an NEP6 wallet.

vidov1:
What about tokens, not token sales ? I deployed a token on the testnet, it has its own hash, but NEON doesn't want to display my balance with that token. What's so difficult ? Even neo-python can import a token with its hash and display the balance, even generate transfer transactions etc.

metachris:
@snowypowers I'm not sure this is a feasible approach. For one -- there are outdated NEP-5 tokens on the blockchain, for instance the first Narrative SC which people should absolutely not use. 

Having users maintain their own token list is okay for developers, but too complicated for normal people that just want to check their balances and do transfers. Even I would like to just open neon-wallet and see my token balances, without having to go to some weird websites and copy hashes around and hope things work. Neon-wallet should continue to provide this out of the box.

That's the rationale for replacing hard-coded token hashes with the ones from a simple JSON file. Then we wouldn't need to update neon-wallet as soon as anything changes too. Only benefits, easy to maintain, and very little effort, as far as i can see.

vidov1:
Well, the drawback is immediately obvious : in testnet, you deploy a token, you can't see it with NEON. Hence NEON becomes unusable for testing.
Also, I don't see why there is a dedicated menu for inputting a NEP5 token hash in NEON, if anyway NEON is going to be using only hardcoded tokens. 
That looks like a big bug to me.

metachris:
[...]

PR #1292, opened 2018-07-04
Feature/dynamic token hashes

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/910

Currently the list of nep5 tokens is hardcoded meaning we need to release a new version of the app or a user has to manually add a hash. This PR automates that process by pulling down from @dvdschwrtz 's curated list of tokens

In a separate PR against dev OR in design v2 we could leverage the image assets from https://github.com/CityOfZion/neo-tokens/tree/master/assets

- [ ] Unit tests written?

### DIFF
--- a/app/core/nep5.js
+++ b/app/core/nep5.js
@@ -36,7 +36,7 @@ export const getDefaultTokens = async (): Promise<Array<TokenItemType>> => {
       })
       .catch(error => {
         // handle error
-        console.log(error);
+        console.log(error)
         fetchedTokens = TOKENS
       })
   }
--- a/app/core/nep5.js
+++ b/app/core/nep5.js
@@ -3,9 +3,10 @@ import { map } from 'lodash'
 
 import { toBigNumber } from './math'
 import { COIN_DECIMAL_LENGTH } from './formatters'
-import { TOKENS_TEST, MAIN_NETWORK_ID, TEST_NETWORK_ID } from './constants'
+import { TOKENS, TOKENS_TEST, MAIN_NETWORK_ID, TEST_NETWORK_ID } from './constants'
 
 let fetchedTokens
+let requestToGithubSuccessful = false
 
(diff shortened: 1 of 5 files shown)


app/core/nep5.js
```diff
@@ -20,10 +22,16 @@ const getTokenEntry = ((): Function => {
   })
 })()
 
-export const getDefaultTokens = (): Array<TokenItemType> => {
+export const getDefaultTokens = async (): Promise<Array<TokenItemType>> => {
   const tokens = []
 
-  tokens.push(...map(TOKENS, (scriptHash, symbol) => getTokenEntry(symbol, scriptHash, MAIN_NETWORK_ID)))
+  // Prevent duplicate requests here
+  if (!fetchedTokens) {
```
drptbl: This will prevent tokens from being updated if list is updated after it was already fetched (meaning that list will be updated only after wallet restart after it's fetched for the first time).

comountainclimber: I did that on purpose thinking it would be ok to fetch once per login but we can change that

drptbl:
I think that we should fall back to old list if fetch fails.

app/core/nep5.js
drptbl: Just wanted to mention. Not sure if we should change this at this point because fetching of blacklisted address and coinmarketcap prices for example are working in the same way currently. However we probably should address this somehow in future (v2).

evgenyboxer:
@comountainclimber I think this does not belong in settings. Best extract it to its own action and then put that as part of the batched actions in `appActions`. After you do that, I'll put more comments which I have on the implementation

metachris:
lgtm 👍

drptbl:
After starting the wallet, `https://raw.githubusercontent.com/CityOfZion/neo-tokens/master/tokenList.json` is called twice in a row.

comountainclimber:
We will move this into v2 instead and have a nice UI in settings for handling whitelisting tokens...

Issue #1291, opened 2018-07-03
Call SetupVote to construct and send StateTransaction


evgenyboxer:

evgenyboxer:
Preparing an action for it

comountainclimber:
Seeing as priorities have shifted and consensus node voting is not yet a stable feature on the network I am closing all issues related to voting for now however have labeled them for future reference in case we want to reopen.

PR #1296, opened 2018-07-05
Adds setupVote

evgenyboxer:
#1291 

Adds a simple redux action wrapper for neon-js `setupVote`. It displays notification messages for success/failure.

### DIFF
--- a/app/modules/vote.js
+++ b/app/modules/vote.js
@@ -1,5 +1,5 @@
 // @flow
-import { api } from 'neon-js'
+import { api, wallet } from 'neon-js'
 
 import {
   showErrorNotification,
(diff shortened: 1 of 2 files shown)


comountainclimber:
Looks good to man 👍

app/modules/vote.js
```diff
@@ -0,0 +1,47 @@
...
+    showInfoNotification({ message: 'Casting vote...', autoDismiss: 0 })
+  )
+
+  try {
+    const response = await api.setupVote({
+      net,
+      account: address,
+      candidateKeys
+    })
+    if (response) {
```
dvdschwrtz: what if there was no error to catch but the response is a failure

comountainclimber: We should probably handle that edge case here

Issue #1232, opened 2018-06-17
"Retry" button in "Token Balances" section calls wrong paths (localhost instead API host)


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (26820ed)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Prerequisites:**
At least one wallet imported to the app

**Reproduction steps:**
1. Disconnect from internet connection
2. Start Neon Wallet
3. Select any saved wallet
4. Enter your passphrase
5. Click on "Login" button
6. Click on "Retry" button in "Token Balances" section

"Retry" button in "Token Balances" section calls wrong paths (localhost instead API host).

**Screenshots/Video:**
![retry_broken](https://user-images.githubusercontent.com/8177587/41512994-c0449880-7294-11e8-8a3c-7f225cc0d854.PNG)

comountainclimber:
Issue fixed in design-v2

PR #1302, opened 2018-07-07
Fixes token balances section calls wrong path [#1232]

zsiglin:
#1232 

Order matters in recompose's `compose` function. In this case, calling `withProgressPanel` after `withNetworkData` was nuking the props needed in `balancesActions`. I did not go down the rabbit hole of figuring out why this worked when an internet connection was present.

I moved `withNetworkData` before `withProgressPanel`. For consistency's sake I moved a few others to more closely align with the export defined in `AssetBalancesPanel/index.js`.

Manually tested. See proper host called in screenshot below.

screen shot 2018-07-07 at 1 57 41 pm

No.

All tests pass.


comountainclimber:
Nice work @zsiglin

Issue #1313, opened 2018-07-09
NEP-5 token balances are being hidden if neoscan `get_balance` request fails


drptbl:
**Wallet version or commit SHA:**
`dev` branch (ae58feb)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Reproduction steps:**
Drop `get_balance` request response.

Console error is thrown and NEP-5 token balances are being hidden if neoscan `get_balance` request fails.

**Screenshot/Video:**
![drop_getbalance](https://user-images.githubusercontent.com/8177587/42468744-0a2f28fc-83b6-11e8-90a1-b50acd95a895.PNG)

comountainclimber:
fixed in https://github.com/CityOfZion/neon-wallet/releases/tag/2.0.0

PR #1314, opened 2018-07-09
Feature/fixes console error

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1313 

- [ ] Unit tests written?

### DIFF
--- a/app/containers/WalletInfo/TokenBalances.jsx
+++ b/app/containers/WalletInfo/TokenBalances.jsx
@@ -14,8 +14,8 @@ import styles from './TokenBalances.scss'
 
 type Props = {
   tokenBalances: Array<TokenBalanceType>,
-  showModal: Function,
-}
+  showModal: Function
+};
 
 const TokenBalances = ({ tokenBalances, showModal }: Props) => (
   <Table className={styles.table}>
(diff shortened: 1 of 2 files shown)


drptbl:
After logging in to the wallet without any NEP-5 token balances (for example new account), you're able to see unnecessary zero in the balances section:
zero_bug

app/containers/WalletInfo/TokenBalances.jsx
```diff
@@ -26,21 +26,25 @@ const TokenBalances = ({ tokenBalances, showModal }: Props) => (
       </tr>
     </thead>
     <tbody>
-      {tokenBalances.map((token: TokenBalanceType, index: number) => {
+      {!!tokenBalances.length && tokenBalances.map((token: TokenBalanceType, index: number) => {
```
evgenyboxer: if `tokenBalances` is passed as a null or undefined the length check will cause it to die.
you can use lodash `compact` (it will always return an array)
```
import { compact } from 'lodash'
{compact(tokenBalances).map((token: TokenBalanceType, index: number) => {
```

comountainclimber: Thanks for this @evgenyboxer!

comountainclimber:
Closing this PR as this component doesn't exist anymore in `design-v2`

Issue #1249, opened 2018-06-19
[v2.5] Ability to toggle between light and dark themes


comountainclimber:
We need a state management / sass solution for toggling between light and dark themes. This will most likely be accessible via Settings and would be persisted to local storage.

zsiglin:
@comountainclimber I'll take a stab at this.

drptbl:
Tested, looks goof for now. Thanks.

Issue #1239, opened 2018-06-19
[v2.5] Implementing themed color palette


comountainclimber:
Placeholder for implementation of the themed color palettes (light and dark). Reference mockups here: https://github.com/CityOfZion/neon-wallet/tree/design-v2/design

comountainclimber:
The latest assets can be found here https://github.com/CityOfZion/neon-wallet/tree/design-v2/design

comountainclimber:
Unless otherwise noted we will no longer be initially releasing with two themes... This will be one of the biggest features in our first update

drptbl:
Implemented.

PR #1320, opened 2018-07-11
Add theme handler

zsiglin:
Hi all,

I've completed a few bug fixes already in this repo, but this is my first PR for a feature. I tried to closely follow the design patterns already in place, but speak up if not.

This PR specifically addresses the implementation of a system for toggling between themes. It does so by swapping Sass variables for native CSS variables wherever dynamic colors are required. The theme is passed in once at the root level and the variable definitions cascade down through the other components. There are a number of ways this could be done, but I think this route is pretty clean as it didn't require major changes to the way styles are implemented in the app. New themes are also VERY easy to add as they are just a simple object.

Swapping the remaining variables and defining the actual color palettes will be addressed in another PR as this is a separate issue (#1239).

#1249

Swapping color schemes in the application.

Manually tested.

**Other Notes**
You'll see some formatting/spacing changes in a few of the files. I presume this is because when they were initially committed the individual did not have Prettier enabled. Apologies if this makes the actual diff much harder to discern.

**Demo** (Gif is poor quality)
demo

### DIFF
--- a/app/containers/App/App.jsx
+++ b/app/containers/App/App.jsx
@@ -1,6 +1,5 @@
 // @flow
 import React, { Component } from 'react'
-import { compose } from 'recompose'
 
 import Sidebar from './Sidebar'
 import Footer from './Footer'
@@ -49,4 +48,4 @@ class App extends Component<Props> {
   }
 }
 
-export default compose(withThemeData())(App)
+export default withThemeData()(App)
--- a/app/containers/App/App.jsx
+++ b/app/containers/App/App.jsx
@@ -1,13 +1,16 @@
 // @flow
 import React, { Component } from 'react'
+import { compose } from 'recompose'
 
 import Sidebar from './Sidebar'
 import Footer from './Footer'
 import ModalRenderer from '../ModalRenderer'
 import Notifications from '../Notifications'
+import withThemeData from '../../hocs/withThemeData'
 import { upgradeUserWalletNEP6 } from '../../modules/generateWallet'
 
 import styles from './App.scss'
+import themes from '../../themes';
 
 type Props = {
   children: React$Node,
@@ -30,10 +33,10 @@ class App extends Component<Props> {
   }
 
   render() {
-    const { children, address } = this.props
+    const { children, address, theme } = this.props
 
     return (
-      <div className={styles.container}>
+      <div style={themes[theme]} className={styles.container}>
         {address && <Sidebar className={styles.sidebar} />}
         <div className={styles.wrapper}>
           <div className={styles.content}>{children}</div>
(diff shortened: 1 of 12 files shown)


app/containers/Settings/Settings.jsx
```diff
@@ -18,6 +19,8 @@ type Props = {
   explorer: string,
   setCurrency: string => any,
   currency: string,
+  setTheme: string => any,
+  theme: string,
```
comountainclimber: Would be nice to have theme type

app/core/constants.js
```diff
@@ -120,6 +120,8 @@ export const DEFAULT_WALLET = {
 
 export const DEFAULT_CURRENCY_CODE = 'usd'
 
+export const DEFAULT_THEME = 'Light'
```
comountainclimber: Can we change this to an object that contains both valid themes instead

app/containers/App/App.jsx
```diff
@@ -30,10 +32,10 @@ class App extends Component<Props> {
   }
 
   render() {
-    const { children, address } = this.props
+    const { children, address, theme } = this.props
 
     return (
-      <div className={styles.container}>
+      <div style={themes[theme]} className={styles.container}>
```
comountainclimber: I like this approach 👍

comountainclimber:
Overall this looks great... I would really like @evgenyboxer opinion on this approach and if he has any additional feedback.

@zsiglin would really like it if you took a stab at filling in the rest of the palette in `app/themes/Light.js` and `app/themes/Dark.js`. If its easier feel free to break that out into a separate PR

screen_shot_2018-07-10_at_5 14 13_pm
screen_shot_2018-07-10_at_5 14 23_pm

app/styles/main.global.scss
```diff
@@ -21,7 +21,7 @@ body {
   font-size: 16px;
   padding: 0px;
   margin: 0px;
-  background-color: $main-background;
+  background-color: var(--main-background);
```
comountainclimber: 🚀

app/core/constants.js
zsiglin: There can only be one default, so I'm not sure that makes sense. This follows suit with how `DEFAULT_CURRENCY_CODE` is declared. I *can* see the value in making it a constant to avoid littering the code with magic strings, but in this case I don't think it's gaining us much.

app/containers/Settings/Settings.jsx
zsiglin: A theme type would simply be a restricted set of strings. In this case it's currently `Light` or `Dark`. I think we should avoid defining that list over and over again in multiple places, but if you think it adds value here I'll add it.

zsiglin:
@comountainclimber Thanks for the feedback! 👍 I'm happy to work on the actual palette as well, but I'd prefer to do that in a separate PR. I was hoping to keep this one to just the implementation of the mechanism for switching themes.

PR #1325, opened 2018-07-13
fixes hot reloading in dev mode

comountainclimber:
Although no associated issue Im sure this drives anyone contributing to neon-wallet a little crazy as HMR was not configured properly. This PR resolves the issue please see gif below.

WITH THESE CHANGES:
hot-reload

WITHOUT:
without-hot-reload

This will speed up engineering time substantially! Currently every time a change is made developer has to reauthenticate - very painful.

By reading through the documentaion

- [ ] Unit tests written?

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,4 +1,5 @@
 import 'raf/polyfill'
+import { hot } from 'react-hot-loader'
 import React from 'react'
 import { render } from 'react-dom'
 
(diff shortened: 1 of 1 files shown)


zsiglin:
This will make things way more pleasant to work on. Nice work dude. I can't approve this formally, but 👍🏻.

PR #1356, opened 2018-08-09
Add Networkswitch to Dashboard and Settings

FredrikOseberg:
Adding NetworkSwitch back to Dashboard and Settings. Added it back for ease of development to Dashboard. We could consider having it in the top left corner of the dashboard permanently? Though if it is going to be permanent we should change the design to fit more in line with what is currently in the Dashboard,


comountainclimber:
Lets stick with the design mock on this one @FredrikOseberg 

image

Please do not feel like you need to implement any of the design above but I like the network dropdown being only on the settings page. Let me know your thoughts

FredrikOseberg:
@comountainclimber Alright. I added the NetworkSwitch to the Dashboard as a temporary measure because it would be tedious when developing to follow this flow: 

Login -> Go To Settings -> Switch to feature being developed (IE: TokenSale) 

Compared to:

Login -> Switch on Dashboard -> Go to feature being developed.

What do you think of keeping it there for development?

Regarding whether or not to keep the NetworkSwitch only in settings permanently here are my thoughts: Personally, I the number one piece of information I'm looking for when I'm using wallets is what network I'm operating on. I would be terrified of doing any action without knowing exactly the context of where I'm operating. I'm concerned only having information about which network you are operating on obfuscated in settings could hamper usability? Perhaps we could have the switch in settings only, but display the information elsewhere in the application?

FredrikOseberg:
Yes, that's perfect.

app/containers/App/Sidebar/NetworkSwitch/NetworkSwitch.jsx
```diff
@@ -16,9 +18,9 @@ export default class NetworkSwitch extends Component<Props> {
   }
 
   render() {
-    const { networkId, networks } = this.props
+    const { networkId, networks, className } = this.props
     return (
-      <div id="network" className={styles.networkSwitch}>
+      <div id="network" className={classNames(styles.networkSwitch, className)}>
         <span className={styles.label}>Running on</span>
         <select
```
comountainclimber: Please change this before we merge thanks

comountainclimber:
@FredrikOseberg did you see my comment? Can we address this asap so I can merge this work in?

FredrikOseberg:
Sorry, must have missed it. I'll take care of it.

Den fre. 17. aug. 2018, 05:14 skrev Maxwell Lasky <notifications@github.com

FredrikOseberg:
@comountainclimber It's done.

comountainclimber:
Nice job @FredrikOseberg ! Thank you for thia

PR #1362, opened 2018-08-12
chore(explorer): fixed broken link for addresses

DalderupMaurice:
None/ Discussed this with @comountainclimber in discord chat

Broken URL when clicking on addresses (on receive tab your own address for example)

`addressLinkStructure` was written incorrectly

Tested it manually

Nope

nope

- [ ] Unit tests written? --> not needed, tests already exist


comountainclimber:
Looks great @DalderupMaurice thank you very much for reporting and fixing this issue!

PR #1372, opened 2018-08-15
added GUARD token

HayesData:
GUARD token is the replace for the GDM (guardium) token implemented in the token swap on Aug 15th 2018

n/a

adds the GUARD token hash

added GUARD token hash to constants

ran tests

n/a

GUARD is the replacement token for GDM

- [ ] Unit tests written? 
no


comountainclimber:
@HayesData This is no longer the protocol for adding tokens to the application 🎉 please make a PR against this repo instead https://github.com/CityOfZion/neo-tokens

PR #1373, opened 2018-08-16
Fix missing checks for null account key in NEP6 wallet file

nickfujita:
There is currently a gap in the handling of addresses without keys in the NEP6 wallet JSON. The proposal mentions that, `This field can be null (for watch-only address or non-standard address).`, however there were not any checks for this use case on file import or read.

Added checks on wallet file import to filter out accounts with no key, since this is not a part of the user flow for neon-wallet.

Also added checks on usages in the login from local storage and setting page. While the import filter should remove these records in the first place, this should not be needed. However the checks were still added just in case, as the result is an app crash if they do occur.

In order to test this, you can import a NEP6 wallet file with one of the key values deleted. In the current app, it will crash when navigating to settings page. In this branch, it does not.

### DIFF
--- a/app/containers/LoginLocalStorage/LoginLocalStorage.jsx
+++ b/app/containers/LoginLocalStorage/LoginLocalStorage.jsx
@@ -40,9 +40,9 @@ export default class LoginLocalStorage extends Component<Props, State> {
             <option value=''>Select a wallet</option>
             {accounts && accounts.reduce((accum, account, index) => {
               if (account.key) {
-                accum.push(<option value={account.key} key={`wallet${account.label}`}>{account.label}</option>);
+                accum.push(<option value={account.key} key={`wallet${account.label}`}>{account.label}</option>)
               }
-              return accum;
+              return accum
             }, [])}
           </select>
           <div className={loginStyles.loginForm}>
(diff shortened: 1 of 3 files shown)


nickfujita:
If possible, I'd like to get it into the last build pre-V2. I think we all want V2 to be out sooner rather than later, but since we don't have an exact date, I'm afraid things will just keep getting pushed off.

The amount of code changed for this is not very much, but it prevents the app from completely crashing in the case that it occurs. Which I feel is significant enough.

I can also make sure to migrate this into V2, or at lease review the open PRs we have open for the work being done on the settings pages to ensure this does not occur in the dev V2 build.

comountainclimber:
The exact date for the release is slated for the end of this month @nickfujita but fair enough... I will go through the trouble of merging, retagging and recompiling all of the assets we have been regression testing although would much rather just focus on moving forward

nickfujita:
Also, I think it will be helpful to have one last version of the V1 app that supports accounts in the NEP6 wallet that have no key, in the case that the functionality is added in V2 to support watch addresses.

comountainclimber:
Either way @nickfujita your contributions are greatly appreciated. Thanks as always for your help 👍

nickfujita:
Thanks for being accommodating of this last min fix :)

PR #1384, opened 2018-08-19
WIP: Feat/token sale

FredrikOseberg:

- [ ] Unit tests written?


munajah23:
How to create account wallet neon address

PR #1405, opened 2018-08-30
Feature/dashboard cleanup

comountainclimber:

After reviewing the mocks I noticed that there were a few elements missing from the current dashboard. This PR adds them and also makes minor adjustments spacing and layout

- [ ] Unit tests written?


app/components/Dashboard/PortfolioPanel/PortfolioBreakdownChart.jsx
```diff
@@ -63,7 +63,6 @@ export default class PortfolioBreakdownChart extends React.Component<Props> {
               id="totalWalletValue"
             />
           </Pie>
-          <Tooltip formatter={this.formatValue} />
```
comountainclimber: Struggled with trying to style and format this value but ultimately was unsuccessful will revisit at a later date if its deemed important

PR #1421, opened 2018-09-16
WIP: Feature/token sale

FredrikOseberg:

- [ ] Unit tests written?

### DIFF
--- a/app/components/Send/SendPanel/PriorityFee/PriorityFee.jsx
+++ b/app/components/Send/SendPanel/PriorityFee/PriorityFee.jsx
@@ -37,7 +37,7 @@ export default class PriorityFee extends React.Component<Props> {
     return (
       <div>
         <div className={styles.priorityExplanationText}>
-          Prioritize your transfer with a fee?
+          Prioritize your transaction with a fee?
         </div>
         <div className={styles.priorityFeeButtonContainer}>
           {FEE_OPTIONS.map(option => (
--- a/app/components/Send/SendPanel/index.jsx
+++ b/app/components/Send/SendPanel/index.jsx
@@ -15,10 +15,8 @@ import ZeroAssets from './ZeroAssets'
 import { pluralize } from '../../../util/pluralize'
 
 import SendIcon from '../../../assets/icons/send.svg'
-import LightningIcon from '../../../assets/icons/lightning.svg'
 
 import styles from './SendPanel.scss'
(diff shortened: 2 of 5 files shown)


app/components/TokenSale/TokenSaleHeader/index.jsx
```diff
@@ -0,0 +1,16 @@
...
+  <section className={styles.tokenSaleHeader}>
+    <h1>Token Sale</h1>
+    <button type="button">
+      <RefreshIcon />
+      Refresh
+    </button>
+  </section>
+)
+
+export default TokenSaleHeader
```
comountainclimber: Theres actually a common component (`<HeaderBar/>`) for this part of the UI (it gets rendered throughout the app) please use it and remove the files associated with TokenSaleHeader

app/components/Inputs/CheckBox/index.jsx
```diff
@@ -0,0 +1,47 @@
+// flow
```
comountainclimber: For consistency with the rest of the application (I ignored this in your contribution with the Send work but would like to address it in the future) can we call this file CheckBox.jsx with another file called index.js that exports the component as well as everywhere else in this PR where this pattern exists

comountainclimber:
Something funky definitely still going on with this diff... Theres no reason why all of these commits should also be going into design-v2 with your work here but we will get it sorted out. I will try and investigate whats going with your branch later today

Unrelated to the git stuff PR looks like a solid start nice work! One thing: Can we incorporate the ability to add network fees (not in the design mocks) to the "sale" flow - You can reference the work I did in the Send UX but its really as simple as adding a key called fees to the config here: https://github.com/CityOfZion/neon-wallet/blob/dev/app/modules/sale.js#L91

app/components/TokenSale/TokenSaleHeader/index.jsx
FredrikOseberg: Perfect. Will implement.

app/components/Inputs/CheckBox/index.jsx
FredrikOseberg: No problem.

FredrikOseberg:
Yeah something is definitely going on. Thanks for the review!

comountainclimber:
Closed in favor of https://github.com/CityOfZion/neon-wallet/pull/1424

Issue #1394, opened 2018-08-22
Implement new notifications design


comountainclimber:
See mocks... Also note that Error notifications should not autodismiss as this is bad UX

drptbl:
Tested, everything looks fine except one thing - error notifications are still dismissed automatically:
![notif](https://user-images.githubusercontent.com/8177587/46740792-ed6ed780-cca3-11e8-8b7c-d01c57aa2b19.gif)

comountainclimber:
As per discussion with product the error messages will self dismiss **but** with an increased timeout compared to v1... I will go ahead and add border radius to the notifications though good find!

drptbl:
Finished.

PR #1427, opened 2018-09-23
notifications-design

skunkpirates42:
Implement new notifications design #1394
Implements new notification design
I wrote some CSS and a JS object representing override styles for the react-notification-system
Made sure my design matched the mocks in the dev environment and I used yarn test
In app/modules/notifications.js, I changed the autoDismiss key to value of 0 in the error notification object. Line 106 -- this causes one of the tests to fail
I turned the auto dismiss off on the error notification to improve UX
- [ ] Unit tests written?


comountainclimber:
@skunkpirates42 nice work! You can fix the CI tests if you update the object on line 119 👍 

```
        {
          payload: {
            title: 'error',
            autoDismiss: AUTO_DISMISS_TIMEOUT,
            dismissible: false,
            id: 'notification_error_1',
            level: NOTIFICATION_LEVELS.ERROR,
            message: 'error',
            position: DEFAULT_POSITION
          },
          type: SHOW_NOTIFICATION
        }
```

app/modules/notifications.js
```diff
@@ -110,7 +115,8 @@ export const showWarningNotification = (args: NotificationArgsType) => (
     {
       title: DEFAULT_WARNING_TITLE,
       ...args,
-      level: NOTIFICATION_LEVELS.WARNING
+      level: NOTIFICATION_LEVELS.WARNING,
+      position: DEFAULT_POSITION
```
comountainclimber: Can we just keep this notification on the bottom? Thanks

Issue #1434, opened 2018-09-24
Wrong localhost reference after logging in to the wallet


drptbl:
**Wallet version or commit SHA:**
`feature/design-polish` branch (8925755)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Reproduction steps:**
1. Start the wallet
2. Login to any wallet

Seems like most of features are not working because of wrong reference to `app.asar/app/dist/[object%20Object]` at localhost. Take a look at the screenshot for more details.

**Screenshot/Video:**
![issue_localhost](https://user-images.githubusercontent.com/8177587/45977171-1018b380-c049-11e8-9aa0-45f15d5bc6c8.PNG)

comountainclimber:
I cant verify this or reproduce it... The compiled build that I get when running `yarn dist` does not have this error at all

drptbl:
@comountainclimber It was tested with compiled build that you've sent to me directly on Slack.

PR #1445, opened 2018-09-26
[WIP] Feature/fixes bug in login

comountainclimber:
It was discovered that when clearing all data from Neon wallet via deleting Neon in /Application Support that the application would crash on start. This PR addresses that bug and also the following:
- Styles the info notification
- Removes the Content-Security-Policy which was breaking code being executed from neon-js
- removes the override of window.eval (neon-js requires this) 👆 

FIXES https://github.com/CityOfZion/neon-wallet/issues/1434

### DIFF
--- a/app/containers/Notifications/overrideStyles.js
+++ b/app/containers/Notifications/overrideStyles.js
@@ -34,7 +34,12 @@ const overrideStyles = {
       backgroundColor: '#EE6D66'
     },
     warning: {
-      backgroundColor: '#FFCF48'
+      backgroundColor: '#FFCF48',
+      color: 'black'
+    },
+    info: {
+      backgroundColor: '#FFCF48',
+      color: 'black'
     }
   },
 
--- a/app/containers/Notifications/overrideStyles.js
+++ b/app/containers/Notifications/overrideStyles.js
@@ -35,11 +35,11 @@ const overrideStyles = {
     },
     warning: {
       backgroundColor: '#FFCF48',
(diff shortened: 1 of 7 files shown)


app/index.js
```diff
@@ -7,11 +7,6 @@ import Root from './components/Root'
 import './styles/tippy.compiled.global.css'
 import './styles/main.global.scss'
 
-// eslint-disable-next-line
-window.eval = global.eval = function() {
-  throw new Error('Sorry, this app does not support window.eval().')
-}
-
```
comountainclimber: This was causing an error originating from neon-js when trying to encrypt a key 👎 so I am eliminating this block for now

app/containers/Notifications/overrideStyles.js
```diff
@@ -34,7 +34,12 @@ const overrideStyles = {
       backgroundColor: '#EE6D66'
     },
     warning: {
-      backgroundColor: '#FFCF48'
+      backgroundColor: '#FFCF48',
+      color: 'black'
```
evgenyboxer: I think its better to use hex colors to be more consistent

app/containers/Notifications/overrideStyles.js
```diff
@@ -34,7 +34,12 @@ const overrideStyles = {
...
+    },
+    info: {
+      backgroundColor: '#FFCF48',
+      color: 'black'
```
evgenyboxer: same comment as above

PR #1446, opened 2018-09-26
Fixes contacts edit button

evgenyboxer:
N/A

The edit contact button was broken

I removed the extra function which swallowed the action

Tested it manually

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/app/components/Contacts/ContactsPanel/ContactsPanel.jsx
+++ b/app/components/Contacts/ContactsPanel/ContactsPanel.jsx
@@ -158,7 +158,7 @@ export default class ContactsPanel extends React.Component<Props, State> {
     this.setState({ sorting: option })
   }
 
-  handleEdit = (name: string) => () => {
+  handleEdit = (name: string) => {
     this.props.history.push(`/contacts/edit/${encodeURIComponent(name)}`)
   }
 


app/components/Contacts/ContactsPanel/ContactsPanel.jsx
```diff
@@ -158,7 +158,7 @@ export default class ContactsPanel extends React.Component<Props, State> {
     this.setState({ sorting: option })
   }
 
-  handleEdit = (name: string) => () => {
+  handleEdit = (name: string) => {
     this.props.history.push(`/contacts/edit/${encodeURIComponent(name)}`)
   }
```
comountainclimber: While you are here do you mind removing handleDelete (its no longer being used in this component as delete is an option in the edit view now)

app/components/Contacts/ContactsPanel/ContactsPanel.jsx
```diff
@@ -158,7 +158,7 @@ export default class ContactsPanel extends React.Component<Props, State> {
     this.setState({ sorting: option })
   }
 
-  handleEdit = (name: string) => () => {
+  handleEdit = (name: string) => {
```
comountainclimber: Much better thanks!

app/components/Contacts/ContactsPanel/ContactsPanel.jsx
dvdschwrtz: you can just remove it after we merge this PR. It looks good

PR #1447, opened 2018-09-26
[v2] Feature/filters out invalid hashes

comountainclimber:
`getTokenBalances()` will throw an error if any of the token hashes entered do not have a length of 40 chars... You can see the dozens of support requests coming in due to this simple mistake (there was no validation of script hashes in manage tokens and many many people just entered garbage here)

This PR addresses that issue by ignoring token hashes that do not have a length of 40


dvdschwrtz:
see my comment on the other, related PR

PR #1448, opened 2018-09-26
[dev] validate hash strings

comountainclimber:
See https://github.com/CityOfZion/neon-wallet/pull/1447


dvdschwrtz:
This is actually not the correct way to do this. There should be a method for validating script hashes in neon-js. Alternatively, you could try getting the address from the script hash and verify if the address is valid. I can look into neon-js further if you need help with this

PR #1454, opened 2018-09-28
Adds grouping, sorting and ui fixes to contacts

evgenyboxer:
N/A

Adds sorting, grouping and UI fixes to the contacts page. Additionally I have removed the unused function `handleDelete` from `ContactsPanel` and some used variable in the `EditContactPanel`

Converted the contacts object an an array collection, used lodash `groupBy` to group them by their first letter, then used `orderBy` to sort the groups and the contacts in the groups.

Tested it and verified UI/UX matched the required design.

Nope.

Nope.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Contacts/ContactsPanel/ContactsPanel.scss
+++ b/app/components/Contacts/ContactsPanel/ContactsPanel.scss
@@ -38,6 +38,7 @@
       display: flex;
       align-items: center;
       width: 420px;
+      justify-content: space-between;
     }
 
     .actions {
@@ -100,6 +101,12 @@
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
+}
+
+.groupHeader {
+  padding: 10px 24px;
+  background-color: #F6F7F8;
+  font-weight: bold;
 }
 
 .headerSelect {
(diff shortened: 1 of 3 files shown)


app/components/Contacts/ContactsPanel/ContactsPanel.scss
```diff
@@ -102,6 +103,12 @@
   transform: translate(-50%, -50%);
 }
 
+.groupHeader {
+  padding: 10px 24px;
+  background-color: #F6F7F8;
```
comountainclimber: Just curious where you got this color from in the mocks its `#979797` which is also not on the palette 😆... Either way we should probably be referencing one of the color vars here so that we can swap out if needed for alternate themes in the future

comountainclimber:
Pulled down the branch and overall looks and functions great! Nice work @evgenyboxer 💪 Just had one nit picky comment on styles

app/components/Contacts/ContactsPanel/ContactsPanel.scss
evgenyboxer: As discussed, we will consolidate this when we get to the dark theme.

Issue #1457, opened 2018-09-28
Update button styles according to design review


comountainclimber:

drptbl:
Finished.

PR #1472, opened 2018-09-29
Feature/update button styles

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1457

Updates button styles according to design review

- [ ] Unit tests written?

### DIFF
--- a/app/components/Button/Button.scss
+++ b/app/components/Button/Button.scss
@@ -73,7 +73,7 @@
 
   .centeredLabel {
     position: absolute;
-    align-self: center !important;
+    align-self: center;
   }
 
   .icon {
--- a/app/components/Button/Button.scss
+++ b/app/components/Button/Button.scss
@@ -1,7 +1,15 @@
 @import '../../styles/variables';
 
 .active {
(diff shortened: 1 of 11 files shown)


app/components/Button/Button.scss
```diff
@@ -46,42 +47,34 @@
...
       background-origin: border-box;
       background-clip: content-box, border-box;
       border-color: transparent;
       border-color: transparent;
-      color: white;
-
-      path {
-        fill: #fff;
-      }
+      background: #f2f2f2 !important;
```
evgenyboxer: I guess we have to use `!important`?

app/components/Button/Button.scss
```diff
@@ -46,42 +47,34 @@
...
     }
   }
 
   span {
     width: 100%;
   }
 
+  .centeredLabel {
+    position: absolute;
+    align-self: center !important;
```
evgenyboxer: same ^

comountainclimber:
Thanks for the review @evgenyboxer 👍

Issue #1469, opened 2018-09-28
Update sidebar according to design review


comountainclimber:

drptbl:
Finished.

PR #1478, opened 2018-10-03
[design-review] updates side bar style

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1469

- [ ] Unit tests written?


app/containers/App/Sidebar/Sidebar.scss
```diff
@@ -73,7 +73,9 @@ $size: 64px;
   &.active,
   &:hover {
     background: #fff;
-    border-image: linear-gradient(to bottom, #66eb8e, #6bdaf6) 1 100%;
+    //border-image: linear-gradient(to bottom, #66eb8e, #6bdaf6) 1 100%;
```
dvdschwrtz: minor comment but I would remove all comments for code being merged

Issue #1466, opened 2018-09-28
Update tx history history according to design review


comountainclimber:

drptbl:
Finished.

PR #1479, opened 2018-10-03
[design-review] transaction history updates

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1466

- Updates tx history UI per design review
- fixes broken layout of add contact modal
- Parses transactions to see if they come from an address in contacts and adds label with tooltip if found

screen shot 2018-10-02 at 10 37 32 pm
screen shot 2018-10-02 at 10 37 42 pm

- [ ] Unit tests written?

### DIFF
--- a/app/components/Blockchain/Transaction/Transaction.jsx
+++ b/app/components/Blockchain/Transaction/Transaction.jsx
@@ -3,6 +3,7 @@ import React from 'react'
 import classNames from 'classnames'
 import moment from 'moment'
 
+import Button from '../../Button'
 import { openExplorerTx } from '../../../core/explorer'
 import styles from './Transaction.scss'
 import { ASSETS } from '../../../core/constants'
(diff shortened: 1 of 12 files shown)


app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -34,17 +39,45 @@ export default class Transaction extends React.Component<Props> {
...
-          {txid.substring(0, 32)}
-        </span>
+          View
+        </Button>
       </div>
     )
   }
 
+  displayModal = () => {
+    const { showAddContactModal, tx } = this.props
```
dvdschwrtz: you can do this all in one line const { showAddContactModal, tx: { to } } = this.props

app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -34,17 +39,45 @@ export default class Transaction extends React.Component<Props> {
         </div>
         <div className={styles.txLabelContainer}>{label}</div>
         <div className={styles.txAmountContainer}>{amount}</div>
-        <div className={styles.txToContainer}>{to}</div>
-        <span
-          className={classNames(styles.txidLink, className)}
+        <div className={styles.txToContainer}>
+          {to}
+          {to !== 'NETWORK FEES' && (
```
dvdschwrtz: both this and to === 'NETWORK FEES' can be combined in one ternary expression. to !== 'NETWORK FEES' ? () : ()

also, you might want to make NETWORK FEES a variable in case you change the text, it will apply to where you created it originally and inside this ternary as well

comountainclimber: Because the closing </div> is not in the ternary im going to leave these as two separate logical operators for now... but I did go ahead and leverage a boolean value passed down from the action instead of checking for equality against that hard coded string

app/components/Blockchain/Transaction/Transaction.jsx
comountainclimber: 👍

Issue #1462, opened 2018-09-28
Update dashboard UI according to design review


comountainclimber:

drptbl:
Finished.

PR #1481, opened 2018-10-04
[design-review] dashboard

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1462

- Updates price history graph to reflect design feedback from review
- Updates UI layout in price history chart
- Implements tooltip design in price history chart
- Various color updates
- Updates layout in asset balances panel
- Fixes bug related to color of secondary button when disable in Home.jsx
- Fixes a layout issue with ZeroAssets.jsx

- [ ] Unit tests written?


app/components/Dashboard/PriceHistoryPanel/PriceHistoryChart.jsx
```diff
@@ -67,86 +66,52 @@ export default class PriceHistoryChart extends React.Component<Props, State> {
...
-        >
-          {change >= 0 && '+'}
-          {(change * 100).toFixed(2)}%
-        </text>
-      </BoundingBox>
+  tooltipContent = (tooltipProps: Object) =>
+    !!tooltipProps.payload.length && (
+      <div className={styles.tooltipContainer}>
+        <div className={styles.tooltipTime}>
+          {this.formatLabel(tooltipProps.payload[0].payload.time)}
```
dvdschwrtz: I think using a lodash get here would be better incase the payload ever changes

app/components/Dashboard/PriceHistoryPanel/PriceHistoryChart.jsx
```diff
@@ -67,86 +66,52 @@ export default class PriceHistoryChart extends React.Component<Props, State> {
...
+        </div>
+        <div className={styles.tooltipPrice}>
+          {/* $FlowFixMe */}
+          {this.formatPrice(tooltipProps.payload[0].payload.close, null)}
```
dvdschwrtz: a lodash get here would be better incase the payload ever changes

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -101,4 +109,49 @@ export default class PriceHistoryPanel extends React.Component<Props> {
...
+    return (
+      <div className={styles.currentPrice}>
+        {this.formatPrice(staticPrice, formatFiat)}
+      </div>
+    )
+  }
+
+  formatPrice = (
+    price: number,
+    formatter: Function = formatThousands
```
dvdschwrtz: generally I think it is better to be more descriptive than Function. But it depends on how many various formatters you might use and their various outputs. This might be ok

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -101,4 +109,49 @@ export default class PriceHistoryPanel extends React.Component<Props> {
...
+  ): string => {
+    const { symbol } = CURRENCIES[this.props.currency]
```
dvdschwrtz: what if there is no symbol for that currency. consider lodash get and some defensive programming around this case

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -101,4 +109,49 @@ export default class PriceHistoryPanel extends React.Component<Props> {
...
+    const { prices, priceKey } = this.props
+    return prices[0][priceKey]
+  }
+
+  getPriceChange = () =>
+    (this.getLatestPrice() - this.getInitialPrice()) / this.getInitialPrice()
+
+  getLatestPrice = (): number => {
+    const { prices, priceKey } = this.props
+    return prices[prices.length - 1][priceKey]
```
dvdschwrtz: is there always going to be a length to prices?

Issue #1467, opened 2018-09-28
Update send according to design review


comountainclimber:

drptbl:
Finished.

PR #1483, opened 2018-10-04
[design-review] send

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1467

- Updates layout and sizing of send panel
- Repositions coming soon tooltip
- Updates icon colors in panel header
- Disables "Add recipient" until details are filled in per row
-  Bumps arbitrary recipient cap up to 10 from 5

- [ ] Unit tests written?

### DIFF
--- a/app/components/Send/SendPanel/SendPanelHeader/index.jsx
+++ b/app/components/Send/SendPanel/SendPanelHeader/index.jsx
@@ -19,7 +19,8 @@ type Props = {
   sendError: boolean,
   noSendableAssets: boolean,
   disabled: boolean,
-  hasNetworkFees: boolean
+  hasNetworkFees: boolean,
+  maxNumberOfRecipients: number
 }
 
 const SendPanelHeader = ({
@@ -31,12 +32,13 @@ const SendPanelHeader = ({
   sendSuccess,
   noSendableAssets,
   disabled,
-  hasNetworkFees = false
+  hasNetworkFees = false,
+  maxNumberOfRecipients
 }: Props) => {
   const numberOfItems = sendRowDetails.length
 
   let headerTitle = 'Select Assets'
-  let headerSubtitle = `${numberOfItems} of 10 Recipients`
+  let headerSubtitle = `${numberOfItems} of ${maxNumberOfRecipients} Recipients`
   let buttons = (
     <div className={styles.sendPanelHeaderButtons}>
       <Tooltip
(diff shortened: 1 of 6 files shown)


app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
```diff
@@ -143,7 +143,7 @@ class SendRecipientListItem extends Component<Props> {
 
     return (
       <li className={styles.sendRecipientListItem}>
-        <div className={styles.rowNumber}>{`0${index + 1}`}</div>
+        <div className={styles.rowNumber}>{`${`0${index + 1}`.slice(-2)}`}</div>
```
dvdschwrtz: this feels weird with .slice(-2) but not sure what is happening. if everything is working correctly it should be fine

dvdschwrtz: this feels weird with .slice(-2) but not sure what is happening. if everything is working correctly it should be fine

app/containers/Send/Send.jsx
```diff
@@ -150,7 +150,7 @@ export default class Send extends React.Component<Props, State> {
     this.setState((prevState: Object) => {
       const newState = [...prevState.sendRowDetails]
 
-      if (newState.length < 5) {
+      if (newState.length < 10) {
```
dvdschwrtz: maybe 10 should be a variable that you can reuse throughout. This is the number of rows or recipients or something, right? What if we want to change it to 12 or 15. It would be easier if it was only defined in one place

app/components/Send/SendPanel/SendPanelHeader/index.jsx
```diff
@@ -29,18 +30,19 @@ const SendPanelHeader = ({
...
   sendSuccess,
   noSendableAssets,
+  disabled,
   hasNetworkFees = false
 }: Props) => {
   const numberOfItems = sendRowDetails.length
 
   let headerTitle = 'Select Assets'
-  let headerSubtitle = `${numberOfItems} of 5 Recipients`
+  let headerSubtitle = `${numberOfItems} of 10 Recipients`
```
dvdschwrtz: see comment below about making 10 a variable

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
comountainclimber: Its just taking the last two digits of whatever string is generated here and always prefixing 0

app/containers/Send/Send.jsx
comountainclimber: 👍

app/components/Send/SendPanel/SendPanelHeader/index.jsx
comountainclimber: 👍

Issue #1453, opened 2018-09-27
Hide password and private key by default on final screen of wallet creation


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (8a1a780)

Wallet password and private key should be hidden by default on final screen of wallet creation (because it's hidden by default on previous screen while typing the password).

**Screenshot/Video:**
![create_wallet](https://user-images.githubusercontent.com/8177587/46165653-7a239980-c291-11e8-9635-29a45fbdce68.gif)

comountainclimber:
Thanks for reporting this after careful consideration lets just stick with hiding the password for now in this view adding to the v2 board now...

drptbl:
Not ready for QA yet.

drptbl:
Fixed.

PR #1487, opened 2018-10-06
Implement password hiding

zsiglin:
#1453 

> Thanks for reporting this after careful consideration lets just stick with hiding the password for now in this view adding to the v2 board now...

Per @comountainclimber's comment, this simply passes [type='password'] into the password input. It doesn't do the same for the private key. I wasn't sure if your intention was to have the display feature for this togglable, so I made the simplest assumption based on my interpretation of the comment. If you in fact intended for the hide/show feature to be there as well that can be addressed.

Tested locally.

screen shot 2018-10-05 at 9 52 39 pm


zsiglin:
👍 I'll slot this in after I finish up the QR code stuff.

comountainclimber:
closing in favor of https://github.com/CityOfZion/neon-wallet/pull/1504

Issue #1459, opened 2018-09-28
Add custom script hashes in settings


comountainclimber:
We should support the ability to add script hashes in settings for testing in all networks other than mainnet (currently these get pulled down via neo-tokens so may not be necessary for MVP)

drptbl:
Not ready for QA yet. Waiting for refactor according to code review.

drptbl:
Tested, found issues:
#1496, #1497, #1498

Issue #1458, opened 2018-09-28
Add script hashes in token sale flow


comountainclimber:
When addressed we should also remove the stale list of hardcoded ICO script hashes

drptbl:
Tested.

PR #1495, opened 2018-10-09
Feature/manage tokens modal

comountainclimber:

- This PR updates the select dropdown component styles (hopefully for the last time) fixing a bug causing the value to "jump".
**FIXED**
fixed
fixed-2

- It also adds a redesigned "manage tokens" modal to be used in both settings and in token sale.
- Updates validation and logic in manage tokens modal
- Adds a "transparent" prop to `StyledReactSelect.jsx`

- Updates the alignment of "header" buttons throughout application
screen shot 2018-10-08 at 10 46 41 pm

https://github.com/CityOfZion/neon-wallet/issues/1458
https://github.com/CityOfZion/neon-wallet/issues/1459

Lots of manual testing...
screen shot 2018-10-08 at 9 15 16 pm
screen shot 2018-10-08 at 9 15 22 pm

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/TokenModal/TokenModal.jsx
+++ b/app/components/Modals/TokenModal/TokenModal.jsx
@@ -1,6 +1,6 @@
 // @flow
 import React, { Component } from 'react'
-import { reject, noop } from 'lodash-es'
+import { reject, noop, isEqual } from 'lodash-es'
 
 import { getNewTokenItem, validateTokens } from './utils'
 
@@ -95,13 +95,19 @@ class TokenModal extends Component<Props, State> {
   shouldDisableSaveButton = () => {
     const { tokens } = this.state
 
-    const invalidUserGeneratedTokens = tokens
-      .filter((token: TokenItemType) => token.isUserGenerated)
-      .filter((token: TokenItemType) => !token.scriptHash.length)
+    const invalidUserGeneratedTokens = tokens.filter(
+      (token: TokenItemType) =>
+        token.isUserGenerated && !token.scriptHash.length
+    )
 
-    if (tokens === this.props.tokens) return true
-    if (invalidUserGeneratedTokens.length) return true
-    if (tokens.length === this.props.tokens.length) return true
+    if (
+      isEqual(tokens, this.props.tokens) ||
+      invalidUserGeneratedTokens.length ||
+      tokens.length === this.props.tokens.length
+    ) {
+      return true
+    }
+
     return false
   }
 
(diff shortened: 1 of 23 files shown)


app/components/Modals/TokenModal/TokenModal.jsx
```diff
@@ -90,28 +92,48 @@ class TokenModal extends Component<Props, State> {
...
+  shouldDisableSaveButton = () => {
+    const { tokens } = this.state
+
+    const invalidUserGeneratedTokens = tokens
+      .filter((token: TokenItemType) => token.isUserGenerated)
+      .filter((token: TokenItemType) => !token.scriptHash.length)
+
+    if (tokens === this.props.tokens) return true
+    if (invalidUserGeneratedTokens.length) return true
+    if (tokens.length === this.props.tokens.length) return true
```
evgenyboxer: since all three conditions return `true` then you can write them in the same condition
if (tokens === this.props.tokens || invalidUserGeneratedTokens.length || tokens.length === this.props.tokens.length) { ... }

also, I assume `tokens` is an object? if so, better use lodash `isEqual`

app/components/Modals/TokenModal/TokenModal.jsx
```diff
@@ -90,28 +92,48 @@ class TokenModal extends Component<Props, State> {
     })
   }
 
+  shouldDisableSaveButton = () => {
+    const { tokens } = this.state
+
+    const invalidUserGeneratedTokens = tokens
+      .filter((token: TokenItemType) => token.isUserGenerated)
+      .filter((token: TokenItemType) => !token.scriptHash.length)
```
evgenyboxer: you can use just one filter for both conditions - 
```
.filter((token: TokenItemType) => token.isUserGenerated && !token.scriptHash.length)
```

app/containers/Dashboard/Dashboard.jsx
```diff
@@ -59,18 +61,18 @@ export default class Dashboard extends Component<Props> {
           renderRightContent={() => (
             <div className={classNames(styles.dashboardHeaderButonContainer)}>
               <NavLink id="wallet-manager" exact to={ROUTES.WALLET_MANAGER}>
-                <span> Manage Wallets </span>
                 <Wallet id="manage-wallets" />
+                <span> Manage Wallets </span>
```
evgenyboxer: is the space around `Manage Wallets` intentional? or this can be removed?

app/components/Modals/TokenModal/TokenModal.jsx
comountainclimber: 👍

app/components/Modals/TokenModal/TokenModal.jsx
comountainclimber: 👍

PR #1500, opened 2018-10-10
Refactor/token sale

FredrikOseberg:

- [ ] Unit tests written?


app/components/PanelHeaderButton/PanelHeaderButton.jsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+class PanelHeaderButton extends Component<Props> {
+  render() {
+    const { renderIcon, onClick, buttonText, disabled, className } = this.props
+
+    return (
+      <button
+        type="button"
+        className={`${styles.panelHeaderButton} ${className || ''}`}
```
evgenyboxer: Could you use `classnames` for this? (for consistency)

app/components/PanelHeaderButton/PanelHeaderButton.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+type Props = {
+  onClick: () => void,
+  renderIcon: () => React$Node,
+  buttonText: string,
+  className?: string,
+  disabled?: boolean
+}
+
+class PanelHeaderButton extends Component<Props> {
```
comountainclimber: Great thinking @FredrikOseberg!

app/containers/Buttons/RefreshButton/RefreshButton.jsx
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    const { loadWalletData, loading } = this.props
+
+    return (
+      <div className={styles.refreshButton}>
+        <span
+          onClick={loading ? null : loadWalletData}
+          className={styles.refreshButtonSpan}
+        >
+          {' '}
```
comountainclimber: Pretty sure we do not need these empty strings

comountainclimber:
Overall this looks really good... I will continute to review what you have already done here however while you are polishing off token sale I was hoping you could finish the implementation of the routing set up in `TokenSale.jsx` as you can see :

```
  handlePurchase = () => {
    this.setState({ inputErrorMessage: '' }, () => {
      const validFields = this.isValid()

      if (validFields) {
        this.setStep(TOKEN_SALE_CONFIRM)
        this.props.history.push('/token-sale-confirm')
      }
    })
  }
```

Routing has already been set up but is not in place for the final confirmation step. If you could address this that would be fantastic!

app/containers/Buttons/RefreshButton/RefreshButton.jsx
dvdschwrtz: they often get added by style formatters automatically

app/components/TokenSale/TokenSalePanel/TokenSalePanel.jsx
```diff
@@ -6,9 +6,12 @@ import TokenSaleSelection from './TokenSaleSelection/TokenSaleSelection'
 import TokenSaleConditions from './TokenSaleConditions/TokenSaleConditions'
 import DialogueBox from '../../DialogueBox'
 import Button from '../../Button'
+// import Tooltip from '../../components/Tooltip/Tooltip'
```
dvdschwrtz: very minor but please remove all commented code before merging

FredrikOseberg: Of course. Good catch, we should setup lint rules for commented code. @comountainclimber

app/containers/Buttons/RefreshButton/RefreshButton.jsx
FredrikOseberg: @dvdschwrtz is right. This is Prettier's work.

app/util/getICOTokens.js
```diff
@@ -1,22 +1,24 @@
...
+  /* eslint-disable-next-line */
+  const tokens = get(state, 'spunky.settings.data.tokens')
+  if (!Array.isArray(tokens)) return null
+  const userGeneratedTokens = tokens.filter(token => token.isUserGenerated)
+
+  const combinedTokenList = [
+    ...userGeneratedTokens,
+    ...ICOTokenList.ICOTokens
+  ].map(token => ({
+    token: token.symbol,
```
comountainclimber: I dont think user added scripts will ever have a symbol at this point we probably want to call `getToken`  here http://cityofzion.io/neon-js/docs/en/api/plugin/nep5.html#methods to obtain symbol

app/util/getICOTokens.js
```diff
@@ -1,22 +1,24 @@
...
+    supply: token.supply || null,
+    id: token.id,
+    scriptHash: token.scriptHash
+  }))
```
comountainclimber: We need to filter by network here based on current selected network id

ICOTokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ICOTokens": [
+    {
```
comountainclimber: Lets leave this array empty... we cant ship to production with this script hash in the app (its not valid) also this data structure needs to include network id info

app/components/Send/SendPanel/index.jsx
```diff
@@ -35,7 +35,7 @@ type Props = {
   handleSubmit: () => any,
   handleSend: () => any,
   clearErrors: (index: number, field: string) => any,
-  addRow: (row: Object) => any,
+  addRow: (row?: Object) => any,
```
comountainclimber: When I remove this change on your branch I do not see any flow errors I think we should leave this as is as row must always be defined (and is always defined)

app/util/getICOTokens.js
comountainclimber: we already do this in balancesActions.js for example...

app/util/getICOTokens.js
```diff
@@ -1,22 +1,24 @@
...
-  ]
+import { get } from 'lodash'
+import store from '../store/configureStore'
+
+import ICOTokenList from '../../ICOTokens.json'
+
+export const getICOTokens = () => {
+  const state = store.getState()
+  /* eslint-disable-next-line */
+  const tokens = get(state, 'spunky.settings.data.tokens')
```
comountainclimber: Everywhere else where we have functions that reference tokens on the state tree they are passed in as function params rather than accessed in this manner... Can we please update this so that if behaves like the rest of the app?

Issue #1498, opened 2018-10-10
After adding a custom token with wrong script hash user gets "Failed to load. Retry?" messages instead of asset balances


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (6c03338)

**Reproduction steps:**
1. Start wallet
2. Login to any wallet
3. Click on "Settings" button in the menu
4. Click on "Add Token" button
5. Click on "Add a new token" button
6. Type anything inside input field, for example "test"
7. Click on "Save" button
8. Click on "Wallet" button in the menu

**Description:**
After saving a custom token with wrong script hash, wallet dashboard gets "broken" and stops showing any balances for user - it shows "Failed to load. Retry?" messages instead.

**Screenshot:**
<img width="1194" alt="screenshot 2018-10-10 at 15 26 39" src="https://user-images.githubusercontent.com/8177587/46740386-fdd28280-cca2-11e8-961c-1e39c8f2036f.png">

drptbl:
FIxed :)!

Issue #1496, opened 2018-10-10
Don't switch wallet network while switching network in custom tokens modal


drptbl:
I think that wallet network shouldn't be switched after switching the network inside custom tokens modal. This may confuse users, because if you want to switch wallet network, you should use drop-down on dashboard.

**Screenshot:**
<img width="611" alt="screenshot 2018-10-10 at 15 34 11" src="https://user-images.githubusercontent.com/8177587/46739959-f65ea980-cca1-11e8-97c0-2bdb452988b2.png">

drptbl:
Fixed.

PR #1501, opened 2018-10-10
Feature/design review 2

comountainclimber:
This PR incorporates all of the design review feedback including the new receive functionality **except for the settings UI refactor** It also addresses a number of other items discovered by QA.

- https://github.com/CityOfZion/neon-wallet/issues/1496
- https://github.com/CityOfZion/neon-wallet/issues/1498
- https://github.com/CityOfZion/neon-wallet/issues/1502
- https://github.com/CityOfZion/neon-wallet/issues/1394

It also clears up a bug I found related to abstracts represented as "receive" in tx history (previously it was not showing who it was from just always the users wallet address). Furthermore rendering logic in tx history is simplified

The network switch component is also refactored and simplified so that it can propagate network switch event or serve as a simple select

tabbed-receive

- [ ] Unit tests written?


app/components/Inputs/AssetInput/AssetInput.jsx
```diff
@@ -1,79 +1,44 @@
...
+  selectedAsset: Object
 }
 
 export default class AssetInput extends React.Component<Props, State> {
-  static defaultProps = {
-    onChange: noop,
-    onFocus: noop,
-    onBlur: noop
-  }
+  parsedSymbols = this.props.symbols.map(symbol => ({
```
evgenyboxer: I think its better if this was a function that would get symbols and check if its an array and then return a mapped result (outside of the class)

app/components/Modals/TokenModal/utils.js
```diff
@@ -8,14 +8,28 @@ export const getNewTokenItem = (networkId: string) => ({
   isUserGenerated: true
 })
 
+const ensureHex = (token: string): boolean => {
+  const hexRegex = /^([0-9A-Fa-f]{2})*$/
```
evgenyboxer: Its better to move the `hexRegex` declaration out of the function (no need to recreate it every time, just reuse it)

comountainclimber: 👍 will definitely move this to a util file (I meant to do that but was moving quickly) thanks

app/components/Inputs/AssetInput/AssetInput.jsx
comountainclimber: 👍

Issue #1460, opened 2018-09-28
Ability to scan QR and autofill send details (nep9)


comountainclimber:

drptbl:
Works only for NEO and GAS currently, throws error and infinite loader in any other asset:
<img width="1194" alt="screenshot 2018-10-26 at 20 50 45" src="https://user-images.githubusercontent.com/8177587/47586587-f4167380-d960-11e8-8a44-e3fcda010953.png">
<img width="1198" alt="screenshot 2018-10-26 at 20 51 04" src="https://user-images.githubusercontent.com/8177587/47586588-f4167380-d960-11e8-9fba-b8a6d39475b3.png">

comountainclimber:
Fixed

PR #1507, opened 2018-10-11
Feature/qr code reader

zsiglin:
 #1460 

Implements QR code reading via camera

Tested manually

- Per @comountainclimber we will implement URL decoding later
- In MacOS Mojave the camera module crashes when running the application from iTerm in development mode due to security restrictions. This can be avoided by starting the dev environment from Terminal. This is a known problem and occurs in v1 of this application as well. Production builds are not affected.
- The copy has been adjusted to accommodate the removal of the URL input.


app/components/Modals/SendModal/ReadCode/ReadCode.jsx
```diff
@@ -0,0 +1,81 @@
...
+    return <img src={CozDonationQrCode} alt="Donate to CoZ" />
+  }
+
+  render() {
+    const { scannerActive } = this.state
+
+    return (
+      <div className={baseStyles.contentContainer}>
+        <div className={baseStyles.header}>
+          <GridIcon className={baseStyles.icon} />
```
comountainclimber: This icon should be `#69e27e` (neon green)

app/components/Modals/SendModal/ConfirmDetails/ConfirmDetails.jsx
```diff
@@ -0,0 +1,53 @@
...
+    this.props.recipientData[key] || 'Not specified'
+
+  render() {
+    return (
+      <div className={baseStyles.contentContainer}>
+        <div className={baseStyles.header}>
+          <GridIcon className={baseStyles.icon} />
+          <div className={baseStyles.title}>QR Code Identified!</div>
+        </div>
+
```
comountainclimber: Lets add a block displaying the amount requested 👍

comountainclimber:
Overall this looks fantastic... Some minor comments on some things that should be addressed after that we can merge this in 🎉 👍

app/util/hashToSymbol.js
```diff
@@ -0,0 +1,9 @@
+import { TOKENS } from '../core/constants'
+
+const hashToSymbol = hash =>
+  Object.keys(TOKENS).find(currentSymbol => {
+    const currentHash = TOKENS[currentSymbol].networks['1'].hash
```
comountainclimber: Ideally we would actually pass the network id into this function as well for testing on other networks in the future (and testnet right now)

app/util/parseQRCode.js
```diff
@@ -0,0 +1,38 @@
+import hashToSymbol from './hashToSymbol'
+
+const INVALID_FORMAT = 'Invalid format'
+const INVALID_PROTOCOL = 'Invalid protocol'
+const MISSING_ADDRESS = 'Missing recipient address'
+const UNRECOGNIZED_ASSET = 'Unrecognized asset'
```
comountainclimber: These are awesome nice job

app/components/Modals/SendModal/ConfirmDetails/ConfirmDetails.jsx
```diff
@@ -0,0 +1,53 @@
...
+import styles from './ConfirmDetails.scss'
+
+type Props = {
+  recipientData: Object,
+  confirmAndClose: () => any
+}
+
+export default class ConfirmDetails extends React.Component<Props> {
+  getRecipientData = (key: string) =>
+    this.props.recipientData[key] || 'Not specified'
```
dvdschwrtz: prefer lodash get over this syntax - if recipientData is not defined there will be an error get(this.props.recipientData, key)

app/components/Modals/SendModal/ReadCode/ReadCode.jsx
```diff
@@ -0,0 +1,81 @@
+// @flow
+import React from 'react'
+import classNames from 'classnames'
+
+import Button from '../../../Button'
+import QrCodeScanner from '../../../QrCodeScanner'
+import GridIcon from '../../../../assets/icons/grid.svg'
+import CozDonationQrCode from '../../../../assets/images/coz-donation-qr-code.png'
```
dvdschwrtz: for a future PR it would be nice to alias the path for assets so we don't have to find the exact path. 'assets/images...'. No need to change anything. Just some food for thought

app/components/Modals/SendModal/ReadCode/ReadCode.scss
```diff
@@ -0,0 +1,65 @@
...
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 100%;
+    height: 170px;
+    border-radius: 5px;
+    background-color: $light-grey;
+
+    :nth-child(1){
```
dvdschwrtz: doing it this way is great but its a bit less explicit to know that the divs are being styled here. In general, I think everything should be as clear as possible in an open source project; however I think its ok to do this if you don't want to change it

app/components/Send/SendPanel/index.jsx
```diff
@@ -63,9 +65,11 @@ const SendPanel = ({
   handleAddPriorityFee,
   fees,
   address,
-  maxNumberOfRecipients
+  maxNumberOfRecipients,
+  showSendModal,
+  pushQRCodeData
 }: Props) => {
-  function shouldDisableSendButton(sendRowDetails) {
+  const shouldDisableSendButton = sendRowDetails => {
```
dvdschwrtz: sendRowDetails should have a type here

app/util/hashToSymbol.js
dvdschwrtz: I think you can do this using lodash get as well. its either get(TOKENS, `${currentSymbol}.networks.1.hash`) or get(TOKENS, [currentSymbol, 'networks', '1', 'hash'], default goes here)

app/components/Send/SendPanel/index.jsx
zsiglin: This passes flow validation and actually wasn't even a part of the functionality implemented within this PR. I'm happy to add it, but for some reason flow was barking at me for adding types to a function defined within a stateless component.

zsiglin:
Thanks for the feedback @comountainclimber and @dvdschwrtz. I've addressed it in my latest commit.

app/components/Modals/SendModal/ReadCode/ReadCode.scss
zsiglin: I went back and forth on this a few times when I was writing it. I agree that explicit is probably better in this context. Updated.

app/components/Modals/SendModal/ReadCode/ReadCode.jsx
zsiglin: I've added an alias for `assets` to clean-up these admittedly ugly imports.

app/components/Modals/SendModal/ConfirmDetails/ConfirmDetails.jsx
```diff
@@ -14,7 +15,7 @@ type Props = {
 
 export default class ConfirmDetails extends React.Component<Props> {
   getRecipientData = (key: string) =>
-    this.props.recipientData[key] || 'Not specified'
+    get(this.props.recipientData, key) || 'Not specified'
```
zsiglin: Just to call this out before it is commented on - I'm aware the third argument in `get` acts as a default, but this only works for `undefined` values _not_ `null`. In this case, the function that is responsible for generating `recipientData` returns `null`'s for empty keys (the `.get` method on the native implementation of `URL` is the source of this). Since this doesn't cause a problem anywhere else I thought it'd be cleaner to write it this way rather than having to convert them to `undefined`'s in `parseQrCodeData`.

PR #1512, opened 2018-10-14
leverages crypto compare for pricing data

comountainclimber:
Because coinmarketcap is deprecating their public facing API this PR switches the pricing data to be pulled from cryptocompare instead.

- [ ] Unit tests written?

### DIFF
--- a/app/actions/pricesActions.js
+++ b/app/actions/pricesActions.js
@@ -2,32 +2,44 @@
 import axios from 'axios'
 import { createActions } from 'spunky'
 
+import { getDefaultTokens } from '../core/nep5'
 import { DEFAULT_CURRENCY_CODE } from '../core/constants'
 
 type Props = {
   currency?: string
 }
 
-function mapPrices(tickers, currency) {
+function mapPrices(pricingData: Array<Object>, currency) {
   const mapping = {}
-
-  tickers.forEach(ticker => {
-    mapping[ticker.symbol] = parseFloat(
-      ticker[`price_${currency.toLowerCase()}`]
-    )
+  const upperCasedCurrency = currency.toUpperCase()
+  pricingData.forEach((price: { currency: { upperCasedCurrency: Object } }) => {
+    const priceInSelectedCurrency = price[upperCasedCurrency]
+    if (price && priceInSelectedCurrency) {
+      mapping[priceInSelectedCurrency.FROMSYMBOL] = parseFloat(
+        priceInSelectedCurrency.PRICE
+      )
+    }
   })
 
   return mapping
 }
 
-function getPrices(currency) {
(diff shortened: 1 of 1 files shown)


app/actions/pricesActions.js
```diff
@@ -2,32 +2,43 @@
...
+    .concat([ASSETS.NEO, ASSETS.GAS])
+    .join(',')
+  const url = `https://min-api.cryptocompare.com/data/pricemultifull?fsyms=${joinedTokens}&tsyms=${currency.toUpperCase()}`
 
-  return axios.get(url).then(response => {
-    const { data } = response
-    if (data.error) throw new Error(data.error)
-    return mapPrices(data, currency)
-  })
+  const priceDataResponse = await axios
```
evgenyboxer: could you wrap this inside a try/catch (also put `pricingArray` inside it), that will help catching the errors that might arise if the response changes, etc.

app/actions/pricesActions.js
```diff
@@ -2,32 +2,43 @@
...
-function mapPrices(tickers, currency) {
+function mapPrices(pricingData: Array<any>, currency) {
   const mapping = {}
-
-  tickers.forEach(ticker => {
-    mapping[ticker.symbol] = parseFloat(
-      ticker[`price_${currency.toLowerCase()}`]
-    )
+  const upperCasedCurrency = currency.toUpperCase()
+  pricingData.forEach((price: { currency: { upperCasedCurrency: Object } }) => {
```
evgenyboxer: Can probably replaced by reduce
```
const mapping  = pricingData.reduce((accumulator, item) => {
     const priceInSelectedCurrency = price[upperCasedCurrency]
     if (price && priceInSelectedCurrency) {
       accumulator[priceInSelectedCurrency.FROMSYMBOL] = parseFloat(
          priceInSelectedCurrency.PRICE
       )
     }
     return accumulator
   }, {})
```

app/actions/pricesActions.js
```diff
@@ -2,32 +2,43 @@
...
 
-function mapPrices(tickers, currency) {
+function mapPrices(pricingData: Array<any>, currency) {
   const mapping = {}
-
-  tickers.forEach(ticker => {
-    mapping[ticker.symbol] = parseFloat(
-      ticker[`price_${currency.toLowerCase()}`]
-    )
+  const upperCasedCurrency = currency.toUpperCase()
```
dvdschwrtz: this method should just be a reduce in my opinion rather than forEach building a new map

app/actions/pricesActions.js
dvdschwrtz: we could also use joi to validate response data

app/actions/pricesActions.js
comountainclimber: updated to use reduce thanks

PR #1514, opened 2018-10-15
Add NNS name lookup to send screen address input

nickfujita:
This implements a name to address resolutions using NEO Name Service.

When a user inputs a `.neo` domain name into the address field on the send screen, a quick attempt will be made to resolve it to an address. If successful, the field will be populated with the appropriate address. If the name is not found a error message will display on the input field.

screen recording 2018-10-16 at 03 16 am

### DIFF
--- a/app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
+++ b/app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
@@ -30,11 +30,12 @@ type Props = {
 
 type State = {
   nnsError: string,
-  isNnsResolving: boolean,
+  isNnsResolving: boolean
 }
 
 class SendRecipientListItem extends Component<Props, State> {
-  state = { addressLabel: '', isNnsResolving: false }
+  state = { nnsError: '', isNnsResolving: false }
+
   nnsFetchTimeout = undefined
 
   handleFieldChange = (e: Object) => {
@@ -58,25 +59,27 @@ class SendRecipientListItem extends Component<Props, State> {
 
     const { name } = e.target
     let { value } = e.target
-    if (value > max) value = max
-    name === 'address' && this.setState({ nnsError: '' })
+    if (name === 'address') this.setState({ nnsError: '' })
     if (name === 'address' && value.endsWith('.neo')) {
       this.setState({ isNnsResolving: true })
-      this.nnsFetchTimeout && clearTimeout(this.nnsFetchTimeout)
+      if (this.nnsFetchTimeout) clearTimeout(this.nnsFetchTimeout)
       this.nnsFetchTimeout = setTimeout(() => {
         resolveNnsDomain(value)
-        .then(address => {
-          clearErrors(index, name)
-          this.setState({ isNnsResolving: false })
-          return updateRowField(index, name, address)
-        })
-        .catch(() => this.setState({
-          nnsError: 'NNS domain not found',
-          isNnsResolving: false
-        }))
+          .then(address => {
+            clearErrors(index, name)
+            this.setState({ isNnsResolving: false })
+            return updateRowField(index, name, address)
+          })
+          .catch(() =>
+            this.setState({
+              nnsError: 'NNS domain not found',
+              isNnsResolving: false
+            })
+          )
       }, 500)
     }
 
+    if (value > max) value = max
     clearErrors(index, name)
     return updateRowField(index, name, value)
   }
@@ -95,7 +98,7 @@ class SendRecipientListItem extends Component<Props, State> {
     const { name } = e.target
     const { clearErrors, index } = this.props
     clearErrors(index, name)
-    name === 'address' && this.setState({ nnsError: '' })
+    if (name === 'address') this.setState({ nnsError: '' })
   }
 
   createAssetList = (): Array<string> => Object.keys(this.props.sendableAssets)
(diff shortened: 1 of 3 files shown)


app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
```diff
@@ -48,6 +59,26 @@ class SendRecipientListItem extends Component<Props> {
 
     const { name } = e.target
     let { value } = e.target
+    if (name === 'address') this.setState({ nnsError: '' })
+    if (name === 'address' && value.endsWith('.neo')) {
+      this.setState({ isNnsResolving: true })
```
dvdschwrtz: does any of the following code need to happen after setting state? If so, setState takes a callback that we can use. If not, its fine the way it is

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
```diff
@@ -48,6 +59,26 @@ class SendRecipientListItem extends Component<Props> {
...
+      if (this.debounceTimer) clearTimeout(this.debounceTimer)
+      this.debounceTimer = setTimeout(() => {
+        resolveNnsDomain(value)
+          .then(address => {
+            clearErrors(index, name)
+            this.setState({ isNnsResolving: false })
```
dvdschwrtz: same comment here with setState callback. would look like this this.setState({ isNnsResolving: false }, () => doSomething())

app/util/nns.js
```diff
@@ -0,0 +1,9 @@
+import { rpc } from 'neon-js'
+
+export const resolveNnsDomain = name =>
+  rpc
+    .queryRPC('https://apiwallet.nel.group/api/mainnet', {
+      method: 'getresolvedaddress',
+      params: [name]
+    })
+    .then(data => data.result[0].data)
```
dvdschwrtz: use lodash get here in case the data has a different shape or else add a catch clause to capture any errors

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
```diff
@@ -48,6 +59,26 @@ class SendRecipientListItem extends Component<Props> {
 
     const { name } = e.target
     let { value } = e.target
+    if (name === 'address') this.setState({ nnsError: '' })
```
dvdschwrtz: I think its easier to read if we use brackets around all if statements, but I don't want to enforce my personal preferences. Perhaps we should come up with some sort of code style guide for neon-wallet that everyone agrees upon. @comountainclimber what do you think?

comountainclimber: We actually enforce very specific linting and formatting rules **already** via eslint and prettier... While I agree if else blocks without the `{` are slightly weird - they are perfectly valid within the context of our style guide (although I would be open to excluding them)

comountainclimber:
Fantastic job @nickfujita! Will merge after these minor comments are addressed

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
nickfujita: Nope, this is just to set the loading animation.

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
nickfujita: Just to unset the loading animation.

app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
nickfujita: It is also my personal preference to always have brackets around an if statement. However, I think in the context of this open source repo, it's more important to keep consistent with the code styles already present in the repo.

I originally had the following as my personal preference, but the linter does not allow it:
`name === 'address' && this.setState({ nnsError: '' })`

app/util/nns.js
nickfujita: While looking into adding this, it turns out that logic becomes has the same effect whether or not I parse using lodash get.

```
const address = get(data, 'result[0].data')
if (!address) throw 'malformatted response'
return address
```

In the case that the data is not in the shape as expected, an error will occur, and the promise will look for a catch statement. In this case, we want this to happen so that the user of this method can provide their own catch statement in the case of an error. As seen here:
https://github.com/CityOfZion/neon-wallet/blob/eb4dbb4ec5eaed56fb509439965ed9b68901eb83/app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx#L73

An alternative would be to return a `null` value in the case where there was an error parsing the response, but then this requires method users to have 2 separate places where they need to catch errors, in the `then` clause where the response argument equals `null`, and in a `catch` statement for when a name is not found. In either case, from the UI perspective, the result is that the user will be provided with the error that the name was not found.

For now, will leave this as-is, unless there may be a case I am missing.

nickfujita:
@comountainclimber Feedback addresses as comments, no code changes made. Waiting for replies from @dvdschwrtz, else we are good to go.

app/util/nns.js
comountainclimber: Makes sense to me

PR #1515, opened 2018-10-16
feat(AmountsPanel): Retweak/redesign amounts panel to match requirements

evgenyboxer:
N/A

Retweak and redesign the amounts panel header according to the new logic:
1. Show top 10 assets sorted by asset worth (price * balance)
2. When <= 5 assets are shown - render only a single row.
3. When > 5 assets are shown - render them in 2 rows and fill the rest if < 10 assets.
4. Handles the case where there are zero assets to show.

The design requirements of this are non trivial (different rendering for single/multi lines with different alignment requirements). Special handling is added for single row/multi row rendering. 

Tested it with mock data (no assets, 1-5 assets and 6-10 to see it in multi rows)

I put a width of 1150px so that boxes would look great with the new wallet dimensions. It does not support a responsive layout yet.

Nope.

- [ ] Unit tests written?
screen shot 2018-10-17 at 4 12 25 pm
screen shot 2018-10-17 at 4 17 31 pm
screen shot 2018-10-16 at 5 05 35 pm

### DIFF
--- a/app/components/AmountsPanel/AmountsPanel.scss
+++ b/app/components/AmountsPanel/AmountsPanel.scss
@@ -49,7 +49,7 @@ h1 {
 .zeroAmount {
   min-height: 70px;
   max-height: 70px;
-  overflow-y: auto;
+  overflow: hidden;
 }
 
 .amountsPanelSingleRow {
--- a/app/components/AmountsPanel/index.jsx
+++ b/app/components/AmountsPanel/index.jsx
@@ -26,6 +26,7 @@ type Props = {
 const ORDER_BY_FIELD = 'totalBalanceWorth'
 const ORDERY_BY_DIRECTION = 'desc'
 const RESULTS_PER_ROW = 5
+const MAX_RESULTS = 10
 
 const validateAmountDataItem = (amountDataItem: AmountDataItem) =>
   has(amountDataItem, 'symbol') &&
@@ -47,7 +48,7 @@ const AmountsPanel = ({ amountsData, currencyCode }: Props) => {
     amountsData,
     item => (item.price ? item[ORDER_BY_FIELD] : '0'),
     ORDERY_BY_DIRECTION
-  )
+  ).slice(0, MAX_RESULTS)
 
   const amountsInSingleRow = orderedAmounts.length <= RESULTS_PER_ROW
   const amountToFill = amountsInSingleRow


app/components/AmountsPanel/index.jsx
```diff
@@ -1,32 +1,89 @@
...
-        remainingAmount={dataset.currentBalance}
-        totalBalanceWorth={dataset.totalBalanceWorth}
-        remainingBalanceWorth={dataset.remainingBalanceWorth}
-        fiatCurrencySymbol={CURRENCIES[currencyCode].symbol}
-      />
-    ))}
-  </section>
-)
+const ORDER_BY_FIELD = 'totalBalanceWorth'
+const ORDERY_BY_DIRECTION = 'asc'
```
comountainclimber: dont we want these in descending order of total balance worth not ascending? User probably expects to see their most valuable asset first

comountainclimber:
Nice work so far @evgenyboxer! Code looks good to me however something weird is going on when I run this branch locally with the layout of this component... It seems to be getting cut off breaking the layout. The other big thing I noticed is that we need we need to incorporate padding into `.amountsInfoBox` as the design mocks show as well as have three different font color and weight variances for the information being displayed
screen shot 2018-10-16 at 8 57 56 am

app/components/AmountsPanel/AmountsPanel.scss
```diff
@@ -1,17 +1,62 @@
...
+  line-height: 32px;
+  margin-bottom: 0;
+}
+
+.container,
+.zeroAmount {
+  margin-bottom: 20px;
+}
+
+.container {
```
comountainclimber: Should have a background color of `#f2f2f2` so that when expanding the window the layout doesnt look "off"

comountainclimber:
Here is a screen shot of the amounts getting "cut off"
screen shot 2018-10-16 at 9 59 03 am
 for reference

app/components/AmountsPanel/index.jsx
```diff
@@ -1,32 +1,89 @@
...
+const MAX_RESULTS = 10
+const RESULTS_PER_ROW = 5
+
+const AmountsPanel = ({ amountsData, currencyCode }: Props) => {
+  if (amountsData.length === 0) {
```
dvdschwrtz: what if it is not an array?

Issue #1520, opened 2018-10-16
Misplaced elements on "Add token" page


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (a64a531)

**Environment:**
MacBook Pro (13-inch, 2017), macOS Mojave 10.14

**Description:**
Looks like some elements are misplaced on "Add token" page:
1. "Save" button looks like it's stretched and not moved to the bottom of the modal
2. After hovering on "Add token" button (top left) it seems that gray highlight merged with top "Neon" bar which looks quite weird
3. After adding a new token, script hash input field and "trash" icon is moved to the left, should be centered inside modal instead

**Screenshots:**
<img width="591" alt="screenshot 2018-10-16 at 20 40 14" src="https://user-images.githubusercontent.com/8177587/47039468-4ea31900-d184-11e8-8d25-d2e1f7f25b2e.png">
<img width="643" alt="screenshot 2018-10-16 at 20 41 49" src="https://user-images.githubusercontent.com/8177587/47039469-4ea31900-d184-11e8-9236-53858e89e81e.png">

drptbl:
Fixed.

PR #1529, opened 2018-10-17
fixes padding issue in token modal

comountainclimber:
fixes https://github.com/CityOfZion/neon-wallet/issues/1520

### DIFF
--- a/app/containers/Send/Send.jsx
+++ b/app/containers/Send/Send.jsx
@@ -139,7 +139,10 @@ export default class Send extends React.Component<Props, State> {
         )
         const price = prices[asset]
 
-        const totalBalanceWorth = multiplyNumber(balance, price)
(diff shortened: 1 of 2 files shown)


app/containers/Send/Send.jsx
```diff
@@ -139,7 +139,10 @@ export default class Send extends React.Component<Props, State> {
         )
         const price = prices[asset]
 
-        const totalBalanceWorth = multiplyNumber(balance, price)
+        const totalBalanceWorth = multiplyNumber(
+          Number(balance).toFixed(15),
```
comountainclimber: mul throws an error if the string or number is greater than 15 significant digits - this fixes that bug

```
export const multiplyNumber = (x: string | number, y: string | number) =>
  toBigNumber(x)
    .mul(y)
    .toNumber()
```

PR #1532, opened 2018-10-17
Update NNS domain resolver to direct contract call

nickfujita:
This PP adds back the UI on the send screen for domain name resolution, and updates the calls for resolution to be direct contract calls rather than using the NEL API.

The same logic is in the process of being added to `neon-js` as a plugin, but the `design-v2` branch of `neon-wallet` is only compatible up to version v3.11.0. So upgrading to the next iteration to include the [soon to be](https://github.com/CityOfZion/neon-js/pull/328) `neon-nns` plugin will require more potential updates to `neon-wallet` than are in the scope of just the name service integration.

At such a time where `neon-wallet` us updated to the latest version of `neon-js` with the future plugin, the contract call logic can simply be replaced with the plugin.

### DIFF
--- a/app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
+++ b/app/components/Send/SendPanel/SendRecipientList/SendRecipientListItem/index.jsx
@@ -1,10 +1,12 @@
 // @flow
 import React, { Component } from 'react'
+import { resolveNnsDomain } from '../../../../../util/nns'
 
 import SelectInput from '../../../../Inputs/SelectInput'
 import NumberInput from '../../../../Inputs/NumberInput'
 import TextInput from '../../../../Inputs/TextInput'
 import DisplayInput from '../../../DisplayInput'
+import Loader from '../../../../Loader'
 
 import TrashCanIcon from '../../../../../assets/icons/delete.svg'
 
@@ -26,7 +28,16 @@ type Props = {
   updateRowField: (index: number, field: string, value: any) => any
 }
 
-class SendRecipientListItem extends Component<Props> {
+type State = {
+  nnsError: string,
+  isNnsResolving: boolean
+}
+
+class SendRecipientListItem extends Component<Props, State> {
+  state = { nnsError: '', isNnsResolving: false }
+
+  debounceTimer = undefined
+
   handleFieldChange = (e: Object) => {
     const {
       index,
@@ -48,6 +59,26 @@ class SendRecipientListItem extends Component<Props> {
 
     const { name } = e.target
     let { value } = e.target
+    if (name === 'address') this.setState({ nnsError: '' })
+    if (name === 'address' && value.endsWith('.neo')) {
+      this.setState({ isNnsResolving: true })
+      if (this.debounceTimer) clearTimeout(this.debounceTimer)
+      this.debounceTimer = setTimeout(() => {
+        resolveNnsDomain(value)
+          .then(address => {
+            clearErrors(index, name)
+            this.setState({ isNnsResolving: false })
+            return updateRowField(index, name, address)
+          })
+          .catch(() =>
+            this.setState({
+              nnsError: 'NNS domain not found',
+              isNnsResolving: false
+            })
+          )
+      }, 500)
+    }
+
     if (value > max) value = max
     clearErrors(index, name)
     return updateRowField(index, name, value)
@@ -67,6 +98,7 @@ class SendRecipientListItem extends Component<Props> {
     const { name } = e.target
     const { clearErrors, index } = this.props
     clearErrors(index, name)
+    if (name === 'address') this.setState({ nnsError: '' })
   }
 
   createAssetList = (): Array<string> => Object.keys(this.props.sendableAssets)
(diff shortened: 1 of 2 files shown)


canesin:
@nickfujita there is currently no plans to support NNS.

canesin:
@jeroenptrs that was reverted, is not because it is a call to a contract that it is decentralized. That word is being used too lightly in the ecosystem. I will let my view (that I know is shared by many but  on this here:

NNS has several problems, it is a monopolized and centralized solution for name services. We already have that, is called ICANN. There is no need to have a ICO for a name service, and funds collected should be a public service redistributed to NEO holders. Not to holders of the ICO token. The name contract should not be restricted in features to allow higher profitability and multi-level sub domains should be allowed, so one can have "sectorA.mycompany.neon.neo or sectorA.mycompany.o3.neo". Finally it should be a standard of the network and the contract should not have a single entity as owner, NNS could take inspiration from where it gets the name from:

"The goal is to replace this interim registrar with a permanent one, which has a target date of two years after launch of this initial registrar. The intervening period will give the community an opportunity to observe the interim registrar in operation, and to submit proposals for a permanent registrar via the EIP (Ethereum Improvement Proposal) standardisation process. The accepted proposal will be implemented and deployed to replace the interim registrar. We expect the proposal to provide a fair process for transferring names registered with the initial registrar to the permanent registrar."

This is is on the homepage of: https://ens.domains/

Finally the auctions that happened in NNS already show that there was no protection for domain squatting (see google.neo), something that is illegal in most countries across the globe and WIPO itself has a complain mechanism that can be used to retrieve domains.

Are we truly striving to a open smart economy that empowers individual freedom or are we replacing monopolies of registrars and stamps with new ones ?
Best

jeroenptrs:
@canesin thanks for the input! I just wanted to point out the existing (apparently reverted) merged PR but it's very interesting to read about your view on NNS and I definitely agree with everything you discuss in it.

merl111:
@canesin I agree on several points with you and I also think a standard should be the final goal.

A proposal for a Neo Name Service standard exists already, but it has not been approved:
 https://github.com/neo-project/proposals/pull/65

NNS has never had an ICO as there was only a private sale for the initial funding. The other part of the tokens have been airdropped to NEO holders and I know a significant part of the tokens are still held by the team. I really hope they find a fair solution on how to distribute them. 

>The name contract should not be restricted in features to allow higher profitability 

What features do you think of?

>and multi-level sub domains should be allowed, so one can have "sectorA.mycompany.neon.neo or sectorA.mycompany.o3.neo". 

Multi level subdomains are planned and also mentioned in the whitepaper: http://doc.neons.name/en/latest/nns_system.html?highlight=subdomain#domain-name-storage

>Finally the auctions that happened in NNS already show that there was no protection for domain squatting (see google.neo)

I think this is a very difficult topic. Having something that's decentralized on one side, but then restrict (or censor, because it's a censorship at some point) the usage of it on the other side doesn't work out. Anyway, if there would be a standard for name services (or even a name service within the NEO chain) I think some will complain mechanism could be implemented and might be necessary, even if it would weaken the decentralized aspects of a blockchain and I thin decentralized censorship is not possible. 

Finally, I don't think NNS is perfect and has several aspects that can be improved, but it's a start. Maybe, if one could get the right people on the same table they can find a solution that would fit.

There is just one thing I am sure of, having multiple name services working on the same chain without any communication is going to hurt NEO more than having no name service.

Best regards

canesin:
@merl111 

> NNS has never had an ICO as there was only a private sale for the initial funding. The other part of the tokens have been airdropped to NEO holders and I know a significant part of the tokens are still held by the team. I really hope they find a fair solution on how to distribute them.

Initial Coin Offering doesn't imply public, specially because it is a Chinese project and public offerings are banned in China. It was 99% private with a 1% airdrop. Why a name service needs a token ? Sorry but for me a domain system should be a "public utility" of the network.

> What features do you think of?

When you acquire a DNS domain you can set multi-level subdomains for free, this should be implemented before any auction. We have people looking for "name.neo" when that should be almost zero value and created by their wallet when they create a account on it.

> Multi level subdomains are planned and also mentioned in the whitepaper

When writing anything is possible. I do think NNS team is competent and will deliver, but why start mainnet operations and hold auctions when something with so big impact on domains is not available ?

> I think this is a very difficult topic. Having something that's decentralized on one side, but then restrict (or censor, because it's a censorship at some point) the usage of it on the other side doesn't work out.

It doesn't need to be sophisticated. One can simply not auction Alexa top 5k until the system is fully operational. Anything like that is clear domain squatting. I think the worse is not learning from ENS after several discussions on the top.

> Finally, I don't think NNS is perfect and has several aspects that can be improved, but it's a start. Maybe, if one could get the right people on the same table they can find a solution that would fit.

I do hope so, NNS could be great.

> There is just one thing I am sure of, having multiple name services working on the same chain without any communication is going to hurt NEO more than having no name service.

Somewhat agree, there was no conversation about NNS with CoZ, why should we support it ? It is a for profit individual project. CoZ has a very important rule that it should not be a vehicle for promotion of individual projects. That is why even with several members having their projects you don't see they connecting it to CoZ projects. If we open that door I think NEO will be severely damaged.

QingmingZi:
@canesin Even in the DNS, Google has a com, which does not necessarily mean that it must have io or other very small top-level domain. Protection, itself is also a very central thing, who has the right to decide that google.neo must belong to Google?

PR #1542, opened 2018-10-18
Feature/security audit

comountainclimber:
 - Removes fetching of token info from github
 - Ports over the work that @nickfujita did in the neo-qrcode repo so that we can be completely independent from any third part APIs or servcies
- Adds svgs for all nep5 tokens that I could find
- Adds an HOC called withCopyCanvasToClipboard.js for easy reuse in the future
- Fixes the layout in scan
- Fixes the layout in nep9 QR generation

NOTE: the logo placement of the nep5 tokens on the nep9 QRs is not 100% perfect... spent a really long time messing with the positioning and this was the best I could come up with

- [ ] Unit tests written?


app/assets/nep5/raw-svg-export/index.js
```diff
@@ -0,0 +1,97 @@
+import NEO from '../NEO.svg'
```
comountainclimber: If you look in my changes to the webpack configurations this file actually bypasses the `react-svg-loader` by design for use in Nep9QrGenerator.js

app/components/Modals/ReceiveModal/ReceiveModal.jsx
```diff
@@ -18,22 +24,33 @@ type Props = {
...
     loading: true
   }
 
+  state = {
+    loading: true,
+    copied: false
+  }
+
+  handleCopyIcon = () => {
+    this.setState({ copied: true })
```
dvdschwrtz: this is ok, but I would put it in the callback after copied: true is complete just in case it takes some time to run

app/hocs/withCopyCanvasToClipboard.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    handleCreateCanvas = (ref: ?HTMLCanvasElement, data: string) => {
+      QRCode.toCanvas(ref, data, { version: 5 }, err => {
+        if (err) console.error(err)
+      })
+    }
+
+    handleCopyIcon = () => {
+      this.setState({ copied: true })
```
dvdschwrtz: same comment as above

app/modules/nep9/Nep9QrGenerator.js
```diff
@@ -0,0 +1,127 @@
...
+/* eslint-disable */
+import QRCode from 'qrcode/lib/browser'
+import nep9 from './nep9'
+import tokenList from '../../core/tokenList.json'
+import { imageMap } from '../../assets/nep5/raw-svg-export'
+
+const TOKENS = Object.keys(tokenList)
+  .map(key => tokenList[key])
+  .reduce((accum, token) => {
+    accum[token.networks[1].hash] = token
```
dvdschwrtz: this is prob ok if the list is hardcoded but I still prefer lodash get

app/modules/nep9/Nep9QrGenerator.js
```diff
@@ -0,0 +1,127 @@
...
+      .then(([nep5Data, c]) => {
+        canvas = c
+        const token = {}
+        if (nep9Data.asset === 'NEO') token.symbol = 'NEO'
+        if (nep9Data.asset === 'GAS') token.symbol = 'GAS'
+        else {
+          const foundToken =
+            Object.keys(tokenList)
+              .map(key => tokenList[key])
+              .find(token => token.networks[1].hash === nep9Data.asset) || {}
```
dvdschwrtz: same comment as above

app/modules/nep9/Nep9QrGenerator.js
```diff
@@ -0,0 +1,127 @@
...
+          token.symbol = foundToken.symbol
+        }
+        const logo = imageMap[token.symbol] || imageMap.NEO
```
dvdschwrtz: same comment as above

app/components/Modals/ReceiveModal/ReceiveModal.jsx
comountainclimber: 👍

app/hocs/withCopyCanvasToClipboard.js
comountainclimber: 👍

app/modules/nep9/Nep9QrGenerator.js
comountainclimber: I think we are ok here without `get` because this token will always default to an empty object meaning that the lookup against `imageMap`  would be `imageMap[undefined]` which will not throw an exception

app/modules/nep9/Nep9QrGenerator.js
comountainclimber: Because its hardcoded and cannot change I am going to leave this as is for now

app/modules/nep9/Nep9QrGenerator.js
comountainclimber: Because its hardcoded and cannot change I am going to leave this as is for now

comountainclimber:
@dvdschwrtz tried implementing setting state on those clipboard functions based on when the thing is actually finished being copied and its simply too fast for the animation - looks like a flicker and super buggy so going to leave those items as is for now

Issue #1541, opened 2018-10-17
batched calls to getTokenBalances()


comountainclimber:

drptbl:
Implemented.

PR #1543, opened 2018-10-18
updates balancesActions.js to work with 2.9.0 nodes

comountainclimber:
Because nodes running 2.9.0 have been capped at 10 GAS limit for invoke scripts we can no longer request balance info in one RPC call... This PR chunks out the request so that balance data for no more than 4 tokens is ever made at once.

Fixes https://github.com/CityOfZion/neon-wallet/issues/1541

Related update: https://github.com/neo-project/neo/pull/397

### DIFF
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -1,6 +1,6 @@
 // @flow
 import { api } from 'neon-js'
-import { extend, isEmpty } from 'lodash-es'
+import { extend, isEmpty, unionBy } from 'lodash-es'
 import { createActions } from 'spunky'
 
 import { getNode } from './nodeStorageActions'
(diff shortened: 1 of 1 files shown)


app/actions/balancesActions.js
```diff
@@ -21,23 +21,42 @@ async function getBalances({ net, address, tokens }: Props) {
...
-    address
+  const chunks = tokens
+    .filter(token => !token.isUserGenerated)
+    .reduce((accum, currVal, index) => {
+      if (!accum.length) {
+        accum.push([currVal.scriptHash])
+        return accum
+      }
+
+      if (accum[accum.length - 1].length < 4) {
```
dvdschwrtz: why 4? a good variable name would help

app/actions/balancesActions.js
```diff
@@ -21,23 +21,42 @@ async function getBalances({ net, address, tokens }: Props) {
...
+        accum[accum.length - 1].push(currVal.scriptHash)
```
dvdschwrtz: this push does not have an array surrounding currVal.scripthash but the rest do

app/actions/balancesActions.js
```diff
@@ -21,23 +21,42 @@ async function getBalances({ net, address, tokens }: Props) {
...
+      }
+      return accum
+    }, [])
+
+  const promiseMap = chunks.map(chunk =>
+    api.nep5.getTokenBalances(endpoint, chunk, address)
   )
+  const results = await Promise.all(promiseMap)
+  const resultObject = results.reduce((result, currentObject) => {
+    Object.keys(currentObject).forEach(key => {
```
dvdschwrtz: maybe just use return Object.assign({}, currentObject)

app/actions/balancesActions.js
```diff
@@ -21,23 +21,42 @@ async function getBalances({ net, address, tokens }: Props) {
...
+      // eslint-disable-next-line
+      result[key] = currentObject[key]
+    })
+    return result
+  }, {})
 
-  const parsedTokenBalances = Object.keys(tokenBalances)
+  const parsedTokenBalances = Object.keys(resultObject)
```
dvdschwrtz: you could also return the format you need from the reduce method instead of calling object.keys on the data after calling reduce. its not a big deal though unless the data is large and we want to be efficient

app/actions/balancesActions.js
comountainclimber: Thanks for taking a look guys... so 4 represents the "maximum chunk size" I will go ahead and make this a constant

app/actions/balancesActions.js
comountainclimber: Thats because its pushing into an existing arr... Think of the structure like this:
`[ [hash1, hash2, hash3, hash4], [hash5, hash6, hash7, hash8] ... ]`

app/actions/balancesActions.js
comountainclimber: 👍

Issue #1538, opened 2018-10-17
Defect: wallet version not displayed on application


lllwvlvwlll:
**Wallet version or commit SHA:**
0.0.7

**Environment:**
win 10

**Prerequisites:**
none

**Reproduction steps:**

The wallet version is not displayed in the application.

drptbl:
Implemented.

PR #1548, opened 2018-10-19
fixes layout bug and adds version

comountainclimber:
fixes https://github.com/CityOfZion/neon-wallet/issues/1538

and a layout bug causing a vertical overflow at the default resolution...
screen shot 2018-10-18 at 9 58 26 pm
screen shot 2018-10-18 at 9 59 23 pm

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Settings/Settings.jsx
+++ b/app/containers/Settings/Settings.jsx
@@ -32,6 +32,7 @@ import CogIcon from '../../assets/icons/cog-icon.svg'
 import NodeSelectIcon from '../../assets/icons/node-select.svg'
 import TimeIcon from '../../assets/icons/time-icon.svg'
 import SaveIcon from '../../assets/icons/save-icon.svg'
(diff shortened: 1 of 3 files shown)


app/containers/Settings/Settings.jsx
```diff
@@ -32,6 +32,7 @@ import CogIcon from '../../assets/icons/cog-icon.svg'
 import NodeSelectIcon from '../../assets/icons/node-select.svg'
 import TimeIcon from '../../assets/icons/time-icon.svg'
 import SaveIcon from '../../assets/icons/save-icon.svg'
+import pack from '../../../package.json'
```
evgenyboxer: very minor, but it would be better to call this import `packageJSON` or `packageJson`.

comountainclimber: 👍 updated to just use a destructured import thanks for calling this out

PR #1568, opened 2018-10-23
Feature/wif qr login

comountainclimber:
- Adds QR canvas to "creation/import success" panel
- Adds the ability to login via private key QR
- Fixes some CSS var assignments
screen shot 2018-10-22 at 4 40 55 pm
screen shot 2018-10-22 at 7 57 31 pm
screen shot 2018-10-22 at 7 57 27 pm


app/containers/LoginPrivateKey/LoginPrivateKey.jsx
```diff
@@ -31,27 +47,98 @@ export default class LoginPrivateKey extends React.Component<Props, State> {
...
+            <div className={styles.centeredInput}>
+              <PasswordInput
+                placeholder="Enter your private key here"
+                value={wif}
+                onChange={(e: Object) => this.setState({ wif: e.target.value })}
+                autoFocus
+              />
+            </div>
+          )}
+          {scannerActive ? (
```
dvdschwrtz: You can combine this and the ternary above.

comountainclimber: 👍 good catch @dvdschwrtz

Issue #1606, opened 2018-10-25
MainNet nodes are shown on TestNet node selection page


drptbl:
It seems that nodes from mainnet are also shown after switching to testnet and going to node selection page.

Reason:
https://raw.githubusercontent.com/CityOfZion/neon-wallet/design-v2/app/core/nodes-test-net.json

Most of the don't work on TestNet (while fetching balances, etc).

drptbl:
Fixed.

PR #1607, opened 2018-10-26
feature/fixes-node-select-in-testnet

comountainclimber:
fixes https://github.com/CityOfZion/neon-wallet/issues/1606

- [ ] Unit tests written?

### DIFF
--- a/app/components/NodeSelectPanel/index.js
+++ b/app/components/NodeSelectPanel/index.js
@@ -7,7 +7,8 @@ import {
   alreadyLoadedStrategy,
   progressValues,
   withProgressComponents,
-  withReset
+  withReset,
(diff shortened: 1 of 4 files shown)


app/components/NodeSelectPanel/index.js
```diff
@@ -7,7 +7,8 @@ import {
   alreadyLoadedStrategy,
   progressValues,
   withProgressComponents,
-  withReset
+  withReset,
+  withRecall
```
dvdschwrtz: this was added but is not being used

dvdschwrtz: aha, I saw it in the next PR!

Issue #1522, opened 2018-10-16
Wallet crashes while going to "Send" page too fast after logging in to the wallet


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (a64a531)

**Environment:**
MacBook Pro (13-inch, 2017), macOS Mojave 10.14

**Pre-requirements:**
1. Have any amount of any token on wallet balance

**Reproduction steps:**
1. Login to your wallet
2. Click on "Send" button in the left menu after login (depending on network speed, I think that you have around 5 seconds to do this in normal conditions, more time when speed is slower)

**Description:**
Wallet crashes after going to "Send" page too fast after logging in to the wallet.

**Screenshots/Video:**
![2018-10-16 21 06 19](https://user-images.githubusercontent.com/8177587/47041183-6ed4d700-d188-11e8-9cd3-60efc3fc38b6.gif)

drptbl:
Still causes a crash, but not instantly like before, but after clicking on "recipient" input field:
https://i.imgur.com/Fgn3jtr.gifv

drptbl:
Fixed. 👍

PR #1609, opened 2018-10-26
feature/fixes-bug-in-send

comountainclimber:
fixes https://github.com/CityOfZion/neon-wallet/issues/1522

### DIFF
--- a/app/containers/Send/Send.jsx
+++ b/app/containers/Send/Send.jsx
@@ -347,7 +347,7 @@ export default class Send extends React.Component<Props, State> {
       const newState = [...prevState.sendRowDetails]
 
       const objectToClear = newState[index]
-      if (objectToClear.errors[field]) {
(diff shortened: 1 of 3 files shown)


app/containers/Send/Send.jsx
```diff
@@ -347,7 +347,7 @@ export default class Send extends React.Component<Props, State> {
       const newState = [...prevState.sendRowDetails]
 
       const objectToClear = newState[index]
-      if (objectToClear.errors[field]) {
+      if (objectToClear.errors && objectToClear.errors[field]) {
```
dvdschwrtz: just use lodash get for this. you can pass an array I believe for a case like this where you don't know some of the keys and just use variables

comountainclimber: Thanks @dvdschwrtz 💪

PR #1610, opened 2018-10-28
feat(scrollbar): design a better looking scrollbar

evgenyboxer:
N/A

Designs a nicer scrollbar which looks better (especially in the dark theme)

Added the special scrollbar rules in the main scss file and added variables for them in the dark/light theme files.

Manually tested it.

Since Electron is based on Chromium the special CSS rule will work on it.

Looking for a design feedback from Ben regarding the scroll width and colors for each theme.

Before: (Dark)
screen shot 2018-10-28 at 10 17 18 pm

After: (Dark)
screen shot 2018-10-28 at 10 10 42 pm

Before (Light)
screen shot 2018-10-28 at 10 16 23 pm

After (Light)
screen shot 2018-10-28 at 10 14 31 pm


app/styles/main.global.scss
```diff
@@ -61,6 +61,19 @@ body {
   }
 }
 
+::-webkit-scrollbar {
```
comountainclimber: 🔥

PR #1613, opened 2018-10-28
fix(transactions): disables mint tokens address from being added as a contact + cleanups

evgenyboxer:
N/A

1. Disables mint tokens address from being added as a contact. (the address "MINT TOKENS" is invalid anyway).
2. The "Add" was inconsistent for an existing contact for the send/receive sections... Basically, the receive section would remove the "Add" button while the send would just disable it. I disabled it in both cases.
3. Code cleanups.

Manually tested it

No

No

Before:
screen shot 2018-10-28 at 10 35 03 pm

After: 
screen shot 2018-10-29 at 1 23 26 am

### DIFF
--- a/app/components/Blockchain/Transaction/Transaction.jsx
+++ b/app/components/Blockchain/Transaction/Transaction.jsx
@@ -79,6 +79,12 @@ export default class Transaction extends React.Component<Props> {
   renderAbstract = (type: string) => {
     const { tx } = this.props
     const { iconType, time, label, amount, isNetworkFee, to, from } = tx
+
+    const formattedTime = moment.unix(time).format('MM/DD/YYYY | HH:MM:ss')
+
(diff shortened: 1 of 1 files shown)


app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -79,6 +79,12 @@ export default class Transaction extends React.Component<Props> {
   renderAbstract = (type: string) => {
     const { tx } = this.props
     const { iconType, time, label, amount, isNetworkFee, to, from } = tx
+
+    const formattedTime = moment.unix(time).format('MM/DD/YYYY | HH:MM:ss')
+
+    const contactTo = this.findContact(to)
+    const contactToExists = contactTo !== to
```
comountainclimber: Nice, thank you 🙏

PR #1618, opened 2018-10-30
Feature/release defects

comountainclimber:
### Fixes numerous defects discovered during regression testing of release candidate

1. Removes potential for scientific notation for display of balance data
2. Fixes encrypt key functionality
3. Error icon will no longer overlap password reveal on error state
4. Token sale symbol drop down will always update when token list is updated
5. Fixes broken CSS in confirm modal and removes logo
6. Increases chunk size in balance actions to 5 to reduce i/o against nodes

### DIFF
--- a/app/hocs/withTokensData.js
+++ b/app/hocs/withTokensData.js
@@ -1,8 +1,13 @@
 // @flow
-import { withData } from 'spunky'
+import { withData, withRecall } from 'spunky'
+import { compose } from 'recompose'
 
 import settingsActions from '../actions/settingsActions'
+import icoTokensActions from '../actions/icoTokensActions'
 
 export default function withTokensData(key: string = 'tokens') {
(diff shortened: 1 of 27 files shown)


app/hocs/withTokensData.js
```diff
@@ -1,8 +1,13 @@
...
+import { compose } from 'recompose'
 
 import settingsActions from '../actions/settingsActions'
+import icoTokensActions from '../actions/icoTokensActions'
 
 export default function withTokensData(key: string = 'tokens') {
-  return withData(settingsActions, settings => ({ [key]: settings.tokens }))
+  return compose(
+    withData(settingsActions, settings => ({ [key]: settings.tokens })),
+    withRecall(icoTokensActions, ['tokens'])
```
comountainclimber: Refetch data for token sale dropdown whenever tokens get updated

PR #1645, opened 2018-11-06
feature: various defect fixes

comountainclimber:
- selected node now applies to token sale and send tx
- gas fee applied in token sale
- Fixes styling in dark theme of priority fee button
- adds the ability to toggle priority fee button
- Corrects verbiage in send confirmation updating TRANSACTIONS => TRANSFERS
- Fixes link in confirmation success to use selected explorer
- Filters custom added hashes by network in balancesActions.js (reducing unnecessary requests and eliminating error being thrown in console)
- Fixes navigation in authenticated version of confirmation qr codes close button

### DIFF
--- a/app/components/Send/PriorityFee/PriorityFee.jsx
+++ b/app/components/Send/PriorityFee/PriorityFee.jsx
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react'
+import classNames from 'classnames'
 
 import LightningIcon from '../../../assets/icons/lightning.svg'
 import Button from '../../Button/Button'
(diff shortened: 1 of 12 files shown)


app/components/Send/PriorityFee/PriorityFee.jsx
```diff
@@ -43,16 +44,22 @@ export default class PriorityFee extends React.Component<Props> {
...
               renderIcon={() => <LightningIcon />}
-              active={fees === option.fee ? true : undefined}
+              active={fees === option.fee}
               disabled={this.shouldDisableFeeButton(
                 disabled,
                 availableGas,
                 option.fee
               )}
-              onClick={() => handleAddPriorityFee(option.fee)}
+              onClick={() =>
```
evgenyboxer: how about `() => handleAddPriorityFee(fees === option.fee ? option.fee : 0)`

comountainclimber: 👍

app/modules/transactions.js
```diff
@@ -139,16 +141,23 @@ export const sendTransaction = ({
...
+      net,
+      tokensBalanceMap,
+      address: fromAddress,
+      publicKey,
+      privateKey: new wallet.Account(wif).privateKey,
+      signingFunction: isHardwareSend ? signingFunction : null,
+      fees
+    }
+
+    if (url) {
```
dvdschwrtz: I would double check neon-js, but I think if url is undefined it won't use it, so I don't see the purpose of this check to add the url, you can just add it to the config above: eg. config = { net, ..., fees, url }

dvdschwrtz: I would double check neon-js, but I think if url is undefined it won't use it, so I don't see the purpose of this check to add the url, you can just add it to the config above: eg. config = { net, ..., fees, url }

PR #1647, opened 2018-11-06
fix(number-input): use raw value instead of formatted

evgenyboxer:
N/A

Currently, the wallet crashes when you try to send a large amount > `999`. This PR solves this problem by only storing the raw value and not the formatted one produced by `Cleave.js`.

I have removed the `customChangeEvent` which was sort of a workaround around the fact that the `onChange` event did not return the `event` object. A better approach for the handler is to use `onChange => (event, value)`. By returning the raw value from the NumberInput it is now possible to remove all the instances of the comma removal.

I have also cleaned `handleFieldChange` method a bit.

Tested a lot manually!

Token sale also uses the number input and I haven't tested it well enough.

Next thing to do is to remove `customChangeEvent` from the `SelectInput`

Before:
screen shot 2018-11-06 at 4 50 04 pm

After:
screen shot 2018-11-06 at 4 45 25 pm

### DIFF
--- a/app/components/Inputs/NumberInput/NumberInput.jsx
+++ b/app/components/Inputs/NumberInput/NumberInput.jsx
@@ -25,7 +25,6 @@ type Props = {
   onBlur?: Function,
   handleMaxClick: Function,
   error?: string,
-  customChangeEvent?: boolean,
   options?: {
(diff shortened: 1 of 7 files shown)


comountainclimber:
Looks good to me thank you for simplifying this logic and taking care of the bugs

Issue #1649, opened 2018-11-06
Give ability to send NEP-5 assets while neoscan is down


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (fff7987)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.376

**Reproduction steps:**
1. Start wallet
2. Set intercepting on `https://api.neoscan.io/api/main_net/v1/get_balance/<wallet_address>` url
3. Login to the wallet
4. Pass all requests excluding the one mentioned above

**Description:**
Currently if `get_balance` request fails (however requests for NEP-5 balances were success), we're clearing the list of assets to send together with "Token Balances" section. In this case we could allow users to at least send their NEP-5 assets.

**Current behavior:**
![offline-issue1](https://user-images.githubusercontent.com/8177587/48096844-a0d1da00-e218-11e8-9097-b6209f1e55b8.gif)

drptbl:
Re-tested, looks good.

PR #1652, opened 2018-11-07
Feature/1649

comountainclimber:
- Fixes https://github.com/CityOfZion/neon-wallet/issues/1649 by manually constructing the Balance class if neoscan is down allowing users to still send, receive and view nep5 tokens independent of neoscan
- Fixes `balancesActions.js` to not fail if request to nodes is successful but req to neoscan fails

### DIFF
--- a/app/modules/transactions.js
+++ b/app/modules/transactions.js
@@ -23,7 +23,7 @@ import {
   validateTransactionsBeforeSending,
   getTokenBalancesMap
 } from '../core/wallet'
-import { toNumber } from '../core/math'
+import { toNumber, toBigNumber } from '../core/math'
 import { getNode } from '../actions/nodeStorageActions'
 
 const extractTokens = (sendEntries: Array<SendEntryType>) =>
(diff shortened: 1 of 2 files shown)


app/modules/transactions.js
```diff
@@ -151,6 +152,23 @@ export const sendTransaction = ({
...
 
+    await api
+      .getBalanceFrom({ net, address: fromAddress }, api.neoscan)
+      .catch(e => {
+        // indicates that neo scan is down and that api.sendAsset and api.doInvoke
+        // will fail unless balances are supplied
+        console.error(e)
+        const balance = new wallet.Balance({ address: fromAddress, net })
+        Object.keys(tokensBalanceMap).forEach(key => {
+          balance.addAsset(tokensBalanceMap[key].name, {
```
evgenyboxer: I think `Object.values` might work better here - 
```
Object.values(tokensBalanceMap).forEach(({ name, balance }) =>
  balance.addAsset(name, { balance, unspent: [] })
)

```

comountainclimber: I agree! 👍

Issue #1523, opened 2018-10-16
"Token Sale" page shows "no funds" message (despite having funds) while moving to it too fast after logging in to the wallet


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (a64a531)

**Environment:**
MacBook Pro (13-inch, 2017), macOS Mojave 10.14

**Pre-requirements:**
1. Have any amount of any token on wallet balance

**Reproduction steps:**
1. Login to your wallet
2. Click on "Token Sale" button in the left menu after login (depending on network speed, I think that you have around 5 seconds to do this in normal conditions, more time when speed is slower)

**Description:**
"Token Sale" page shows "no funds" message (despite having funds) while moving to it too fast after logging in to the wallet. Data is loaded fine after requests are completed - but user is required to move out of that page and then back to it.

**Screenshots/Video:**
![2018-10-16 21 09 39](https://user-images.githubusercontent.com/8177587/47041345-cffcaa80-d188-11e8-9222-b06353f183d7.gif)

drptbl:
Still not fixed. Same reproduction steps..

PR #1654, opened 2018-11-07
fixes TokenSale asset update bug

comountainclimber:
- Fixes https://github.com/CityOfZion/neon-wallet/issues/1523

### DIFF
--- a/app/containers/TokenSale/TokenSale.jsx
+++ b/app/containers/TokenSale/TokenSale.jsx
@@ -1,5 +1,6 @@
 // @flow
 import React, { Component, Fragment } from 'react'
+import { isEmpty } from 'lodash-es'
 
 import {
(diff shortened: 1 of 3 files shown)


app/containers/TokenSale/TokenSale.jsx
```diff
@@ -90,7 +90,20 @@ class TokenSale extends Component<Props, State> {
     }
   }
 
+  componentDidUpdate(prevProps: Props) {
+    if (
+      prevProps.assetBalances !== this.props.assetBalances &&
```
evgenyboxer: I think it's better to use `lodash` isEqual for comparison.

comountainclimber: 👍

PR #1657, opened 2018-11-08
refactor/remove dead code

FredrikOseberg:

- [ ] Unit tests written?


app/hocs/withSuccessNotification.js
```diff
@@ -29,7 +29,7 @@ export default function withSuccessNotification(
   message: Message,
   options: Object = {}
 ) {
-  const mapDisptchToProps = (dispatch, ownProps) => ({
+  const mapDispatchToProps = dispatch => ({
```
dvdschwrtz: dispatch should have a type

evgenyboxer:
@FredrikOseberg awesome stuff! could this be added as an eslint rule so that future dead code will fail the linter?

FredrikOseberg:
@dvdschwrtz Thank you. I'll take care of those.
@evgenyboxer That is definitely possible, I'll update the PR with the rule specification.

evgenyboxer:
@FredrikOseberg ?

FredrikOseberg:
Apologies @evgenyboxer. I had a chat with @comountainclimber regarding this PR and we decided to put it on hold until v2 was released and include this in a patch at a later point. Since release my timetable has been filled, but I should be able to wrap this up tomorrow.

comountainclimber:
This looks good to me @FredrikOseberg if we can address the missing type definitions that @dvdschwrtz identified we can merge this 👍

FredrikOseberg:
@comountainclimber I took care of those in this commit: https://github.com/CityOfZion/neon-wallet/pull/1657/commits/574c82942033f152f16f5cafbc16d8562f8c7f7a 

I found another instance where it should be added in those files though, so I pushed those changes up.

Issue #1626, opened 2018-11-01
design-v2(UX): [Enhance] Chart scale dropdown menu


jseagrave21:
**Wallet version or commit SHA:**
2.0.0

**Environment:**
Win 10 Pro

**Prerequisites:**
None

**Reproduction steps:**
1. Login to a wallet

- On the home screen it looks like the scale for the chart is a dropdown menu but when I click on it, it cycles through (no dropdown menu). The icon should actually yield a dropdown menu.
- Also the best fidelity I get is 1 Day. We should be able to get better resolution. (1 hour, 4 hour)
- Finally, the default timeframe is 1 month. I think this scale is much too large for the average user. Perhaps 1 Day would be a better default timeframe.

**Screenshots/Video:**
![dropdown](https://user-images.githubusercontent.com/40873301/47866631-8f25a800-ddbc-11e8-8baa-dad2897d48ec.png)

ranbena:
Heyo, I'm working on this and will soon have a PR.
Plz assign me or mark In Progress (I lack the permissions).

comountainclimber:
Fixed 🎉

PR #1659, opened 2018-11-09
[#1626] Implemented chart duration dropdown

ranbena:
Currently, the dashboard chart duration switcher is a click-to-change implementation.
This commit swaps it for a dropdown as requested in #1626.

screen recording 2018-11-09 at 18 03 28

Implemented using `<StyledReactSelect>` with some custom styling to match the design.
Passes all tests, no new tests added (lmk if this feature requires it).

### DIFF
--- a/__tests__/components/__snapshots__/NetworkSwitch.test.js.snap
+++ b/__tests__/components/__snapshots__/NetworkSwitch.test.js.snap
@@ -633,7 +633,7 @@ exports[`NetworkSwitch renders without crashing 1`] = `
                     }
                   >
                     <div
-                      className="css-k5geue react-select__value-container react-select__value-container--has-value"
+                      className="css-6zqqc react-select__value-container react-select__value-container--has-value"
                     >
                       <SingleValue
                         clearValue={[Function]}
@@ -783,7 +783,7 @@ exports[`NetworkSwitch renders without crashing 1`] = `
                         }
                       >
                         <div
-                          className="css-8bm2m9 react-select__single-value"
+                          className="css-ljsv20 react-select__single-value"
                         >
                           MainNet
                         </div>
--- a/app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
+++ b/app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
@@ -1,12 +1,12 @@
 // @flow
 import React from 'react'
 import classNames from 'classnames'
-import { keys } from 'lodash-es'
+import { find } from 'lodash-es'
 
 import PriceHistoryChart from './PriceHistoryChart'
 import Panel from '../../Panel'
 import BoundingBox from './BoundingBox'
-import DropdownIcon from '../../../assets/icons/dropdown.svg'
+import StyledReactSelect from '../../Inputs/StyledReactSelect/StyledReactSelect'
 import {
   ASSETS,
   CURRENCIES,
@@ -16,6 +16,11 @@ import styles from './PriceHistoryPanel.scss'
 import { formatFiat, formatThousands } from '../../../core/formatters'
 
 type Duration = '1m' | '1w' | '1d'
+// react-select option format
+type SelectOption = {
+  value: Duration,
+  label: string
+}
 
 type Price = {
   time: number,
@@ -39,11 +44,17 @@ type Props = {
   priceKey: string
 }
 
-const DURATIONS: { [key: Duration]: string } = {
-  '1m': '1 month',
-  '1w': '1 week',
-  '1d': '1 day'
-}
+const DURATIONS: Array<[Duration, string]> = [
+  ['1d', '1 DAY'],
+  ['1w', '1 WEEK'],
+  ['1m', '1 MONTH']
+]
+
+// convert DURATIONS to react-select option format
+const DURATION_OPTIONS: Array<SelectOption> = DURATIONS.map(([k, v]) => ({
+  value: k,
+  label: v
+}))
 
 export default class PriceHistoryPanel extends React.Component<Props> {
   static defaultProps = {
@@ -81,9 +92,17 @@ export default class PriceHistoryPanel extends React.Component<Props> {
         {this.renderLatestPrice()}
         {this.renderPriceChange()}
       </span>
-      <span className={styles.duration} onClick={this.handleChangeDuration}>
-        {this.getDuration()}
-        <DropdownIcon className={styles.icon} />
+      <span className={styles.duration}>
+        <StyledReactSelect
+          defaultValue={this.getDuration()}
+          onChange={this.handleChangeDuration}
+          options={DURATION_OPTIONS}
+          isSearchable={false}
+          fontWeight="normal"
+          transparent
+          hideHighlight
+          textAlign="right"
+        />
       </span>
     </div>
   )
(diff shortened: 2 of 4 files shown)


__tests__/components/__snapshots__/NetworkSwitch.test.js.snap
```diff
@@ -633,7 +633,7 @@ exports[`NetworkSwitch renders without crashing 1`] = `
                     }
                   >
                     <div
-                      className="css-k5geue react-select__value-container react-select__value-container--has-value"
+                      className="css-6zqqc react-select__value-container react-select__value-container--has-value"
```
ranbena: Not sure why this changed since I touched no css.. ?

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -40,11 +46,17 @@ type Props = {
 }
 
 const DURATIONS: { [key: Duration]: string } = {
-  '1m': '1 month',
+  '1d': '1 day',
   '1w': '1 week',
-  '1d': '1 day'
+  '1m': '1 month'
```
ranbena: Since #1626 requests the default duration be 1 day, I assume this should be the dropdown option order.

ranbena: Sidenote: afaik the order of these entries is not guaranteed. If this is an issue, we can either substitute for an Array (which required some code adjustments I would rather avoid) or perhaps utilize Immutable's `OrderedMap`.

__tests__/components/__snapshots__/NetworkSwitch.test.js.snap
dvdschwrtz: it is prob an autogenerated thing related to css modules

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
dvdschwrtz: The new js Map is ordered as well (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) no need for immutable OrderedMap

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -81,9 +93,16 @@ export default class PriceHistoryPanel extends React.Component<Props> {
...
+      <span className={styles.duration}>
+        <StyledReactSelect
+          defaultValue={this.getDuration()}
+          onChange={this.handleChangeDuration}
+          options={DURATION_OPTIONS}
+          isSearchable={false}
+          fontWeight="normal"
+          transparent
+          hideHighlight
+        />
```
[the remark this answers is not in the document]
dvdschwrtz: just nest it inside of .header in PriceHistoryPanel.scss. 

`.header {
  ...,
  .react-select__single-value {
    right: 0
  }
}`

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
ranbena: Awesome! Ok that it's supported in IE11 and up?

ranbena: Oh right, it's all node based.. takes getting used to after a lifetime of browser compatibility considerations 😅

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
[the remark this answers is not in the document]
ranbena: Doesn't work for some reason. I went another way. Updating PR soon.

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
```diff
@@ -39,11 +44,17 @@ type Props = {
...
-const DURATIONS: { [key: Duration]: string } = {
-  '1m': '1 month',
-  '1w': '1 week',
-  '1d': '1 day'
-}
+const DURATIONS: Array<[Duration, string]> = [
+  ['1d', '1 DAY'],
+  ['1w', '1 WEEK'],
+  ['1m', '1 MONTH']
+]
```
ranbena: Went for Tuple array instead of Object to guarantee entry order.

Didn't go for `new Map()` since it's the same syntax (gets exact array as argument) and iteration is more cumbersome.

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.scss
```diff
@@ -6,8 +6,7 @@
     align-items: center;
     justify-content: space-between;
 
-    .asset,
-    .duration {
+    .asset {
```
ranbena: Before:
screen shot 2018-11-10 at 10 45 46

After:
screen shot 2018-11-10 at 10 45 32

app/components/Inputs/StyledReactSelect/StyledReactSelect.jsx
```diff
@@ -69,6 +70,8 @@ const customStyles = {
     return {
       ...styles,
       fontSize: props.selectProps.fontSize,
+      fontWeight: props.selectProps.fontWeight,
+      [props.selectProps.textAlign]: 0,
```
ranbena: Aligns the single value to the right.
Before:
screen shot 2018-11-10 at 10 33 42

After:
screen shot 2018-11-10 at 10 33 16

ranbena:
@dvdschwrtz all issues solved and ready for review 🤟🏻

evgenyboxer:
@ranbena really excited to see a contribution from you! 🔥Welcome to CoZ!

comountainclimber:
This looks great nice work @ranbena

app/components/Dashboard/PriceHistoryPanel/PriceHistoryPanel.jsx
zsiglin: @ranbena Just curious why a Tuple array was chosen instead of a collection as this also requires the map you've added below.

ranbena: Hey @zsiglin, here's my thought process:

### Option 1 - No mapping
```
DURATION_OPTIONS: Array<SelectOption> = [
  { value: '1d', label: '1 DAY' },
  { value: '1w', label: '1 WEEK' },
  { value: '1m', label: '1 MONTH' }
]
```
### Option 2 - Array
```
const DURATIONS: Array<[Duration, string]> = [
  ['1d', '1 DAY'],
  ['1w', '1 WEEK'],
  ['1m', '1 MONTH']
]

const DURATION_OPTIONS: Array<SelectOption> = DURATIONS.map(([k, v]) => ({
  value: k,
  label: v
}))
```
### Option 3 - Map
```
const DURATIONS: Map<Duration, string> = new Map([
  ['1d', '1 DAY'],
  ['1w', '1 WEEK'],
  ['1m', '1 MONTH']
])

const DURATION_OPTIONS: Array<SelectOption> = Array.from(DURATIONS, ([ k, v ]) => ({
  value: k,
  label: v
}))
```

### Considerations
Option 1 is best for perf, simpler and has less code. Only downside is that it doesn't adhere to "separation of data and implementation" school of thought. Not optimal if in future dropdown component is swapped for another.

Option 2 is not as readable as the original Object `{'1d': '1 DAY', ...}` but does separate data/impl.

Option 3 has almost identical syntax as the same as previous, mildly lengthier. I see no advantage to it.

PR #1663, opened 2018-11-10
featue: updates node auto select logic

comountainclimber:
- This PR makes it so that by default during automatic node select we exclude ngd and neo.org RPC nodes. This also makes it so that we are not dependent on neoscan for automatic node selection
- This PR also removes depending on neoscan for automatic node selection

### DIFF
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -7,7 +7,7 @@ import { Howl } from 'howler'
 import coinAudioSample from '../assets/audio/coin.wav'
 
 import { getSettings } from './settingsActions'
-import { getNode } from './nodeStorageActions'
+import { getNode, getRPCEndpoint } from './nodeStorageActions'
 import { ASSETS } from '../core/constants'
 import { COIN_DECIMAL_LENGTH } from '../core/formatters'
 import { toBigNumber } from '../core/math'
@@ -72,6 +72,9 @@ async function getBalances({ net, address }: Props) {
   const network = findNetworkByLabel(net)
 
   let endpoint = await getNode(net)
+  if (isEmpty(endpoint)) {
+    endpoint = getRPCEndpoint(net)
+  }
 
   let networkHasChanged = true
   if (net === inMemoryNetwork) networkHasChanged = false
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -11,7 +11,7 @@ import { getNode } from './nodeStorageActions'
 import { ASSETS } from '../core/constants'
 import { COIN_DECIMAL_LENGTH } from '../core/formatters'
 import { toBigNumber } from '../core/math'
-import { findNetworkIdByLabel } from '../core/networks'
+import { findNetworkByLabel } from '../core/networks'
 
 const MAX_SCRIPT_HASH_CHUNK_SIZE = 5
 
@@ -69,7 +69,7 @@ function determineIfBalanceUpdated(
 
 async function getBalances({ net, address }: Props) {
   const { soundEnabled, tokens } = await getSettings()
-  const network = findNetworkIdByLabel(net)
+  const network = findNetworkByLabel(net)
 
   let endpoint = await getNode(net)
 
--- a/app/actions/balancesActions.js
+++ b/app/actions/balancesActions.js
@@ -72,7 +72,7 @@ async function getBalances({ net, address }: Props) {
   const network = findNetworkByLabel(net)
 
   let endpoint = await getNode(net)
-  if (isEmpty(endpoint)) {
+  if (!endpoint) {
     endpoint = await getRPCEndpoint(net)
   }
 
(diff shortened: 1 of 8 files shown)


app/actions/nodeStorageActions.js
```diff
@@ -1,17 +1,78 @@
...
 
+export const getRPCEndpoint = async (
+  net: string,
+  excludeCritera: Array<string> = NODE_EXLUSION_CRITERIA
+) => {
+  try {
+    if (cachedRPCUrl) return cachedRPCUrl
+    const NETWORK = findNetworkByLabel(net)
+    let nodeList
+    switch (NETWORK.id) {
```
evgenyboxer: are we sure that `NETWORK` will contain an `id` ?

app/actions/nodeStorageActions.js
```diff
@@ -1,17 +1,78 @@
...
+    const nodes = data.sort(
+      (a, b) => b.client.lastSeenHeight - a.client.lastSeenHeight
+    )
+    if (nodes.length === 0) throw new Error('No eligible nodes found!')
+    const heightThreshold = nodes[0].client.lastSeenHeight - 1
+    const goodNodes = nodes.filter(
+      n => n.client.lastSeenHeight >= heightThreshold
+    )
+    const randomlySelectedRPCUrl =
+      goodNodes[Math.floor(Math.random() * goodNodes.length)].client.net
```
evgenyboxer: I think its better to use lodash `_.random` - https://lodash.com/docs/4.17.11#random

app/actions/icoTokensActions.js
```diff
@@ -20,7 +20,7 @@ async function getICOTokens({ net, tokens }: Props) {
 
   let endpoint = await getNode(net)
   if (isEmpty(endpoint)) {
```
evgenyboxer: see comment above ^

app/modules/sale.js
```diff
@@ -38,7 +38,11 @@ export const participateInSale = (
   const address = getAddress(state)
   const isHardwareLogin = getIsHardwareLogin(state)
   const signingFunction = getSigningFunction(state)
-  const url = await getNode(net)
+  let url = await getNode(net)
+
+  if (isEmpty(url)) {
```
evgenyboxer: see comment above regarding `isEmpty` ^

app/modules/transactions.js
```diff
@@ -113,7 +113,12 @@ export const sendTransaction = ({
     const signingFunction = getSigningFunction(state)
     const publicKey = getPublicKey(state)
     const isHardwareSend = getIsHardwareLogin(state)
-    const url = await getNode(net)
+    let url = await getNode(net)
+    if (isEmpty(url)) {
```
evgenyboxer: see comment above regarding `isEmpty` ^

app/modules/transactions.js
```diff
@@ -113,7 +113,12 @@ export const sendTransaction = ({
...
+      url = await getRPCEndpoint(net)
+    }
+
+    console.log({ url })
```
evgenyboxer: Sneaky `console.log`

app/actions/balancesActions.js
```diff
@@ -69,9 +69,12 @@ function determineIfBalanceUpdated(
 
 async function getBalances({ net, address }: Props) {
   const { soundEnabled, tokens } = await getSettings()
-  const network = findNetworkIdByLabel(net)
+  const network = findNetworkByLabel(net)
 
   let endpoint = await getNode(net)
+  if (isEmpty(endpoint)) {
```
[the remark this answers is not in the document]
comountainclimber: @evgenyboxer if you refer to the documentation for electron-json-storage, when the key does not exist and empty object is returned https://www.npmjs.com/package/electron-json-storage#storagegetkey-options-callback thus `isEmpty` works really well here because:
- On a clean install the node will always be an empty object
- If the user manually selects "automatic" node selection we save an empty string for the key

app/actions/nodeStorageActions.js
comountainclimber: It must contain an ID because no matter what a NetworkItemType is always returned 
```
export const findNetworkByLabel = (networkLabel: string): NetworkItemType => {
  const networks = getNetworks()
  return networks.find(({ label }) => networkLabel === label) || networks[0]
}
```

app/actions/nodeStorageActions.js
comountainclimber: 👍

app/actions/nodeStorageActions.js
```diff
@@ -1,17 +1,78 @@
...
+    return randomlySelectedRPCUrl
+  } catch (error) {
+    console.log('An error occurred attempting to obtain RPC endpoint', {
+      error
+    })
+    return ''
+  }
+}
+
 export const getNode = async (net: string): Promise<string> =>
```
evgenyboxer: re: your comment - the promise may also return an empty object, could you amend the return type?

comountainclimber: ✔️

PR #1679, opened 2018-11-12
Changing SPOT contract address

daghdha:
we had to change our smart contract for SPOT before the distribution so this is the new contract address. also updated the supply as the supply number was incorrect.

There are some issues because we are pointing to the old smart contract and some users are getting confused as the neon-wallet points to the old smart contract instead of the new one.  

That users will see their correct amount as well as able to transfer tokens correctly.

I updated the script hash and 

I verified the script hash with the developers and they believe that is the problem. 

no, just a change in the constant. 

- [ ] Unit tests written?

### DIFF
--- a/app/core/tokenList.json
+++ b/app/core/tokenList.json
@@ -696,9 +696,9 @@
     "networks": {
       "1": {
         "name": "Spotcoin",
-        "hash": "0a91cdc3c5ff89983c79e3c72e1ccd9e5beaa5d5",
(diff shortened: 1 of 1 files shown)


daghdha:
I assume that I should be doing this in the dev branch.  as we are coming close to the distribution, wondering when this would be merged into the master?

comountainclimber:
closing because of merge here https://github.com/CityOfZion/neon-wallet/pull/1681

PR #1697, opened 2018-11-14
[#1640] Added wallet import name duplication validation

### LINKED ISSUE #1640: There is no validation on wallet name while importing new wallets from *.json file
`design-v2` branch (b2519ef)

**Environment:**

**Reproduction steps:**
1. Create a new wallet named "test"
2. Login to newly created wallet
3. Click on "Settings" button in the left menu
4. Click on "Export" button and save *.json file anywhere
5. Close your Neon wallet and clear it's app data (fresh Neon wallet installation)
6. Start Neon wallet
7. Create a new wallet named "test"
8. Login to newly created wallet
9. Click on "Settings" button in the left menu
10. Click on "Import" button and choose previously saved *.json file
11. Logout from current wallet
12. Click on drop-down with saved wallet => you can see 2x "test", both of them will login you to the wallet which was created as 1st

**Description:**
If you have a saved wallet named "test" and then you import a *.json file with another wallet which has exactly the same name ("test") - you will end up having 2x "test" wallet in the app. You're able to login only to first wallet created (saved wallet in this case), doesn't matter which "test" wallet you're going to choose from drop-down list (1st "test" or 2nd "test") - it will always log you in to the 1st one.

**Screenshots/Video:**
Video is taken from 6th reproduction step (after saving *.json file with a wallet named "test"):
https://i.imgur.com/h3LEgkq.gifv

ranbena:
Fixes #1640 

Showing error message if account label in imported wallet already exists in a current account.
Intersecting account labels in current wallet and imported wallet as an indication of dupe name.

import

### DIFF
--- a/app/modules/generateWallet.js
+++ b/app/modules/generateWallet.js
@@ -1,7 +1,7 @@
 // @flow
 import storage from 'electron-json-storage'
 import { wallet } from 'neon-js'
-import { isEmpty } from 'lodash-es'
+import { isEmpty, intersectionBy } from 'lodash-es'
 
 import {
   showErrorNotification,
@@ -149,6 +149,13 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
 
       if (!accounts.length) {
         reject(Error('No accounts found in recovery file.'))
+      }
+
+      // check if wallet label already exists
+      const duplicateLabels = intersectionBy(data.accounts, accounts, 'label')
+
+      if (duplicateLabels.length > 0) {
+        reject(Error('A wallet with this name already exists locally.'))
       }
 
       // eslint-disable-next-line


app/modules/generateWallet.js
```diff
@@ -151,6 +151,13 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
         reject(Error('No accounts found in recovery file.'))
       }
 
+      // check if wallet label already exists
+      const duplicateLabels = intersectionBy(data.accounts, accounts, 'label')
+
+      if (duplicateLabels.length) {
```
evgenyboxer: @ranbena I think we should be explicit about number comparison... how about `duplicateLabels.length === 0`?

ranbena: Np, but why tho? It's readable and fail-safe.

ranbena: It's a fetish of mine.. writing code as minimal as possible.

Especially this one 🐒
```
if (~arr.indexOf('foo')) {
  ...
}
```

evgenyboxer: or this one: `if (!0) { ... }` 😄

ranbena: Anywho, I updated the PR.

Issue #1700, opened 2018-11-16
Unable to send RHT tokens


dondaniel83:
**Wallet version or commit SHA:**
Neon Wallet 2.0.0

**Environment:**
Win 7 Ultimate

**Prerequisites:**
Using ledger for sending 1 RHT token to another wallet

**Reproduction steps:**

1. Start Neon Wallet 2.0.0 with ledger
2. Click on Send buttom
3. Select RHT Asset
4. Add ammount: 1
5. Add recipient adress: private wallet

In summary:
I'm not able to send the token RHT to another wallet, I got this error:

"You can only send RHT up to 0 decimals". 
The ammount used is 1 with no decimals.

Screenshots:

![rht_error](https://user-images.githubusercontent.com/43292863/48624415-9fd54100-e9ac-11e8-9c87-104d550726c7.jpg)

drptbl:
Re-tested, looks good now.

Aesedai:
Where can we get the updated exe for this? I'm still getting this error.

drptbl:
@Aesedai it was not released yet.

PR #1708, opened 2018-11-18
fix for zero decimal token size calculation bug

hal0x2328:
#1700 

The regression issue with Neon being unable to send tokens with 0 decimals again. This is caused by a comparison relying on the following assignment in Send.jsx:
```const decpoint = amountNum.toString().length - 1 - amountNum.toString().indexOf('.')```

When no decimals are provided, `indexOf('.')` returns a value of `-1`, which causes the expression to evaluate to `1` instead of `0`, so the comparison of ```decpoint > toNumber(get(foundToken, 'decimals', 8))``` below it results in an error for zero-decimal tokens, since `1 > 0`.

Added logic to ensure zero decimals in the input translates to zero decimals in the comparison check in Send.jsx::validateAmount

Tested sending RHT on MainNet with the patch applied

No

No

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Send/Send.jsx
+++ b/app/containers/Send/Send.jsx
@@ -383,8 +383,11 @@ export default class Send extends React.Component<Props, State> {
     }
 
     if (asset !== 'NEO' && asset !== 'GAS') {
-      const decpoint =
-        amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
+      let decpoint = 0
+      if (amountNum.toString().indexOf('.') > -1) {
+        decpoint =
+          amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
+      }
 
       const foundToken: TokenItemType | void = tokens.find(
         token => token.symbol === asset && token.networkId === networkId
--- a/app/containers/Send/Send.jsx
+++ b/app/containers/Send/Send.jsx
@@ -383,17 +383,22 @@ export default class Send extends React.Component<Props, State> {
     }
 
     if (asset !== 'NEO' && asset !== 'GAS') {
-      let decpoint = 0
-      if (amountNum.toString().indexOf('.') > -1) {
-        decpoint =
-          amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
+      let decimalPlaces = 0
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/containers/Send/Send.jsx [hunk 1]
```diff
@@ -383,8 +383,11 @@ export default class Send extends React.Component<Props, State> {
     }
 
     if (asset !== 'NEO' && asset !== 'GAS') {
-      const decpoint =
-        amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
+      let decpoint = 0
+      if (amountNum.toString().indexOf('.') > -1) {
+        decpoint =
+          amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
```
evgenyboxer: @hal0x2328 can we clean this up to be something like:
```
let decimalPlaces = 0
const amountStr = amountNum.toString();
const decimalPointIndex = amountStr.indexOf('.')

if (decPointIndex !== -1) { // or > -1
  decimalPlaces = amountStr.length - 1 - decPointIndex
}
```

Basically removed the `toString()` x3 and the `.indexOf` x2.

hal0x2328: Refactored, tested and pushed the suggested changes

app/containers/Send/Send.jsx
```diff
@@ -383,17 +383,22 @@ export default class Send extends React.Component<Props, State> {
...
-      let decpoint = 0
-      if (amountNum.toString().indexOf('.') > -1) {
-        decpoint =
-          amountNum.toString().length - 1 - amountNum.toString().indexOf('.')
+      let decimalPlaces = 0
+      const amountStr = amountNum.toString()
+      const decPointIndex = amountStr.indexOf('.')
+
+      if (decPointIndex !== -1) {
+        decimalPlaces = amountStr.length - 1 - decPointIndex
```
ranbena: My suggestion (perf and readability)
```
let decimalPlaces = 0
if (!Number.isInteger(amountNum)) {
  const amountStr = amountNum.toString()
  decimalPlaces = amountStr.length - 1 - amountStr.indexOf('.')
}
```

PR #1717, opened 2018-11-21
[#1715] Fixed wrong chart currency when changing duration

### LINKED ISSUE #1715: Graph values are shown in USD values after changing time frame (while having default currency set to any other than USD)
`dev` branch (fad203a)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.376

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Settings" button in the left menu
3. Change default currency to any other than USD
4. Click on "Wallet" button in the left menu
5. Click on "1 DAY" and change it to any other time frame
6. Hover your mouse on the graph

**Description:**
After hovering your mouse on the last reproduction step, you can see that values on the graph are shown in USD values, even if you've chosen any other currency in settings.

**Screenshots/Video:**
market_data_err

ranbena:
Fixes #1715 

The duration change action works with `duration` and `currency`
```
const mapPriceHistoryActionsToProps = (actions, props) => ({
  setDuration: (duration: Duration) => {
    props.setDuration(duration)
    return actions.call({ duration, currency: props.currency })
  }
})
```
But I noticed `withCurrencyData` doesn't update `props` until _after_ the actions are mapped - hence I assume it falls back to default currency which is the bug reported.

Bumped it up and it's works now🤘

### DIFF
--- a/app/components/Dashboard/PriceHistoryPanel/index.js
+++ b/app/components/Dashboard/PriceHistoryPanel/index.js
@@ -31,11 +31,11 @@ export default compose(
   // Refetch price history if the selected asset or duration changes
   withState('asset', 'setAsset', ASSETS.NEO),
   withState('duration', 'setDuration', '1d'),
+  withCurrencyData('currency'),
   withActions(priceHistoryActions, mapPriceHistoryActionsToProps),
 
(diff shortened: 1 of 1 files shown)


comountainclimber:
Great catch @ranbena 👍 keep up the fantastic work

PR #1720, opened 2018-11-22
[#1689] Integrated reusable TransactionList into SendSuccess

### LINKED ISSUE #1689: Disable "Add" buttons on transaction confirmation page next to addresses which are added to contacts list
Also instead of wallet addresses, we should show their names from contact list?

added

Correct example from "Activity" page:
hist

ranbena:
Fixed #1689 

### The problem
Code duplication with discrepancies was preventing similar transaction components to display and behave consistently across different controllers. Although a simple request - add "disabled" state to "add" button in transaction summary, I didn't want to contribute to the divide and [elaborated here](https://github.com/CityOfZion/neon-wallet/issues/1689#issuecomment-440404743).

### The solution
I wanted to make `<Transaction/>` reusable and allow any page to incorporate it, making it ,mildly more configurable and performant so it could be integrated into both Activity and Send pages.

So now it just works :P Activity page remains unaltered and now Send page has feature parity.

boxer

### Reviewing the PR
I split the PR into 5 commits:
1) A little perf improvement for "findContact"
2) Since Send page  transaction shows no date, made it render optionally in `<Transaction />`
3) The heart of the matter:
```
<TransactionList className={className} alternateRows>
  {transactions.map((tx, i) => (
    <Transaction tx={tx} key={i} />
  ))}
</TransactionList>
```
4) Renaming css variable --tx-history to --tx-list to accommodate the newly created `<TransactionList>`
5) The big moment! Swapping `SendSuccessTransaction` with `TransactionList`!


app/components/Blockchain/Transaction/index.js
```diff
@@ -17,7 +18,9 @@ const mapDispatchToProps = dispatch => ({
     dispatch(showModal(MODAL_TYPES.ADD_CONTACT, props))
 })
 
-const mapContactsDataToProps = (contacts: Object) => ({ contacts })
+const mapContactsDataToProps = (contacts: Object) => ({
+  contacts: invert(contacts)
+})
```
ranbena: Since `contacts` is a `{ name: address}` dict, instead of looping it to find key by value, I invert it to be `{ address: name }` and now all you need is `name = contacts[address]`🏎

app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -79,9 +74,9 @@ export default class Transaction extends React.Component<Props> {
 
   renderAbstract = (type: string) => {
     const { tx } = this.props
-    const { iconType, time, label, amount, isNetworkFee, to, from } = tx
 
     const formattedTime = moment.unix(time).format('MM/DD/YYYY | HH:mm:ss')
+    const { time, label, amount, isNetworkFee, to, from } = tx
```
ranbena: Oops shouldn't be here but `iconType` isn't being used anyhow..

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.scss
```diff
@@ -22,3 +22,7 @@
 .oddNumberedRow {
   background: #f8f8f8;
 }
+
+.transactions {
+  margin: -24px;
```
ranbena: Moved from the now generic `Transactions.scss .transactionList` to the more specific `TransactionHistoryPanel.scss .transactions` which it's meant for.

app/components/Blockchain/Transaction/index.js
comountainclimber: Haha this is really clever never used lodash invert before 🔥

app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -46,12 +46,7 @@ export default class Transaction extends React.Component<Props> {
   findContact = (address: string): Node => {
     const { contacts } = this.props
     if (contacts && !isEmpty(contacts)) {
-      let label
-      Object.keys(contacts).forEach(key => {
-        if (contacts[key] === address) {
-          label = key
-        }
-      })
+      const label = contacts[address]
```
comountainclimber: 👍

app/components/Send/SendPanel/SendSuccess/index.jsx
```diff
@@ -23,6 +24,14 @@ export default class SendSuccess extends React.Component<Props> {
 
   render() {
     const { sendRowDetails, txid } = this.props
+    // normalize tx data for Transaction
+    const transactions = sendRowDetails.map(row => ({
+      iconType: 'SEND',
```
comountainclimber: I really like this approach 😄- now that iconType is being used across the application do you mind adding an exported enum  to `constants.js` for these constant values?

Issue #1580, opened 2018-10-23
Add descriptions to payment request details


drptbl:
**Wallet version or commit SHA:**
`design-v2` branch (9e2d017)

It would be nice to add descriptions in payment request details, for example "Amount", "Address" and "Description".

**Example:**
<img width="574" alt="screenshot 2018-10-21 at 18 35 05" src="https://user-images.githubusercontent.com/8177587/47376273-74836d00-d6f2-11e8-9ce5-87cfd27df82b.png">

drptbl:
Looks good 👍 . Thanks :)

PR #1723, opened 2018-11-24
feature: adds labels to payment request details

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1580

This PR labels the details of a nep9 payment request for UI clarity... It also updates the labels in Receive for consistent verbage and style

Manually reviewing the UI.

screen shot 2018-11-24 at 10 41 15 am

screen shot 2018-11-24 at 10 38 40 am

This PR also updates the style of the back button in modals as it was using the wrong color previously.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/BaseModal/BaseModal.scss
+++ b/app/components/Modals/BaseModal/BaseModal.scss
@@ -40,7 +40,7 @@
 
   svg {
     path {
-      fill: var(--panel-full-height-nav-background);
+      fill: var(--panel-full-height-header-icon-color);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


app/components/Modals/BaseModal/BaseModal.scss
```diff
@@ -40,7 +40,7 @@
 
   svg {
     path {
-      fill: var(--panel-full-height-nav-background);
+      fill: var(--panel-full-height-header-icon-color);
```
comountainclimber: Also realized we are using the wrong color here in 2.0.0 (no issue currently open)

PR #1726, opened 2018-11-26
[#1640] Fixed import wallet non-terminating execution bug

ranbena:
@drptbl reported dup wallet importing shows error but persists the import. It's a bug I introduced in #1716. SHAME 😓

#### Solution
TIL `reject` doesn't terminate `Promise` execution.. Added the missing `return` after rejection.


comountainclimber:
Good catch 👍

Issue #1712, opened 2018-11-19
User node selection/reporting improvement suggestions


hal0x2328:
The new version of Neon has been well-received but still has minor issues with users experiencing
node issues as if they were Neon issues. Neon should have a very clear indicator on the dashboard
of the current node that is being used, and perhaps a health indicator for that node as well.

Also, while users can manually select a node during times of network issues, it may cause a later 
problem if that node gets behind or starts experiencing latency - the user must continually remember to go back to settings to select another node or go back to automatic mode - IMHO the node selection widget should be accessible directly from the dashboard.

Finally, sorting nodes by block height and ping time combined may not be the best approach - I think
the list should prefer block height over ping time, with limits for both. Since the list only shows the
"top 15" nodes, it causes out-of-date nodes to block out current nodes from being selected simply because the out-of-date nodes have less latency in responding.

PR #1727, opened 2018-11-27
feature: node cache/storage expiration

comountainclimber:
This PR begins the iterative process of issue/enhancement https://github.com/CityOfZion/neon-wallet/issues/1712 - Specifically it adds an expiration to the "best node" algo and also sets an expiration on manually set nodes... This will resolve a number of _possible_ latent issues around the current set up

Because nodes could potentially fall behind go offline etc leaving them in memory (or local storage) permanently is generally a bad idea

node url is set in storage with a timestamp which is always compared to the current time... if its expired the node selection algorithm is run

By manual and automated test coverage

- [x] Unit tests written? 😎

### DIFF
--- a/app/actions/nodeStorageActions.js
+++ b/app/actions/nodeStorageActions.js
@@ -1,6 +1,6 @@
 // @flow
 import { createActions } from 'spunky'
-import { isEmpty, random, get } from 'lodash-es'
+import { isEmpty, random, get, compact } from 'lodash-es'
 import { rpc, api } from 'neon-js'
 
 import { getStorage, setStorage } from '../core/storage'
@@ -31,15 +31,16 @@ export const determineIfCacheIsExpired = (
   expiration: number = CACHE_EXPIRATION
 ): boolean => timestamp + expiration < new Date().getTime()
 
-export const parseUrlData = (data: {
+export const buildNodeUrl = (data: {
   height: number,
   port: string,
   protocol: string,
   url: string
 }): string => {
-  let url = data.protocol ? `${data.protocol}://${data.url}` : data.url
-  url = data.port ? `${url}:${data.port}` : url
-  return url
+  const { protocol, url, port } = data
+  return compact([protocol && `${protocol}://`, url, port && `:${port}`]).join(
+    ''
+  )
 }
 
 export const getRPCEndpoint = async (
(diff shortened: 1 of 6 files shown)


app/actions/nodeStorageActions.js
```diff
@@ -81,14 +108,20 @@ export const getRPCEndpoint = async (
...
+export const getNode = async (net: Net): Promise<string> => {
   const storage = await getStorage(`${STORAGE_KEY}-${net}`).catch(console.error)
-  if (!storage) return ''
-  return isEmpty(storage) ? '' : storage
+  const nodeInStorage = get(storage, 'node')
+  const expiration = get(storage, 'timestamp')
+  if (!nodeInStorage) return ''
+  if (!expiration) return ''
+  if (determineIfCacheIsExpired(expiration)) {
+    return ''
```
evgenyboxer: these expressions can be combined into one `if (! nodeInStorage || !expiration || !determineIfCacheIsExpired(expiration)) { return '' }`

flow-typed/declarations.js
```diff
@@ -28,6 +29,8 @@ declare type ReduxAction = () => {
 
 declare type NetworkType = string
 
+declare type NetworkLabelTypes = $Values<typeof NETWORK_LABELS>
```
evgenyboxer: hmm is this actually working? thinking if this should just be `MAIN_NETWORK_LABEL | TEST_NETWORK_LABEL`

app/actions/nodeStorageActions.js
```diff
@@ -15,19 +15,44 @@ import { findNetworkByLabel } from '../core/networks'
...
+
+export const parseUrlData = (data: {
+  height: number,
+  port: string,
+  protocol: string,
+  url: string
+}): string => {
+  let url = data.protocol ? `${data.protocol}://${data.url}` : data.url
+  url = data.port ? `${url}:${data.port}` : url
+  return url
```
evgenyboxer: I think the function name is a bit confusing, it's more like `buildURL`

Regarding the implementation, how about using lodash compact and an array that you join? (https://lodash.com/docs/4.17.11#compact)


```
const { protocol, url, port } = url

return compact([
  protocol && `${protocol}://`,
  url,
  port && `:${port}`
]).join('')
```

flow-typed/declarations.js
comountainclimber: When using this syntax initially there was previously an error that seems to have disappeared ✨ thanks @evgenyboxer

comountainclimber:
Thanks for the solid review @evgenyboxer 💪

comountainclimber:
Thanks @dvdschwrtz 😄

PR #1730, opened 2018-11-28
[#1318] PrepWork 1/3: Added util/poll.js unit tests

ranbena:
Preparation work for solving #1318.

**tl;dr** Added unit tests to `util/poll.js` before screwing with it

#### The #1318 Pull Request stack
1. (#1730) PrepWork 1/3: Added util/poll.js unit tests ☜☜☜ _**YOU ARE HERE**_
2. (#1731) PrepWork 2/3: Made util/poll.js cancellable with cancellablePoll
3. (#1731) PrepWork 3/3: Added cancellablePoll unit tests
4. (#1732) Actual 1/2: Enable “Scan QR” button only if camera available
5. (#1733) Actual 2/2: Added camera availability unit tests

#### What's this commit about?
This is prepare work for the actual feature - Enabling/disabling “Scan QR” button depending on camera availability.
Since the next commit will be adding functionality to `util/poll.js` (making it cancellable), it would be wise to first cover the existing functionality with a few tests, so we're confident nothing breaks.

**Tests:**
1. Should stop when condition fulfilled
2. Should stop when max attempts reached

- [ ] Unit tests written?

Issue #1318, opened 2018-07-10
Disable "Scan QR Code" button if camera is not available while logging in with private key


drptbl:
**Wallet version or commit SHA:**
`dev` branch (ae58feb)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.112

**Pre-requisites:**
No camera available (VM for example)

**Reproduction steps:**
1. Start Neon Wallet
2. Click on "Login with private key" button

If there is no camera available when going to "Login with private key" page, "Scan QR Code" button should be disabled. Currently if button is clicked, there's an error thrown in console however it's not visible in the wallet UI, so users are not informed about this. Also button changes to "Cancel" - it shouldn't happen in case of error.

If it's not possible to check camera and disable this button, we should at least show the error after clicking on it.

**Screenshot/Video:**
![show_error](https://user-images.githubusercontent.com/8177587/42522837-6cd495c6-846c-11e8-870c-e2ad1e8d1fef.PNG)

ranbena:
I'll take this

tbeck860:
Fix tested successfully on Windows 10 Pro v1803 Build 17134.523
![1-15-2019 5-18-13 pm](https://user-images.githubusercontent.com/31013176/51218525-15e12000-18ea-11e9-9771-156edeada781.png)

comountainclimber:
Please award @tbeck860 100 🚀

lllwvlvwlll:
https://neoscan.io/transaction/220a87f0e9cedf43f05278182794d9c5f49ad6f144e9ebc7fb7aa4b5994c68d8

PR #1731, opened 2018-11-28
[#1318] PrepWork 2+3/3: Added cancellable ability to util/poll.js

ranbena:
Preparation work for solving #1318.

**tl;dr** Added cancellablePoll

NOTICE: This PR is dependent on #1730. Please defer review until merged.

#### The #1318 Pull Request stack
1. (#1730) PrepWork 1/3: Added util/poll.js unit tests
2. (#1731) PrepWork 2/3: Made util/poll.js cancellable with cancellablePoll
3. (#1731) PrepWork 3/3: Added cancellablePoll unit tests ☜☜☜ _**YOU ARE HERE**_
4. (#1732) Actual 1/2: Enable “Scan QR” button only if camera available
5. (#1733) Actual 2/2: Added camera availability unit tests

#### What's this commit about?
This is prepare work for the actual feature - Enabling/disabling “Scan QR” button depending on camera availability.
The main feature (next commit) polls continuously for camera availability. It's important to cancel the polling if the user clicks away from the login tab.

**Tests:**
Previous poll tests keep passing.
Added new "should be cancellable" test

- [ ] Unit tests written?

### DIFF
--- a/app/util/poll.js
+++ b/app/util/poll.js
@@ -3,17 +3,24 @@ import delay from './delay'
 const DEFAULT_ATTEMPTS = Infinity
 const DEFAULT_FREQUENCY = 1000
 
+export class TimeoutError extends Error {
+  constructor() {
+    super('Poll reached maximum attempts')
+    this.name = 'TimeoutError'
+  }
+}
+
 export default function poll(
   request,
   { attempts = DEFAULT_ATTEMPTS, frequency = DEFAULT_FREQUENCY } = {}
 ) {
-  return request().catch(function retry(err) {
+  return request().catch(function retry() {
     // eslint-disable-next-line
     if (attempts-- > 0) {
       return delay(frequency)
         .then(request)
         .catch(retry)
     }
-    throw err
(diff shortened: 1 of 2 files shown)


app/util/poll.js
```diff
@@ -3,17 +3,47 @@ import delay from './delay'
...
+  }
+}
+
+export function cancellablePoll(
   request,
   { attempts = DEFAULT_ATTEMPTS, frequency = DEFAULT_FREQUENCY } = {}
 ) {
-  return request().catch(function retry(err) {
+  let hasCancelled = false
+  const promise = request().catch(function retry() {
```
comountainclimber: Maybe Im missing something but wouldn't it be better to define `retry` outside of this catch block so that it doesn't get redefined every time `request` throws an error?

ranbena: Indeed. I'll change, test and PR separately.

ranbena: Pull request #1734

PR #1732, opened 2018-11-28
[#1318] Actual 1/2: Enable “Scan QR” button only if camera available

### LINKED ISSUE #1318: Disable "Scan QR Code" button if camera is not available while logging in with private key

ranbena:
Fixes #1318.

**tl;dr** The actual feature work

NOTICE: This PR depends on #1730 and #1731 to be merged so defer review till then plz.

#### The #1318 Pull Request stack
1. (#1730) PrepWork 1/3: Added util/poll.js unit tests
2. (#1731) PrepWork 2/3: Made util/poll.js cancellable with cancellablePoll
3. (#1731) PrepWork 3/3: Added cancellablePoll unit tests
4. (#1732) Actual 1/2: Enable “Scan QR” button only if camera available  ☜☜☜ _**YOU ARE HERE**_
5. (#1733) Actual 2/2: Added camera availability unit tests

#### What's this commit about?
The main feature finally. Basically, once the Private Key Login tab is clicked, the "Scan QR" button isn't enabled until a camera source is found (not "active" but rather "available" in the OS).

This enhancement to `LoginPrivateKey` is implemented with an HOC `withCameraAvailability` which adds a prop `cameraAvailable` to the wrapped component and automatically toggles the boolean value according to actual camera availability.

The availability is determined depending on [`navigator.mediaDevices.enumerateDevices()`](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices). This required updating Flow's BOM declarations (this browser feature isn't yet supported in IE 😪...)

The listener is implemented with the `util/poll.js` polling mechanism. When the state of availability changes (for instance, a usb camera was connected), the polling ends, and another is set listening to the opposite value (when the camera is disconnected). Eventually, polling is stopped when component is unmounted.

LETS SEE IT IN ACTION
ezgif com-video-to-gif

### DIFF
--- a/app/hocs/withCameraAvailability.js
+++ b/app/hocs/withCameraAvailability.js
@@ -1,0 +1,99 @@
+// @flow
+import React from 'react'
+import { filter } from 'lodash-es'
+import { cancellablePoll as poll } from '../util/poll'
+import type { CancellableType } from '../util/poll'
+
+const POLL_FREQUENCY = 1000
+
+type State = {
+  avail: boolean
+}
+
+type Props = {}
+
+// $FlowFixMe
+export default function withCameraAvailability(Component) {
+  class CameraAvailability extends React.Component<Props, State> {
+    config = { frequency: POLL_FREQUENCY }
+
+    cancellable: CancellableType
+
+    state = {
+      avail: false
+    }
+
+    componentDidMount() {
+      // start wait for available camera
+      this.awaitAvailable()
+    }
+
+    componentDidUpdate(_: Props, prevState: State) {
+      // cleanup
+      this.cancellable.cancel()
+
+      // if camera found, wait until no longer available
+      // if no camera found, wait until available
+      if (prevState.avail) {
+        this.awaitNotAvailable()
+      } else {
+        this.awaitAvailable()
+      }
+    }
+
+    componentWillUnmount() {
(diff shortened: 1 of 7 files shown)


app/hocs/withCameraAvailability.js
```diff
@@ -0,0 +1,99 @@
...
+const POLL_FREQUENCY = 1000
+
+type State = {
+  avail: boolean
+}
+
+type Props = {}
+
+// $FlowFixMe
+export default function withCameraAvailability(Component) {
```
comountainclimber: Awesome use case for a HOC 👍

PR #1735, opened 2018-11-30
feature: lock dependency versions in package.json

comountainclimber:

This PR locks all dependency versions to avoid issues such as https://github.com/dominictarr/event-stream/issues/116 from affecting neon-wallet.

- [ ] Unit tests written?


dvdschwrtz:
Nice work. I will be working on package management solutions outside of CoZ and will provide feedback on how to maintain high security standards

Issue #743, opened 2018-02-20
Opening links in a new window should open the default system browser


metachris:
When I get the "new version is available" popup, with the link to GitHub from here, it's opening a new Electron window: https://github.com/CityOfZion/neon-wallet/blob/b70de8b40c961db544638069a0270b3b63d69525/app/modules/metadata.js#L20

It would be better if it would open those links in the default system browser. There I would get download progress for instance, which I don't get in the new Electron window.

This could be done by adding an event catcher in main.js here: https://github.com/CityOfZion/neon-wallet/blob/dev/main.js#L37

It would look similar to this to open all external URLs inside the default system browser:

    mainWindow.webContents.on('will-navigate', (e, url) => {
        console.log('will-nav', url, mainWindow.webContents.getURL())
        if (url.substr(0, 21) !== 'http://localhost:3000') {
            e.preventDefault()
            shell.openExternal(url)
        }
    })

drptbl:
Re-tested, it's fixed.

lllwvlvwlll:
# THIS IS JUST AN EXAMPLE WORKFLOW
Please Award:

@metachris : 500
@drptbl : 200

lllwvlvwlll:
(Presending to be @canesin )
@metachris https://neoscan.io/transaction/a11a5ac8ae835f75348b79640ae74424fced637e1ef89a1bd5e1d13e6d9faa16

@drptbl 
https://neoscan.io/transaction/a11a5ac8ae835f75348b79640ae74424fced637e1ef89a1bd5e1d13e6d9faa16

PR #1738, opened 2018-12-03
feature: updates version check notification

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/743

- The link in the version check notification will now behave like the other links found throughout the application
- Fixes a few issues related to styles (low contrast problems)

BEFORE:
screen shot 2018-12-02 at 8 49 52 pm

AFTER:
screen shot 2018-12-02 at 8 48 46 pm

By using electron's `shell.openExternal` https://github.com/electron/electron/blob/master/docs/api/shell.md#shellopenexternalurl-options-callback

Manual testing...

`NotificationArgsType` has been updated to include children - see the React Notificaiton System documentation https://github.com/igorprado/react-notification-system#children

- [ ] Unit tests written?

### DIFF
--- a/app/modules/metadata.js
+++ b/app/modules/metadata.js
@@ -3,7 +3,7 @@ import React from 'react'
 import axios from 'axios'
 import retry from 'axios-retry'
 import compareVersions from 'compare-versions'
-import { shell } from 'electron'
+import { ConditionalLink } from '../util/ConditionalLink'
 
 import { showWarningNotification } from './notifications'
 import {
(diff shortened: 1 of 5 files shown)


comountainclimber:
Sure I can update to just allow message to be either or... will require changing it in one place or another - https://github.com/CityOfZion/neon-wallet/blob/a414ba38edb23077a766e9183df9d856a41231e9/app/containers/Notifications/Notifications.jsx#L62

app/modules/metadata.js
```diff
@@ -53,7 +54,15 @@ export const checkVersion = () => async (dispatch: DispatchType) => {
...
-        `Your wallet is out of date! Please download the latest version from ${DOWNLOAD_LINK}`
+        <div>
+          Your wallet is out of date! Please download the latest version from{' '}
+          <div
+            className="notification-download-link"
+            onClick={() => shell.openExternal(NEON_WALLET_RELEASE_LINK)}
+          >
+            {NEON_WALLET_RELEASE_LINK}
+          </div>
+        </div>
```
ranbena: Perhaps use `<ConditionalLink>` which already opens an external browser (in #1746):
```jsx
<ConditionalLink href={NEON_WALLET_RELEASE_LINK}>
  {NEON_WALLET_RELEASE_LINK}
</ConditionalLink>
```
Pros:
a) Readable and concise
b) Link would behave and style as one
c) Optional `tooltip` attr

comountainclimber: 👍 awesome

Issue #1682, opened 2018-11-12
Names for custom tokens are missing in "Activity" page


drptbl:
After sending a custom token, it's name is missing on "Activity" page.

![rpx_missing](https://user-images.githubusercontent.com/8177587/48369713-a5364100-e6b7-11e8-9f1a-1d471402792b.PNG)

PR #1739, opened 2018-12-03
feature: adds custom tokens to activity page

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1682

Currently is a user adds a script hash the symbol type will not show in the activity page. This PR fixes that problem.
screen shot 2018-12-02 at 9 56 54 pm

This PR uses neon-js's `getToken` to fill missing symbol data in the activity page for the edge case user's token is not included in `tokenList.json`

By manually testing

If request for token info fails for whatever reason the code will default to the current behavior in 2.0.0

- [ ] Unit tests written?

### DIFF
--- a/app/actions/transactionHistoryActions.js
+++ b/app/actions/transactionHistoryActions.js
@@ -66,9 +66,9 @@ async function parseAbstractData(data, currentUserAddress, tokens, net) {
   }
 
   const results = []
-  // eslint-disable-next-line
+  // eslint-disable-next-line no-restricted-syntax
   for (const abstract of data) {
-    // eslint-disable-next-line
+    // eslint-disable-next-line no-await-in-loop
     const asset = await parsedAsset(abstract)
     const type = parsedTxType(abstract)
     const summary: TxEntryType = {
--- a/app/actions/transactionHistoryActions.js
+++ b/app/actions/transactionHistoryActions.js
@@ -3,6 +3,7 @@ import axios from 'axios'
 import { api } from 'neon-js'
 import { createActions } from 'spunky'
 
+import { getRPCEndpoint } from './nodeStorageActions'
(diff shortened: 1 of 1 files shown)


app/actions/transactionHistoryActions.js
```diff
@@ -55,8 +65,11 @@ function parseAbstractData(data, currentUserAddress, tokens) {
     return abstract.address_from
   }
 
-  return data.map(abstract => {
-    const asset = parsedAsset(abstract)
+  const results = []
+  // eslint-disable-next-line
```
evgenyboxer: what's the reason for disabling the lint check for the next line?

app/actions/transactionHistoryActions.js
```diff
@@ -55,8 +65,11 @@ function parseAbstractData(data, currentUserAddress, tokens) {
...
+  for (const abstract of data) {
+    // eslint-disable-next-line
```
evgenyboxer: this one too

comountainclimber:
@evgenyboxer do you still have any concerns about the syntax I used? Perhaps I should update the `.eslintrc` file...

evgenyboxer:
@comountainclimber I think its best to disable the specific rule you are targeting

app/actions/transactionHistoryActions.js
```diff
@@ -55,8 +65,11 @@ function parseAbstractData(data, currentUserAddress, tokens) {
     return abstract.address_from
   }
 
-  return data.map(abstract => {
-    const asset = parsedAsset(abstract)
+  const results = []
```
dvd-schwrtz: I would prefer in general to use reduce instead of creating an array and then pushing things into it from a loop

evgenyboxer:
@comountainclimber can we merge?

PR #1745, opened 2018-12-05
[#1576] 2/3: QR Scan loading indicator

### LINKED ISSUE #1576: Add loading indicator for camera preview while trying to scan QR codes
`design-v2` branch (9e2d017)

**Environment:**

**Reproduction steps:**
1. Login to your wallet
2. Click on "Send" button in the left menu
3. Click on "Enter QR Code" button
4. Click on "Capture" button

**Description:**
I'm experiencing a 5secs+ delay before preview show up, seeing a gray empty space for that time. Loading indicator would indicate that we're waiting for something.

**Screenshots/Video:**
https://i.imgur.com/88Kx0fw.gifv

ranbena:
Fixes #1576 

DEPENDS ON #1744. Plz do not merge before #1744 or the commit history could get messy.

It was reported that the webcam preview might take up to 5 secs to initialize and a loading indicator is needed.

Added a loading indicator to `QRCodeScanner` which affects both Private Key Login and Send modules thanks to #1744 😋.

ezgif com-video-to-gif 1
ezgif com-video-to-gif 2

Manually tested that:
a) Other loading indicators remain unaffected.
b) Loading indicator shows up and runs smoothly in both cases.

Plz notice that commits from #1744 are included so only review commit cd1b19e
I've added specific comments in the GitHub commit code.

The cake is a lie.

- [ ] Unit tests written?

### DIFF
--- a/app/components/QrCodeScanner/QrCodeScanner.jsx
+++ b/app/components/QrCodeScanner/QrCodeScanner.jsx
@@ -1,12 +1,23 @@
 // @flow
-import React, { Component } from 'react'
+import React, { Component, Fragment } from 'react'
 import Instascan from 'instascan'
 
+import Loading, { ANIMATION_DURATION } from '../../containers/App/Loading'
+
 type Props = {
-  callback: (content: string, scannerInstance: any) => any
+  callback: (content: string, scannerInstance: any) => any,
+  theme: string
 }
 
-export default class QrCodeScanner extends Component<Props> {
+type State = {
+  loading: boolean
+}
+
+export default class QrCodeScanner extends Component<Props, State> {
+  state = {
+    loading: false
+  }
+
   scanPreviewElement: ?HTMLVideoElement
 
   scannerInstance: Instascan
@@ -29,15 +40,22 @@ export default class QrCodeScanner extends Component<Props> {
       callback(content, this.stopScanner.bind(this))
     })
 
-    Instascan.Camera.getCameras()
+    // since the browser halts while getting usermedia
+    // let the loading animation finish one round and only then continue
+    this.setState({ loading: true })
+    new Promise(resolve => setTimeout(resolve, ANIMATION_DURATION))
+      .then(() => Instascan.Camera.getCameras())
       .then((cameras: Array<Object>) => {
-        if (cameras.length > 0) {
-          this.scannerInstance.start(cameras[0])
-        } else {
-          console.error('No cameras found.')
+        if (cameras.length === 0) {
+          // shouldn't happen, case covered by withCameraAvailability
+          throw new Error('No cameras found.')
         }
+        return this.scannerInstance.start(cameras[0])
       })
       .catch(e => console.error(e))
+      .finally(() => {
+        this.setState({ loading: false })
+      })
   }
 
   stopScanner() {
@@ -46,7 +64,23 @@ export default class QrCodeScanner extends Component<Props> {
 
   render() {
     return (
-      // eslint-disable-next-line
+      <Fragment>
+        {this.renderLoadingIndicator()}
+        {this.renderScanner()}
+      </Fragment>
+    )
+  }
+
+  renderLoadingIndicator() {
+    const { theme } = this.props
+    const { loading } = this.state
+
+    return loading ? <Loading theme={theme} nobackground /> : null
+  }
+
+  renderScanner() {
+    return (
+      /* eslint-disable-next-line jsx-a11y/media-has-caption */
       <video
         ref={ref => {
           this.scanPreviewElement = ref
--- a/app/containers/Home/Home.scss
+++ b/app/containers/Home/Home.scss
@@ -66,6 +66,7 @@ $navigation-margin: 20px;
   display: flex;
   justify-content: center;
   min-height: 150px;
+  position: relative;
   video {
     min-width: 350px;
     margin-top: -40px;
--- a/app/components/QrCodeScanner/index.js
+++ b/app/components/QrCodeScanner/index.js
@@ -1,1 +1,6 @@
-export { default } from './QrCodeScanner'
+import { compose } from 'recompose'
+
+import withThemeData from '../../hocs/withThemeData'
+import QrCodeScanner from './QrCodeScanner'
+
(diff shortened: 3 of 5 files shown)


app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -29,15 +40,22 @@ export default class QrCodeScanner extends Component<Props> {
       callback(content, this.stopScanner.bind(this))
     })
 
-    Instascan.Camera.getCameras()
+    // since the browser halts while getting usermedia
+    // let the loading animation finish one round and only then continue
+    this.setState({ loading: true })
+    new Promise(resolve => setTimeout(resolve, 900))
```
ranbena: **Why force a timeout?**
Without it the loading animation hangs. Look:
ezgif com-video-to-gif 3

**Why does the animation hang?**
Because getting camera access stalls the browser. I validated this with the Chrome profiler. For about a second once `getUserMedia` is triggered the browser does absolutely nothing! I experimented with GPU layering but there doesn't seem to be a way around it.

**Why 900ms?**
That's the time it takes for the animation to complete a cycle,  and so this way, the animation hang isn't noticeable.

app/containers/Home/Home.scss
```diff
@@ -66,6 +66,7 @@ $navigation-margin: 20px;
   display: flex;
   justify-content: center;
   min-height: 150px;
+  position: relative;
```
ranbena: In order for the loading indicator to position itself correctly.

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -46,7 +64,23 @@ export default class QrCodeScanner extends Component<Props> {
...
+        {this.renderScanner()}
+      </Fragment>
+    )
+  }
+
+  renderLoadingIndicator() {
+    const { theme } = this.props
+    const { loading } = this.state
+
+    return loading ? <Loading theme={theme} nobackground /> : null
```
ranbena: Added a `nobackground` attribute since the default grayish background (in Loading.scss) doesn't suit the Send modal.

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -1,12 +1,23 @@
...
-import React, { Component } from 'react'
+import React, { Component, Fragment } from 'react'
 import Instascan from 'instascan'
 
+import Loading from '../../containers/App/Loading'
+
 type Props = {
-  callback: (content: string, scannerInstance: any) => any
+  callback: (content: string, scannerInstance: any) => any,
+  theme: string
```
ranbena: Whoops.. forgot to add `withThemeData` to `index.jsx`. Would force push but the GitHub comments I put in the code would get trampled :( I'll wait till you approve?

app/components/QrCodeScanner/QrCodeScanner.jsx
dvd-schwrtz: Ha, I just asked about this on another PR. Perhaps add a variable name to explain or a comment

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -29,15 +40,22 @@ export default class QrCodeScanner extends Component<Props> {
       callback(content, this.stopScanner.bind(this))
     })
 
-    Instascan.Camera.getCameras()
+    // since the browser halts while getting usermedia
+    // let the loading animation finish one round and only then continue
+    this.setState({ loading: true })
+    new Promise(resolve => setTimeout(resolve, 900))
+      .then(() => Instascan.Camera.getCameras())
       .then((cameras: Array<Object>) => {
```
dvd-schwrtz: Do we want to run validation through something like joi on the returned objects perhaps to make sure the cameras in the array are in the right format? Maybe this is already happening in the getCameras() call. If not, might be a good idea

dvd-schwrtz: Also, as far as using instascan and other libraries, in general the less libraries we use, the better. we will be implementing more security around npm package management moving forward. However, if it is helpful, of course we can use packages

ranbena: > Also, as far as using instascan and other libraries, in general the less libraries we use, the better. we will be implementing more security around npm package management moving forward. However, if it is helpful, of course we can use packages

Indeed I try to use as less packages as possible. This one ain't new though, it's been used up till now. It's not great in terms of maintenance, error handling nor static typing. I think though this type of consideration would be out of scope for this issue, don't you think?

dvd-schwrtz: Yes, was just bringing it up :)

ranbena: > Do we want to run validation through something like joi on the returned objects perhaps to make sure the cameras in the array are in the right format? Maybe this is already happening in the getCameras() call. If not, might be a good idea

Not sure it's a must in this case since it's an `Instascan` method feeding into another. I changed the original code to Promise chaining in order for the `catch` to attain to any errors in this process.

ranbena:
@dvd-schwrtz requested changes pushed

app/components/QrCodeScanner/QrCodeScanner.jsx
comountainclimber: @ranbena this is awesome just curious is there a fixed amount of time it takes for camera to render? Is that dependent on hardware?

app/components/QrCodeScanner/QrCodeScanner.jsx
comountainclimber: Thats fine 👍

app/components/QrCodeScanner/QrCodeScanner.jsx
ranbena: That's a good question. I found no info on the web. Any suggestion/reference would be welcome.

If the render time takes longer on other systems, it would result in a longer time between the completed animation cycle and the actual preview coming up. I think it's the best we got at this time.

PR #1746, opened 2018-12-05
[#1582] 3/3: Added QR scan error handling

### LINKED ISSUE #1582: Handle a case when user denied system permissions for camera
`design-v2` branch (9e2d017)

**Environment:**

**Description:**
After trying to use camera for something for the first time, you have to allow Neon to use your camera (system permissions). If you deny them (purposely or not), all you can see is gray space instead of camera preview. We could detect if user denied permissions and show proper message or tutorial on how to fix this.

**Screenshots/Video:**
https://i.imgur.com/jURrGJd.gifv

ranbena:
Fixes #1582 

DEPENDS ON #1745. Plz do not merge before #1745 or the commit history could get messy.

Firing up the webcam for QR scanning could fail for the following reasons:
a) Camera already in use by other program (I read it [here](https://stackoverflow.com/questions/43919639/what-is-a-trackstarterror))
b) App denied access to camera (easily reproducible)
c) Some unknown reason

Our UI doesn't reflect that to the user:

screen shot 2018-12-05 at 14 34 56

I now catch all errors and show a "Could not connect to camera" error where the webcam preview is expected.

Specifically for `TrackStartError` which adheres to reasons (a) and (b) I add a short note for the user to check that no other program is using the camera and refer to grant appropriate permissions by operating system.

This change applies to both `Send` modal and `LoginPrivateKey` which implement `QrCodeScanner`.

ezgif com-video-to-gif

And looks fine in `Send` modal:
screen shot 2018-12-06 at 7 14 58

I tested manually and only on my Mac.
@drptbl would appreciate your help in testing on Windows and Ubuntu as well 🙏🏻

Since I gathered links to the relevant docs for Mac, Win and Lin, so other OS shouldn't render a link but rather plain text. So I created a util `ConditionalLink` which renders an `a` only if the `href` prop has a value. So renders like this:

screen shot 2018-12-06 at 7 16 00

Plz notice that commits from #1745 are included so only review commit 4139dba

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/SendModal/ReadCode/ReadCode.jsx
+++ b/app/components/Modals/SendModal/ReadCode/ReadCode.jsx
@@ -11,7 +11,7 @@ import baseStyles from '../SendModal.scss'
 import styles from './ReadCode.scss'
 
 type Props = {
-  gotoNextStep: Function,
+  gotoNextStep: (content: string, stopScanner: Function) => void,
   cameraAvailable: boolean
 }
 
--- a/app/components/QrCodeScanner/QrCodeScanner.jsx
+++ b/app/components/QrCodeScanner/QrCodeScanner.jsx
@@ -1,12 +1,23 @@
 // @flow
-import React, { Component } from 'react'
+import React, { Component, Fragment } from 'react'
 import Instascan from 'instascan'
 
+import Loading, { ANIMATION_DURATION } from '../../containers/App/Loading'
+
 type Props = {
-  callback: (content: string, scannerInstance: any) => any
+  callback: (content: string, scannerInstance: any) => any,
+  theme: string
 }
 
-export default class QrCodeScanner extends Component<Props> {
+type State = {
+  loading: boolean
+}
+
+export default class QrCodeScanner extends Component<Props, State> {
+  state = {
+    loading: false
+  }
+
   scanPreviewElement: ?HTMLVideoElement
 
   scannerInstance: Instascan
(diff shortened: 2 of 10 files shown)


app/components/Modals/SendModal/ReadCode/ReadCode.jsx
```diff
@@ -11,7 +11,8 @@ import baseStyles from '../SendModal.scss'
 import styles from './ReadCode.scss'
 
 type Props = {
-  gotoNextStep: Function
+  gotoNextStep: Function,
```
dvd-schwrtz: I think types that are more descriptive than Function are more helpful. E.g. (networkId: string) => string

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -29,24 +51,99 @@ export default class QrCodeScanner extends Component<Props> {
       callback(content, this.stopScanner.bind(this))
     })
 
-    Instascan.Camera.getCameras()
+    // since the browser halts while getting usermedia
+    // let the loading animation finish one round and only then continue
+    this.setState({ loading: true })
+    new Promise(resolve => setTimeout(resolve, 900))
```
dvd-schwrtz: Is there a reason we are using 900, maybe put in a variable with a name that is more explanatory.

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -29,24 +51,99 @@ export default class QrCodeScanner extends Component<Props> {
...
+          ],
+          win32: [
+            'https://support.microsoft.com/en-ca/help/10557/windows-10-app-permissions',
+            'Windows Support: App permissions'
+          ],
+          linux: [
+            'https://wiki.ubuntu.com/SecurityPermissions',
+            'Ubuntu Wiki: Security Permissions'
+          ]
+        }[process.platform] || []
```
dvd-schwrtz: where is process defined? perhaps this is a good use case for lodash get

ranbena: `process` is an Electron global variable.

I'll go for lodash.get 👍🏻

ranbena:
@dvd-schwrtz requested changes pushed

comountainclimber:
This is awesome @ranbena! Just some minor feedback 👍

app/components/QrCodeScanner/QrCodeScanner.jsx
```diff
@@ -29,24 +52,99 @@ export default class QrCodeScanner extends Component<Props> {
...
+  static getScannerError(err: Error) {
+    const scanErr: ScannerError = {
+      message: 'Could not connect to camera'
+    }
+
+    if (err.name === 'TrackStartError') {
+      // get link info by user os, defaults to nothing
+      const docs = {
+        darwin: [
+          'https://support.apple.com/en-il/guide/mac-help/mh32356/10.14/mac',
```
comountainclimber: Way to put in the extra effort and track down these support articles... really nice touch 👍

ranbena:
@comountainclimber all requested changes pushed.
Also updated the video/screenshots in the PR desc 🤓

PR #1750, opened 2018-12-12
feature: multiple address support for ledger nano

comountainclimber:
There are no issues open regarding supporting multiple addresses on ledger however it is a feature request I have seen in discord/slack multiple times and pretty neat 🤓 

multiple-ledger-address-support

This PR adds support for generating public keys from a ledger device.

I expanded the API of the existing neonLedger class.

Manually testing of multiple sends and gas claim with a few addresses.

~Those double loading indicators are not my favorite if anyone has any alternative design ideas please comment~
better-loading-indicator https://github.com/CityOfZion/neon-wallet/pull/1750/commits/bc397ef361397c05ec47254b8d5ad85de715ba1b

- [ ] Unit tests written?

### DIFF
--- a/app/actions/authActions.js
+++ b/app/actions/authActions.js
@@ -3,6 +3,7 @@ import { wallet } from 'neon-js'
 import { noop } from 'lodash-es'
 import { createActions } from 'spunky'
 
+import { bindArgsFromN } from '../util/bindHelpers'
 import { resetBalanceState } from './balancesActions'
 import { upgradeNEP6AddAddresses } from '../core/account'
 import { validatePassphraseLength } from '../core/wallet'
@@ -14,7 +15,8 @@ type WifLoginProps = {
 
 type LedgerLoginProps = {
   publicKey: string,
-  signingFunction: Function
+  signingFunction: Function,
+  account: number
 }
 
 type Nep2LoginProps = {
@@ -77,14 +79,18 @@ export const nep2LoginActions = createActions(
 
 export const ledgerLoginActions = createActions(
   ID,
-  ({ publicKey }: LedgerLoginProps) => (): AccountType => {
+  ({ publicKey, account }: LedgerLoginProps) => (): AccountType => {
     const publicKeyEncoded = wallet.getPublicKeyEncoded(publicKey)
-    const account = new wallet.Account(publicKeyEncoded)
+    const walletAccount = new wallet.Account(publicKeyEncoded)
 
     return {
       publicKey,
-      address: account.address,
-      signingFunction: ledgerNanoSCreateSignatureAsync,
+      address: walletAccount.address,
+      signingFunction: bindArgsFromN(
+        ledgerNanoSCreateSignatureAsync,
+        3,
+        account
+      ),
       isHardwareLogin: true
     }
   }
--- a/app/components/Inputs/Label/Label.jsx
+++ b/app/components/Inputs/Label/Label.jsx
@@ -18,7 +18,14 @@ export default class Label extends React.Component<Props> {
   render() {
     const { label, loading, renderAdditionalContent } = this.props
 
-    const passDownProps = omit(this.props, 'label', 'loading')
+    const passDownProps = omit(
+      this.props,
+      'label',
+      'loading',
+      'labelClassName',
+      'labelContainerClassName',
+      'renderAdditionalContent'
+    )
 
     const labelClassName = classNames(styles.label, this.props.labelClassName)
     const labelContainerClassName = classNames(
--- a/app/components/Inputs/Label/Label.jsx
+++ b/app/components/Inputs/Label/Label.jsx
@@ -1,0 +1,14 @@
+// @flow
+import React from 'react'
+import styles from './Label.scss'
+
+type Props = {
+  label: string
+}
+
+export default class Label extends React.Component<Props> {
+  render() {
+    const { label } = this.props
+    return <label className={styles.label}> {label} </label>
+  }
+}
--- a/app/components/Inputs/Label/Label.jsx
+++ b/app/components/Inputs/Label/Label.jsx
@@ -3,25 +3,41 @@ import React from 'react'
 import classNames from 'classnames'
 import { omit } from 'lodash-es'
 
+import RefreshIcon from '../../../assets/icons/refresh.svg'
 import styles from './Label.scss'
 
 type Props = {
   label: string,
-  className: string
+  loading: boolean,
+  labelClassName: string,
+  labelContainerClassName: string,
+  renderAdditionalContent: () => React$Node
 }
 
 export default class Label extends React.Component<Props> {
   render() {
-    const { label } = this.props
+    const { label, loading, renderAdditionalContent } = this.props
 
-    const passDownProps = omit(this.props)
+    const passDownProps = omit(this.props, 'label', 'loading')
 
-    const className = classNames(styles.label, this.props.className)
+    const labelClassName = classNames(styles.label, this.props.labelClassName)
+    const labelContainerClassName = classNames(
+      styles.labelContainer,
+      this.props.labelContainerClassName
+    )
 
     return (
-      <label {...passDownProps} className={className}>
-        {label}
-      </label>
+      <div className={labelContainerClassName}>
+        <label {...passDownProps} className={labelClassName}>
+          {label}
+        </label>
+        {loading && <RefreshIcon className={styles.ledgerStageRefreshIcon} />}
+        {renderAdditionalContent && (
+          <div className={styles.additionalContentContainer}>
+            {renderAdditionalContent()}
+          </div>
+        )}
+      </div>
     )
   }
 }
--- a/app/components/Inputs/Label/Label.jsx
+++ b/app/components/Inputs/Label/Label.jsx
@@ -1,14 +1,27 @@
 // @flow
 import React from 'react'
+import classNames from 'classnames'
+import { omit } from 'lodash-es'
+
 import styles from './Label.scss'
 
 type Props = {
-  label: string
+  label: string,
+  className: string
 }
 
 export default class Label extends React.Component<Props> {
   render() {
     const { label } = this.props
-    return <label className={styles.label}> {label} </label>
+
+    const passDownProps = omit(this.props)
+
+    const className = classNames(styles.label, this.props.className)
+
+    return (
+      <label {...passDownProps} className={className}>
+        {label}
+      </label>
+    )
   }
 }
--- a/app/components/Inputs/Label/Label.jsx
+++ b/app/components/Inputs/Label/Label.jsx
@@ -1,14 +1,13 @@
 // @flow
 import React from 'react'
 import classNames from 'classnames'
-import { omit } from 'lodash-es'
 
-import RefreshIcon from '../../../assets/icons/refresh.svg'
 import styles from './Label.scss'
 
 type Props = {
   label: string,
   loading: boolean,
+  children: React$Node,
   labelClassName: string,
   labelContainerClassName: string,
   renderAdditionalContent: () => React$Node
@@ -16,34 +15,31 @@ type Props = {
 
 export default class Label extends React.Component<Props> {
   render() {
-    const { label, loading, renderAdditionalContent } = this.props
+    const {
+      label,
+      loading,
+      children,
+      labelClassName,
+      labelContainerClassName,
+      renderAdditionalContent,
+      ...passDownProps
+    } = this.props
 
-    const passDownProps = omit(
-      this.props,
-      'label',
-      'loading',
-      'labelClassName',
-      'labelContainerClassName',
-      'renderAdditionalContent'
+    const combinedLabelClassName = classNames(
+      styles.label,
+      this.props.labelClassName
     )
-
-    const labelClassName = classNames(styles.label, this.props.labelClassName)
-    const labelContainerClassName = classNames(
+    const combinedLabelContainerClassName = classNames(
       styles.labelContainer,
       this.props.labelContainerClassName
     )
 
     return (
-      <div className={labelContainerClassName}>
-        <label {...passDownProps} className={labelClassName}>
+      <div className={combinedLabelContainerClassName}>
+        <label {...passDownProps} className={combinedLabelClassName}>
           {label}
         </label>
-        {loading && <RefreshIcon className={styles.ledgerStageRefreshIcon} />}
-        {renderAdditionalContent && (
-          <div className={styles.additionalContentContainer}>
-            {renderAdditionalContent()}
-          </div>
-        )}
+        {children}
       </div>
     )
   }
--- a/app/components/Inputs/Label/Label.scss
+++ b/app/components/Inputs/Label/Label.scss
@@ -1,0 +1,8 @@
+.label {
+  font-weight: bold;
+  font-size: 12px;
+  text-transform: uppercase;
+  opacity: var(--input-label-opacity);
+  color: var(--input-label);
+  margin-bottom: 10px;
+}
--- a/app/components/Inputs/Label/Label.scss
+++ b/app/components/Inputs/Label/Label.scss
@@ -1,8 +1,29 @@
+@import '../../../styles/animations';
+
 .label {
   font-weight: bold;
   font-size: 12px;
   text-transform: uppercase;
   opacity: var(--input-label-opacity);
   color: var(--input-label);
-  margin-bottom: 10px;
+}
+
+.labelContainer {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 5px;
+  min-height: 24px;
+
+  svg {
+    animation: spin 2s linear infinite;
+    margin-left: 5px;
+    path {
+      fill: #5c677f;
+    }
+  }
+}
+
+.additionalContentContainer {
+  margin-left: auto;
 }
--- a/app/components/Inputs/Label/Label.scss
+++ b/app/components/Inputs/Label/Label.scss
@@ -1,5 +1,3 @@
-@import '../../../styles/animations';
-
 .label {
   font-weight: bold;
   font-size: 12px;
@@ -14,16 +12,4 @@
   justify-content: space-between;
   margin-bottom: 5px;
   min-height: 24px;
-
-  svg {
-    animation: spin 2s linear infinite;
-    margin-left: 5px;
-    path {
-      fill: #5c677f;
-    }
-  }
-}
-
-.additionalContentContainer {
-  margin-left: auto;
 }
--- a/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
+++ b/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
@@ -47,10 +47,14 @@ type State = {
   isLoading: boolean,
   addressOption: SelectOption | null,
   publicKeys: Array<LedgerPublicKey>,
-  loadingPublicKeys: boolean
+  loadingPublicKeys: boolean,
+  error: string | null
 }
 
 const POLL_FREQUENCY_MS = 1000
+
+const FETCH_INITIAL_KEYS_ERROR = 'Error fetching public keys.'
+const FETCH_ADDITIONAL_KEYS_ERROR = 'Error fetching additional public keys.'
 
 export default class LoginLedgerNanoS extends React.Component<Props, State> {
   constructor(props: Props) {
@@ -59,7 +63,8 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
       ...this.computeStateFromProps(props),
       addressOption: null,
       publicKeys: [],
-      loadingPublicKeys: true
+      loadingPublicKeys: true,
+      error: null
     }
   }
 
@@ -125,13 +130,11 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
       <div id="loginLedgerNanoS" className={styles.flexContainer}>
         <form>
           {this.renderStatus()}
-          <div className={styles.publicAddressLabelContainer}>
-            <Label label="public address:" />
-            {loadingPublicKeys && (
-              <RefreshIcon className={styles.ledgerStageRefreshIcon} />
-            )}
-          </div>
-
+          <Label
+            loading={loadingPublicKeys}
+            label="public address:"
+            renderAdditionalContent={this.renderAdditionalLabelContent}
+          />
           <StyledReactSelect
             value={this.state.addressOption}
             onChange={addressOption => this.setState({ addressOption })}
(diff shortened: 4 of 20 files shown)


app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+    }))
+  }
+
+  unencodedHexToAddress = (hexString: string) => {
+    const encodedKey = wallet.getPublicKeyEncoded(hexString)
+    return new wallet.Account(encodedKey).address
+  }
+
+  createOptionsFromKeys = () => {
+    const options = this.state.publicKeys.map(key => ({
```
evgenyboxer: I think `key.key` is a bit weird, how about if you call it `publicKey`, since you are mapping `publicKeys`?

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -85,10 +121,26 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
   }
 
   render() {
+    const options = this.createOptionsFromKeys()
+    const { loadingPublicKeys } = this.state
```
evgenyboxer: new line after last const declaration

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
     )
   }
 
+  async fetchInitialKeys() {
+    const publicKeys = await getPublicKeys()
```
evgenyboxer: new line after last const declaration

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+    this.setState({
+      publicKeys,
+      loadingPublicKeys: false,
+      address: {
+        value: publicKeys[0].key,
```
evgenyboxer: would be nicer if the initial public key was a constant

```
  const initialPublicKey = publicKeys[0]
```

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -58,11 +83,22 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+    if (this.props !== LOADED && nextProps.progress === LOADED) {
+      this.fetchInitialKeys()
+    }
   }
 
   computeStateFromProps = (props: Props) => {
     if (props.progress === LOADED) {
-      return { ledgerStage: CONNECTED, isLoading: false }
+      if (this.intervalId) {
+        // $FlowFixMe
```
evgenyboxer: why the flow fix?

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+        label: this.unencodedHexToAddress(publicKeys[0].key)
+      }
+    })
+  }
+
+  async fetchAdditionalKeys() {
+    const { publicKeys } = this.state
+    this.setState({ loadingPublicKeys: true })
+    const lastAccountLoaded = publicKeys[publicKeys.length - 1].account
+    const nextBatchOfKeys = await getPublicKeys(lastAccountLoaded)
```
evgenyboxer: new line after last const

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+      address: {
+        value: publicKeys[0].key,
+        label: this.unencodedHexToAddress(publicKeys[0].key)
+      }
+    })
+  }
+
+  async fetchAdditionalKeys() {
+    const { publicKeys } = this.state
+    this.setState({ loadingPublicKeys: true })
```
evgenyboxer: new line after `this.setState`

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +158,42 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+    const lastAccountLoaded = publicKeys[publicKeys.length - 1].account
+    const nextBatchOfKeys = await getPublicKeys(lastAccountLoaded)
+    this.setState(state => ({
+      publicKeys: [...state.publicKeys, ...nextBatchOfKeys],
+      loadingPublicKeys: false
+    }))
+  }
+
+  unencodedHexToAddress = (hexString: string) => {
+    const encodedKey = wallet.getPublicKeyEncoded(hexString)
```
evgenyboxer: new line after const

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -147,8 +235,15 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
   }
 
   handleLogin = () => {
-    this.props.login(this.props.publicKey)
+    if (this.state.publicKeys.length && this.state.address) {
+      const keyData = this.state.publicKeys.find(
+        key => this.state.address && this.state.address.value === key.key
```
evgenyboxer: same comment as above, perhaps use `{ key }`

app/ledger/neonLedger.js
```diff
@@ -103,6 +103,21 @@ export default class NeonLedger {
     return Promise.resolve()
   }
 
+  async getPublicKeys(
+    acct: number = 0,
+    unencodedPublicKeys: Array<{ account: number, key: string }> = [],
+    batchSize: number = 10
+  ) {
+    const res = await this.send('80040000', BIP44(acct), [VALID_STATUS])
+    const key = await res.toString('hex').substring(0, 130)
```
evgenyboxer: new line after last const declaration

app/actions/authActions.js
```diff
@@ -77,14 +79,18 @@ export const nep2LoginActions = createActions(
...
     const publicKeyEncoded = wallet.getPublicKeyEncoded(publicKey)
-    const account = new wallet.Account(publicKeyEncoded)
+    const walletAccount = new wallet.Account(publicKeyEncoded)
 
     return {
       publicKey,
-      address: account.address,
-      signingFunction: ledgerNanoSCreateSignatureAsync,
+      address: walletAccount.address,
+      signingFunction: bindArgsFromN(
```
evgenyboxer: what are you trying to achieve here? I'm a bit confused by the `bindArgsFromN` function

app/components/Inputs/Label/Label.jsx
```diff
@@ -0,0 +1,14 @@
...
+import styles from './Label.scss'
+
+type Props = {
+  label: string
+}
+
+export default class Label extends React.Component<Props> {
+  render() {
+    const { label } = this.props
+    return <label className={styles.label}> {label} </label>
```
evgenyboxer: seems there are extra spaces `> { label } <`
also, shouldn't this be a functional component?

app/components/Inputs/Label/Label.jsx
```diff
@@ -0,0 +1,14 @@
+// @flow
+import React from 'react'
+import styles from './Label.scss'
+
+type Props = {
+  label: string
```
evgenyboxer: we should all this to also be a react component

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -23,33 +28,53 @@ const { NOT_CONNECTED, OPEN_APP, CONNECTED } = LEDGER_CONNECTION_STAGES
...
 const POLL_FREQUENCY_MS = 1000
 
 export default class LoginLedgerNanoS extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
-
-    this.state = this.computeStateFromProps(props)
+    this.state = {
+      ...this.computeStateFromProps(props),
+      address: null,
```
evgenyboxer: I think this is a bit confusing.. usually, an `address` is a string that represents the public address.. but here its actually an option. How about renaming it to `addressOption`?

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -85,10 +121,26 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+            {loadingPublicKeys && (
+              <RefreshIcon className={styles.ledgerStageRefreshIcon} />
+            )}
+          </div>
+
+          <StyledReactSelect
+            value={this.state.address}
+            onChange={address => this.setState({ address })}
+            options={options}
+            onMenuScrollToBottom={() => this.fetchAdditionalKeys()}
```
evgenyboxer: if `fetchAdditionalKeys` was an arrow function, this we could just write `onMenuScrollToBottom={this.fetchAdditionalKeys}`

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -147,8 +235,15 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
   }
 
   handleLogin = () => {
-    this.props.login(this.props.publicKey)
+    if (this.state.publicKeys.length && this.state.address) {
+      const keyData = this.state.publicKeys.find(
```
evgenyboxer: not sure if you considered this, but would it help if `publicKeys` was a hashmap instead?

app/ledger/neonLedger.js
```diff
@@ -226,7 +244,7 @@ export const getPublicKey = async (acct: number = 0): Promise<string> => {
 export const getDeviceInfo = async () => {
   const ledger = await NeonLedger.init()
   try {
-    return await ledger.getDeviceInfo()
+    return ledger.getDeviceInfo()
```
evgenyboxer: no need for `await` here?

app/actions/authActions.js
comountainclimber: So I probably should have left a comment here but binding the account (3rd argument only) to the signingFunc is the only way that this whole process works with the way things are currently set up.

app/components/Inputs/Label/Label.jsx
comountainclimber: While there are not extra spaces according to our formatting rules i should expand the API of the component to include additional props like you are suggesting thanks @evgenyboxer was an oversight on my part

app/ledger/neonLedger.js
comountainclimber: I cleaned up the other instance of return await (that was in the code base before this PR) as `return await` is basically useless most of the time... If you are already returning a promise there is no point in awaiting its result (unless there is a catch block to catch errors from another Promise). More info can be found here even though we dont have this rule configured currently https://eslint.org/docs/rules/no-return-await

app/actions/authActions.js
comountainclimber: Previously nothing was passed in so the logic always defaulted to `account = 0`

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
comountainclimber: I think using an array is pretty simple and works fine but maybe I am missing something... using a hashmap was not something I considered

comountainclimber:
@evgenyboxer thanks as always for the very thorough review... This branch went thru multiple iterations sorry I pushed it up with a number of issues. Everything should be addressed based on your feedback please take a look and let me know if you have any additional comments or concerns.

app/ledger/neonLedger.js
comountainclimber: Looks like by removing the `return await` from `getPublicKeys` the functionality breaks totally puzzled by that one 😕

app/actions/authActions.js
dvd-schwrtz: the signing function is for the ledger. Basically it never shares the private key. What happens is that when the user is using a hardware device, the transaction is passed to the ledger in a message, when the user clicks the button, it signs the transaction and sends the signed message back to the application, thereby never revealing the private key that was used to sign the transaction. I am not sure why we are binding anything, but just wanted to clarify how that works in case it is helpful. Please let me know if you would like me to look into this code/ binding logic further

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
dvd-schwrtz: I don't think we need a hashmap. there should not be a lot of public keys

app/actions/authActions.js
comountainclimber: @dvd-schwrtz like I mentioned before we have to bind the account otherwise the transaction will always default to account 0 (the first account) => https://github.com/CityOfZion/neon-wallet/blob/ff340cc636eca1165c056cddb86031794a1cd129/app/ledger/neonLedger.js#L264

comountainclimber: I am binding the third argument (account) when the user authenticates to prevent this...

dvd-schwrtz: cool. lets change bindArgsFromN to bindActiveAccount or something

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
ranbena: nit: no need to check that `intervalId` exists since `clearInterval` handles `undefined` value silently.

app/components/Inputs/Label/Label.jsx
ranbena: Perhaps better with text as child instead of string property?
```
return <label className={styles.label}>{this.props.children}</label>
```

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
ranbena: Or perhaps
```
const { key: initialKey } = publicKeys[0]
```

app/ledger/neonLedger.js
ranbena: Perhaps without the `await`, the `finally` executes `ledger.close()` immediately, therefore not allowing `getDeviceInfo()` to happen 🤔

app/components/Inputs/Label/Label.jsx
```diff
@@ -0,0 +1,50 @@
...
+    const { label, loading, renderAdditionalContent } = this.props
+
+    const passDownProps = omit(
+      this.props,
+      'label',
+      'loading',
+      'labelClassName',
+      'labelContainerClassName',
+      'renderAdditionalContent'
+    )
```
ranbena: This `omit` method is used across the app but it not ideal - requires any future changes to props to be mirrored and also perhaps needlessly verbose.

Wdyt of this instead?:
```js
const {
  label,
  loading,
  labelClassName,
  labelContainerClassName,
  renderAdditionalContent,
  ...passDownProps // <-- rest of props
} = this.props

const labelClassName = classNames(styles.label, labelClassName)
const labelContainerClassName = classNames(styles.labelContainer, labelContainerClassName)
```
This way props used in this class will automatically be omitted from the nested components.

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -79,16 +122,31 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
+    const options = this.createOptionsFromKeys()
+    const { loadingPublicKeys, publicKeys } = this.state
+
     return (
       <div id="loginLedgerNanoS" className={styles.flexContainer}>
         <form>
           {this.renderStatus()}
+          <Label
+            label="public address:"
+            renderAdditionalContent={this.renderAdditionalLabelContent}
```
ranbena: I think it'd be better to swap this prop for `children`:
```jsx
<Label label="public address:">
  {this.renderAdditionalLabelContent}
</Label>
```

app/components/Inputs/Label/Label.scss
```diff
@@ -0,0 +1,29 @@
...
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 5px;
+  min-height: 24px;
+
+  svg {
+    animation: spin 2s linear infinite;
+    margin-left: 5px;
+    path {
+      fill: #5c677f;
```
ranbena: Swap with `var(--input-label);`

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -106,6 +164,74 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
...
   }
 
+  fetchAdditionalKeys = async () => {
+    const { publicKeys } = this.state
+
+    this.setState({ loadingPublicKeys: true, error: null })
+
+    const lastAccountLoaded = publicKeys[publicKeys.length - 1].account
+
+    const nextBatchOfKeys = await getPublicKeys(lastAccountLoaded + 1).catch(
```
ranbena: From this I deduce that `account` is an array index. If so, I think it should be indicated in the key name (e.g `publicKeys[0].accountIdx`).

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -23,46 +28,84 @@ const { NOT_CONNECTED, OPEN_APP, CONNECTED } = LEDGER_CONNECTION_STAGES
...
   login: Function,
   connect: Function,
-  publicKey: ?string,
   error: ?string
 }
 
+type SelectOption = {
+  label: string,
+  value: string
+}
```
ranbena: Better this be located in `StyledReactSelect.jsx` (used also in `Settings.jsx`)
And if needed to make more specific (e.g `PriceHistoryPanel` and `ContactsPanel`) we can do this:
```jsx
import type { SelectOption } from '../StyledReactSelect'

type Duration = '1m' | '1w' | '1d'

type DurationSelectOption = {
  ...SelectOption, // <-- copies type definition
  value: Duration // <-- overrides specific type
}
```

Wdyt? This can be a separate PR of course.

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -23,46 +28,84 @@ const { NOT_CONNECTED, OPEN_APP, CONNECTED } = LEDGER_CONNECTION_STAGES
...
+
 type State = {
   ledgerStage: LedgerConnectionStage,
-  isLoading: boolean
+  isLoading: boolean,
```
ranbena: Btw, we can use trailing commas for cleaner diffs.
https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

app/components/Inputs/Label/Label.jsx
comountainclimber: Haha this is a much cleaner solution that then the pattern used throughout the app thanks @ranbena! Will update.

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
comountainclimber: Hmm I see what you are saying... `account` is the var name used throughout the other ledger files - I can update all the other instances if you think its better

ranbena: Leave it be them. Let's not go crazy 🤟🏻

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
comountainclimber: Sure thats a great idea! However instead of exporting it from `StyledReactSelect` I will just add it to the flow-typed file where all of our global types are defined

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
comountainclimber: Sure I am not opposed! In a separate PR I will update our prettier end eslint configurations.

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -23,46 +28,84 @@ const { NOT_CONNECTED, OPEN_APP, CONNECTED } = LEDGER_CONNECTION_STAGES
...
+      ...this.computeStateFromProps(props),
+      addressOption: null,
+      publicKeys: [],
+      loadingPublicKeys: true,
+      error: null
+    }
   }
 
-  intervalId: ?number
+  intervalId: IntervalID
```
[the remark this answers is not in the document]
comountainclimber: 🙌

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
ranbena: awesome

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -147,8 +274,17 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
   }
 
   handleLogin = () => {
-    this.props.login(this.props.publicKey)
+    if (this.state.publicKeys.length && this.state.addressOption) {
+      const keyData = this.state.publicKeys.find(
+        publicKey =>
+          this.state.addressOption &&
```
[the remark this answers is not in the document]
comountainclimber: Flow was throwing an error (even though the check was already being performed on 27) will see if I can come up with a cleaner solution.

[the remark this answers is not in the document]
ranbena: Oh that's cause `this.state` might have changed since the check. Flow is a stickler :)
Just put the addressOption in a new variable and you're good to go:
```js
const { addressOption, publicKeys } = this.state
if (publicKeys.length && addressOption) {
  const keyData = publicKeys.find(
    publicKey => addressOption.value === publicKey.key
  )
  if (keyData) {
    this.props.login(keyData)
  }
}
```

app/actions/authActions.js
comountainclimber: The problem with calling it `bindActiveAccount` is that thats not what its doing... `bindArgsFromN` will bind any amount of function args after a specific argument number (N)

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
comountainclimber: tracking here https://github.com/CityOfZion/neon-wallet/issues/1754

app/components/Inputs/Label/Label.jsx
comountainclimber: I decided to keep the text prop and pass children in for the additional content after reviewing the alternative this seemed to make the most sense and the easiest for styling/layout4

PR #1760, opened 2018-12-20
Perf optimization for QR Scanner (Instascan->jsqr)

ranbena:
This is one of a bunch of commits changing the QR Scan code to be more performant, better and simpler code, with full control of all feature aspects and more visually appealing (in later commits).

We use [Instascan](https://www.npmjs.com/package/instascan) for our qr scan feature but it's 1) bloated 2) abandoned 3) doesn't allow full control.

Implemented our own video preview, image capture and scan logic. The QR code analysis itself is done with [jsqr](https://www.npmjs.com/package/jsqr).

There is no change in functionality in this commit.

Had a hard time making tests for this feature as it requires mocking image data so tested manually.
Check Send Modal and Login QR scanning with faulty qr code and successful qr code.

1. Using a slightly different aspect ratio 16:9 instead of Instascan's 16:10. Think it looks better. Easily configurable.
2. Removed loading animation delay from before. It wasn't the best idea anyhow.
3. Pausing the scan when code found but since there's no feedback from parent components as to the state of the callback, I set it to automatically resume within 4 seconds (if not already unmounted and stopped). This is a temp solution and will be fixed with actual progress event listening in an another PR.

Upcoming commits will add `withProgress` to both Send Modal and Login section and visual effects.

- [ ] Unit tests written?

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "howler": "2.0.15",
     "instascan": "1.0.0",
     "isomorphic-fetch": "2.2.1",
+    "jsqr": "^1.1.1",
     "lodash-es": "4.17.11",
     "moment": "2.22.2",
     "neon-js": "git+https://github.com/cityofzion/neon-js.git#3.11.0",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "howler": "2.0.15",
     "instascan": "1.0.0",
     "isomorphic-fetch": "2.2.1",
-    "jsqr": "^1.1.1",
+    "jsqr": "1.1.1",
     "lodash-es": "4.17.11",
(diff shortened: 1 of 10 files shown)


ranbena:
@dvd-schwrtz https://github.com/CityOfZion/neon-wallet/pull/1745#discussion_r239178419 😁

package.json
```diff
@@ -91,6 +91,7 @@
     "howler": "2.0.15",
     "instascan": "1.0.0",
     "isomorphic-fetch": "2.2.1",
+    "jsqr": "^1.1.1",
```
evgenyboxer: Please use exact version

ranbena: Done 00a3f31

PR #1761, opened 2018-12-20
Implemented redux on ReadCode/SendModal

ranbena:
This is a part of a few PRs to better the QR Scan implementation.

The Send Modal code is a bit convoluted imho and now made simpler.
It handles notification errors by itself and lacks in static typing.

1. Added redux, hoc and spunky to deal with data, notifications and callbacks. This also allows adding `withProgress` to be used by the QR Scanner (upcoming commit depending on this one).
3. Removed the "step 1/2" logic and instead changing steps automatically if recipient data exists or not. So simple!

Manually tested. No change in functionality nor regression.
Go to Send -> Enter QR Code -> try it out

- [ ] Unit tests written?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7229,10 +7229,9 @@ jsprim@^1.2.2:
     json-schema "0.2.3"
     verror "1.10.0"
 
-jsqr@^1.1.1:
+jsqr@1.1.1:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/jsqr/-/jsqr-1.1.1.tgz#a0d7f95e6c3b0bec913dfef2ca64a877f28ed05f"
-  integrity sha512-FVoMU2ncTyjaOqN/vwvDnZ7jaAVvFzM3LK3vG3jvQZFWJQlAwJ1XTCOgAEKo+4Rkd6ydMXTTvqGV/4w5VunmTw==
 
 jsx-ast-utils@^2.0.0, jsx-ast-utils@^2.0.1:
   version "2.0.1"


app/util/parseQRCode.js
```diff
@@ -27,10 +35,13 @@ const parseQRCode = data => {
     if (!asset) throw UNRECOGNIZED_ASSET
   }
 
+  const parsedAmount = searchParams.get('amount')
+  const amount = parsedAmount ? Number(parsedAmount) : null
```
dvd-schwrtz: This could be fine, but might want to use big number here. Not sure exactly the use case and if it is needed. Just something to think about

app/components/Modals/SendModal/SendModal.jsx
```diff
@@ -4,91 +4,43 @@ import React from 'react'
...
-  pushQRCodeData: (data: Object) => any
-}
-
-type State = {
-  step: string,
-  error: ?string,
-  recipientData: Object
+  pushQRCodeData: (data: Object) => any,
+  getRecipientData: string => any,
+  clearRecipientData: Function,
```
dvd-schwrtz: very minor comment to avoid using Function but don't need to change it if the rest of the PR is fine we can merge

app/actions/sendModalActions.js
```diff
@@ -0,0 +1,21 @@
...
+  (url: string) => (): RecipientData => {
+    try {
+      return parseQRCode(url)
+    } catch (msg) {
+      throw new Error(msg)
+    }
+  }
+)
+
+export const clearRecipientData = createActions(ID, () => () => null)
```
ranbena: Is there a better way to do this?
`getRecipientData` sets `recipientData` and `clearRecipientData` empties it. This function `() => () => null` just looks a bit weird and unnecessary.. Is there a `spunky` way to clear the data?

dvd-schwrtz: Not sure. Maybe @evgenyboxer has some time to look at this. Otherwise, I will try and do some research later today or tomorrow.

evgenyboxer: Yeah, I believe this is exactly what `withReset` is for - 

```
withReset(getRecipientData, ((prevProps, nextProps) => {
  // put your reset condition here
})
```

hopefully that helps! (you can put it after your `withData` higher order component in your compose function.

app/components/Modals/SendModal/SendModal.jsx
```diff
@@ -4,91 +4,43 @@ import React from 'react'
...
-    ({
-      '1': <ReadCode gotoNextStep={this.gotoNextStep} />,
-      '2': (
-        <ConfirmDetails
-          recipientData={this.state.recipientData}
-          confirmAndClose={this.confirmAndClose}
-        />
-      )
-    }[this.state.step])
+  getStepComponent = () => {
```
evgenyboxer: @ranbena perhaps in a preference, but this would actually look pretty nice with a class getter

`get stepComponent() { .. }`

then inside the render function:

`this.stepComponent`

WDYT?

app/actions/sendModalActions.js
ranbena: I looked into that before and it didn't make sense to me in this context actually. Lmk if you think otherwise.

app/util/parseQRCode.js
ranbena: You're totally right 😓Returned to original `?string` to avoid any regression risk fab0388

app/components/Modals/SendModal/SendModal.jsx
ranbena: 👍🏻e3ff823

app/components/Modals/SendModal/SendModal.jsx
ranbena: Sure! a217283

PR #1762, opened 2018-12-21
Added visual indications to QR Scan process

ranbena:
Just wanted to make the feature better.

ezgif com-video-to-gif

Added the following indications:
1. Scan in progress - pulsating outline
2. Code found - "camera" flash
3. Waiting for confirmation - freeze frame, loading indicator and solid outline

Tested manually on Send modal and Private Key Login

### Send loop protection
Although waiting for `video.play()` to complete before resuming the scan, the previous frozen frame still manages to get scanned causing a pause-resume loop. Mitigation technique:
```js
lastCodeData: ?string

scan() {
  const code = jsqr(...)

  if (!code) {
    this.lastCodeData = null // <-- reset
    exit
  }

  if (code.data === this.lastCodeData) {
    exit // <-- prevented send loop
  }

  this.lastCodeData = code.data // <-- save for next run
  callback(...)
}
```

### Animation completion
Since callback processes are snappy, there ain't enough time for the "flash" animation (500ms) to complete before the component gets unmounted. It's a shame. So instead of a timeout 👎🏻I implemented a wait for the animation to end:
```js
pause(): Promise<void> {
    // allow "flash" animation (QrCodeScanner.scss) to conclude
    return new Promise(resolve => {
      // $FlowFixMe falsely rejects 'animationend'
      video.addEventListener('animationend', resolve)
      video.pause() // freeze frame and run animations
    })
  }
}
```
Notice the `$FlowFixMe` - `animationend` [isn't recognized](https://gist.github.com/ranbena/f8879c5f1e0048e29e8abca58274f2ad) although it is declared in [flow/dom.js](https://github.com/facebook/flow/blob/master/lib/dom.js#L159) so what gives? Help appreciated 💁‍♂️

I played around with endless versions (canvas draws, svg images, object-fit hacks) and eventually settled on this for its simplicity, efficacy and tastefulness. Any suggestions, including design, colors, and animations are welcome.

- [ ] Unit tests written?


comountainclimber:
Looks amazing and is really fun to use 🔥

PR #1765, opened 2018-12-23
QR Scan - Resume on progress event instead of preset time

ranbena:
This is the last part of the QR Scan fix and refactor commits.

Since introducing scan pause state in #1760, the scanner has no way of knowing if and when to resume, so a constant 4 second wait is invoked after pause. That's suboptimal since callback response duration cannot be predetermined.

A safer method is to regard a callback error as an indication that scanning should be resumed. 
So I added `withProgress()` to the relevant components (thence #1761) and the scanner now acts upon progress changed from `LOADING` to `FAILED`.

Note: Listening to callback success events is irrelevant since the scanner component is abandoned thereafter.

Manually tested.

Put in a 1.5 second resume delay just for better user experience, so it doesn't resume before the callback error notification prompts.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Modals/SendModal/ReadCode/ReadCode.jsx
+++ b/app/components/Modals/SendModal/ReadCode/ReadCode.jsx
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react'
 import classNames from 'classnames'
+import { type ProgressState } from 'spunky'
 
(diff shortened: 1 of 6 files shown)


comountainclimber:
Looks good to me. Nice @ranbena 💪

Issue #1741, opened 2018-12-03
NEON should broadcast transactions to multiple nodes


ThomasLobker:
When you transfer NEP-5 tokens, NEON will create a transaction and broadcast it to an RPC node. The node should add the transaction to the mempool and broadcast it to the rest of the network so it can eventually be added to the chain. When this is done, you will get the transaction hash from NEON and you have to assume that the RPC node is actually broadcasting the transaction throughout the NEO network.

The problem is, you cannot trust the RPC nodes to do this. So there is a chance that you will send a transaction that will never be committed, or even worse, that the transaction will be committed much later than anticipated. If this is the case, then the user might have tried to sent the tokens in a new transaction already causing a double-spend scenario.

I would like to propose a couple of improvements to help in these situations.

1. Broadcast the transactions to multiple nodes at once. This should not cause any harm, because transactions can be committed only once to the chain anyway. But the risk that transactions will get lost will be lowered significantly.
2. Keep sent transactions in an internal transaction log and keep trying to broadcast them until they are confirmed, if new blocks keep coming in but the transaction is not getting confirmed.
3. When creating a new transaction, display the hash before broadcasting. This should help the user to see that he is creating a new transaction, and not retrying the same old transaction. Maybe we should even let the user choose the randomness that is added to each transaction.

This applies on NEP-5 transfers only since they are not UTXO based.

comountainclimber:
Please award @ranbena 100 and @evgenyboxer 100 for the code review here https://github.com/CityOfZion/neon-wallet/pull/1781 and here https://github.com/CityOfZion/neon-wallet/pull/1766

lllwvlvwlll:
https://neoscan.io/transaction/d24882f33f56a405c55ec93ea0306423ab06385e3d4937b735c06ea984d67b39

PR #1766, opened 2018-12-24
Feature: send UX enhancements (ready for review)

comountainclimber:
send-enhancements-3

TODO:
- ~Fix bug preventing correct hash from being generated during contract invocation~

- Fix error messaging in Send timeout to direct users to take a look at activity **(separate PR)**
- Add number of pending tx on top of Activity logo based on spunky state **(separate PR)**

Solves at least part of https://github.com/CityOfZion/neon-wallet/issues/1741 and a lot of problems that arise during periods of network instability ie what **exactly** is the status of my transaction

This creates an internal storage log of "pending" transactions that the user has attempted to broadcast to the network. We generate a transaction ID locally before broadcasting to a node and then use this ID to poll the blockchain for confirmations/transaction status. Once the transaction reaches ~40~ 10 confirmations OR the neoscan API responds with this transaction id we purge the id from local storage.

By generating the transaction locally _before_ making the request to the node and then storing the determined tx id in local storage. UPDATE: It is not currently possible (that I know of) to take the an InvocationTransaction script and reverse engineer it to calculate the exact outputs as an alternative I decided to pass around the send entries array and conditionally use its value instead in the event of the transaction being an invocation.

Manual testing - incoming unit and integration tests. **(separate PR)**

- Adds images to the Activity page 🎉 
- Updates neon-js to `3.11.9`  and installs via npm instead of github
- Sets timeout of RPC requests to `60000`MS  instead of the default `30000`

~**NOTE:** for some reason the hashes I am generating for tokens (not GAS or NEO) are not creating the correct hash and I am actively investigating~

- [ ] Unit tests written?

### DIFF
--- a/app/actions/pendingTransactionActions.js
+++ b/app/actions/pendingTransactionActions.js
@@ -191,7 +191,7 @@ export const getPendingTransactionInfo = createActions(
             })
 
           if (result) {
-            if (result.confirmations > MINIMUM_CONFIRMATIONS) {
+            if (result.confirmations >= MINIMUM_CONFIRMATIONS) {
               // eslint-disable-next-line
               await pruneConfirmedOrStaleTransaction(address, transaction.hash)
             } else {
--- a/app/actions/pendingTransactionActions.js
+++ b/app/actions/pendingTransactionActions.js
@@ -3,24 +3,30 @@ import { createActions } from 'spunky'
 import Neon from '@cityofzion/neon-js'
 import { isEmpty } from 'lodash-es'
 
+import { toBigNumber } from '../core/math'
 import { getStorage, setStorage } from '../core/storage'
 import { getNode, getRPCEndpoint } from './nodeStorageActions'
-import { findAndReturnTokenInfo } from '../util/findAndReturnTokenInfo'
+import {
+  findAndReturnTokenInfo,
+  getImageBySymbol,
+} from '../util/findAndReturnTokenInfo'
 
 export const ID = 'pendingTransactions'
 const STORAGE_KEY = 'pendingTransactions'
-const MINIMUM_CONFIRMATIONS = 40
+const MINIMUM_CONFIRMATIONS = 10
 
 type PendingTransactions = {
-  [address: string]: Array<string>,
+  [address: string]: Array<any>,
 }
 
 type PendingTransaction = {
   vout: Array<{ asset: string, address: string, value: string }>,
+  sendEntries: Array<SendEntryType>,
   confirmations: number,
   txid: number,
   net_fee: string,
   blocktime: number,
+  type: string,
 }
 
 type ParsedPendingTransaction = {
@@ -30,10 +36,66 @@ type ParsedPendingTransaction = {
   blocktime: number,
   to: string,
   amount: string,
-  asset: string,
+  asset: {
+    symbol: string,
+    image: string,
+  },
 }
 
-export const parsePendingTxInfo = async (
+export const parseContractTransaction = async (
+  transaction: PendingTransaction,
+  net: string,
+): Promise<Array<ParsedPendingTransaction>> => {
+  const parsedData = []
+  // eslint-disable-next-line camelcase
+  const { confirmations, txid, net_fee, blocktime = 0 } = transaction
+  transaction.vout.pop()
+  // eslint-disable-next-line
+  for (const send of transaction.vout) {
+    parsedData.push({
+      confirmations,
+      txid,
+      net_fee,
+      blocktime,
+      amount: toBigNumber(send.value).toString(),
+      to: send.address,
+      // eslint-disable-next-line no-await-in-loop
+      asset: await findAndReturnTokenInfo(send.asset, net),
+    })
+  }
+  return parsedData
+}
+
+export const parseInvocationTransaction = (
+  transaction: PendingTransaction,
+): Array<ParsedPendingTransaction> => {
+  const {
+    confirmations,
+    txid,
+    // eslint-disable-next-line camelcase
+    net_fee,
+    blocktime = 0,
+    sendEntries,
+  } = transaction
+
+  // things get tricky during invocation transactions as there is no vout array
+  // and it is not straight forward parsing the produced script. Instead we
+  // use the original send entries array.
+  return sendEntries.map(send => ({
+    confirmations,
+    txid,
+    net_fee,
+    blocktime,
+    amount: toBigNumber(send.amount).toString(),
+    to: send.address,
+    asset: {
+      symbol: send.symbol,
+      image: getImageBySymbol(send.symbol),
+    },
+  }))
+}
+
+export const parsePendingContractTxInfo = async (
   pendingTransactionsInfo: Array<PendingTransaction>,
   net: string,
 ) => {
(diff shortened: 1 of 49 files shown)


app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
```diff
@@ -24,11 +30,39 @@ export default class TransactionHistory extends React.Component<Props> {
...
+      pendingTransactions,
+      address,
+      net,
+      getPendingTransactionInfo,
+    } = this.props
+    if (
+      transactions &&
+      transactions.length &&
+      pendingTransactions &&
+      pendingTransactions.length
```
ranbena: nit: Not sure this check is actually needed.
`transactions` and `pendingTransactions` will always be arrays (or Flow would have raised an error) and if empty the following code would still work fine.

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
```diff
@@ -24,11 +30,39 @@ export default class TransactionHistory extends React.Component<Props> {
...
+    ) {
+      const confirmed = transactions.map(tx => tx.txid)
+      // NOTE: removes the '0x' prepended to every txId
+      const pending = pendingTransactions.map(tx => tx.txid.substring(2))
+      const toBePurged = intersection(confirmed, pending)
```
ranbena: nit: if `pendingTransactions.txId` would have been normalized beforehand (is there any reason not to?) you could do this in one line:
```
const toBePurged = intersectionBy(transactions, pendingTransactions, 'txId')
```

app/components/TransactionHistory/TransactionHistoryPanel/Transactions.jsx
```diff
@@ -11,13 +11,20 @@ import styles from './Transactions.scss'
...
+  static defaultProps = {
+    pendingTransactions: [],
+  }
 
-    if (transactions.length === 0) {
+  render() {
+    const { className, transactions, pendingTransactions } = this.props
+    if (
+      !transactions.length === 0 &&
+      (!pendingTransactions || !pendingTransactions.length)
```
ranbena: Lodash nit:
```
if (!isEmpty(transactions) && isEmpty(pendingTransactions)) {
```

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
```diff
@@ -24,11 +30,39 @@ export default class TransactionHistory extends React.Component<Props> {
...
+      // eslint-disable-next-line
+      for (const id of toBePurged) {
+        // eslint-disable-next-line
+        await pruneConfirmedOrStaleTransaction(address, id)
+        getPendingTransactionInfo({ address, net })
```
ranbena: I haven't dug into the code but something here feels like a perf bottleneck.
1. Multiple sequential calls to `getPendingTransactions` and `getPendingTransactionInfo`.
2. The `await` makes these run in sequence . Can instead run in parallel?

ranbena:
@comountainclimber I have a few preliminary questions and suggestions:

1. UX - After confirming a send transaction, the user is prompted "TRANSFER COMPLETED" although it's actually pending. Perhaps update this?

2. How do you determine a stale transaction?

3. Possible to merge `<PendingTransaction>` into `<Transaction>` instead of being entirely separate html implementations? This btw is one of TypeScript's strong features - [implements interface](https://dzone.com/articles/how-to-define-an-interface-in-typescript)

4. Would be great to disconnect these into separate "prep" PRs:
    - util/findAndReturnTokenInfo
    - Add asset image
    - @cityofzion/neon-js

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
comountainclimber: So I am running these in sequence on purpose in case one of the transaction ids stored locally invalid this allows me to sequentially filter them out. Maybe there is a way to do this in parallel that I am not thinking of?

comountainclimber:
Thanks for taking a look @ranbena in regards to your points:
1.) I want to update all the error messaging in a separate PR as to to not bloat this one any further
2.) A transaction is currently "stale" if the RPC server responds with an error otherwise it was never successfully posted - perhaps this logic needs to be updated 🤷‍♂️ 
3.) I like this plan and will go ahead and implement
4.) I think at this point it might be more trouble than its worth breaking this one up any further but will go ahead and add any additional feature work to separate PRs:
- send messaging enhancements / error messaging improvements
- UI enhancements to activity page
- Unit and integration tests

dvd-schwrtz:
I know you are still working on this but the code looks good so far

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
ranbena: Understood. I see also that it updates a single localStorage key which should not be updated in parallel. I'll let you know if I find anything in particular.

app/actions/pendingTransactionActions.js
```diff
@@ -0,0 +1,150 @@
...
+                'An transaction was added to storage that the blockchain does not recognize - purging from storage',
+              )
+              await pruneConfirmedOrStaleTransaction(address, transaction)
+            })
+
+          if (result.confirmations > MINIMUM_CONFIRMATIONS) {
+            // eslint-disable-next-line
+            await pruneConfirmedOrStaleTransaction(address, transaction)
+          }
+          pendingTransactionInfo.push(result)
```
ranbena: Shouldn't already pruned (stale/confirmed) transactions be excluded from this array?

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
ranbena: The part I'm having trouble understanding is the execution of `getPendingTransactionInfo` here. It doesn't expect the return value and it's costly (creates RPC client, storage access, tx parsing, etc.). Can you explain?

app/actions/pendingTransactionActions.js
comountainclimber: Yes thanks!

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
comountainclimber: This is an excellent point (I just addressed this issue in my most recent commit) basically the intention was to update pendingTransactions via dispatching an action... Like you said its much more performant to just synchronously filter out the confirmed transactions (abstracts returned from neoscan API) and not dispatch the action 👍

app/components/TransactionHistory/TransactionHistoryPanel/TransactionHistoryPanel.jsx
comountainclimber: I like this idea and implemented it however it caused me to change the filter logic here:

```
export const pruneConfirmedOrStaleTransaction = async (
  address: string,
  txId: string,
) => {
  const storage = await getPendingTransactions()
  if (Array.isArray(storage[address])) {
    storage[address] = storage[address].filter(
      // use includes here to be indifferent to 0x prefix
      transaction => !transaction.hash.includes(txId),
    )
  }
  await setPendingTransactions(storage)
}
```

app/components/TransactionHistory/TransactionHistoryPanel/Transactions.jsx
```diff
@@ -21,10 +22,7 @@ export default class Transactions extends React.Component<Props> {
 
   render() {
     const { className, transactions, pendingTransactions } = this.props
-    if (
-      !transactions.length === 0 &&
-      (!pendingTransactions || !pendingTransactions.length)
-    ) {
+    if (isEmpty(transactions) && isEmpty(pendingTransactions)) {
```
ranbena: I think `!transactions.length === 0` should be `!isEmpty(transactions)`

ranbena:
Super feature @comountainclimber! Mazal Tov 🍾

app/components/TransactionHistory/TransactionHistoryPanel/Transactions.jsx
comountainclimber: `if (isEmpty(transactions) && isEmpty(pendingTransactions))` should work

comountainclimber:
Thanks for the review @ranbena and @dvd-schwrtz 👍

PR #1768, opened 2018-12-30
chore(app): fixed some linting issues and typos

DalderupMaurice:
/

fixed some linting issues and typos

by fixed some linting issues and typos

running it

/

/

- [ ] Unit tests written?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -34,6 +34,7 @@
     "no-unused-vars": ["warn"],
     "global-require": 0,
     "quotes": ["warn", "single"],
+    "no-console": ["error", { "allow": ["warn", "error"] }],
     "object-curly-spacing": ["error", "always"],
     "import/no-unresolved": 0,
     "import/prefer-default-export": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,7 +41,7 @@
     "import/no-dynamic-require": 0,
     "semi": [2, "never"],
     "max-len": 0,
-    "consistent-return": ["warn"],
+    "consistent-return": ["warn", { "treatUndefinedAsUnspecified": true }],
     "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
     "class-methods-use-this": ["warn"],
     "jsx-a11y/anchor-is-valid": 0,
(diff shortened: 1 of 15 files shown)


app/actions/nodeStorageActions.js
```diff
@@ -110,7 +110,9 @@ export const getRPCEndpoint = async (
 }
 
 export const getNode = async (net: Net): Promise<string> => {
-  const storage = await getStorage(`${STORAGE_KEY}-${net}`).catch(console.error)
+  const storage = await getStorage(`${STORAGE_KEY}-${net}`).catch(e =>
+    console.error(e),
+  )
```
comountainclimber: Fine with this change but if Im not mistaken its the same as what was there previously

.eslintrc
```diff
@@ -34,13 +34,14 @@
     "no-unused-vars": ["warn"],
     "global-require": 0,
     "quotes": ["warn", "single"],
+    "no-console": ["error", { "allow": ["warn", "error"] }],
```
comountainclimber: Awesome 👍

comountainclimber:
Thanks for this clean up PR looks great 💪- had to lookup what `treatUndefinedAsUnspecified` meant https://eslint.org/docs/rules/consistent-return#treatundefinedasunspecified seems like we do want that option enabled so thanks for adding that

app/modules/sale.js
```diff
@@ -77,7 +77,7 @@ export const participateInSale = (
   const scriptHashAddress = wallet.getAddressFromScriptHash(formattedScriptHash)
 
   const intents = [[ASSETS.NEO, neoToMint], [ASSETS.GAS, gasToMint]]
-    .filter(([symbol, amount]) => amount > 0)
+    .filter(([symbol, amount]) => amount > 0) // eslint-disable-line
```
comountainclimber: Why was this added here? Can we specify what rule is actually being disabled

DalderupMaurice: `symbol` was not used, the rule is no-unused-vars, but it would probably change the logic when you leave it out

app/actions/nodeStorageActions.js
DalderupMaurice: Oh sorry, I wasn't 100% sure 😄

app/modules/sale.js
DalderupMaurice: it actually did

PR #1769, opened 2018-12-31
feature: UI enhancements for unconfirmed transactions flow

comountainclimber:
This PR is part II of https://github.com/CityOfZion/neon-wallet/pull/1766 and provides some subtle updates to the UI to help users understand the exact state of their unconfirmed transactions.

send-enhancements-4

screen shot 2018-12-31 at 9 22 19 am
screen shot 2018-12-31 at 9 22 47 am

Further improves our send UX

By extending the work started in https://github.com/CityOfZion/neon-wallet/pull/1766

Manual testing locally

I updated `addPendingTransaction` to always return the retrieved information from the blockchain when adding a tx to storage so that the SideBar can know about only transactions returned from `getRawTransaction`  as opposed to every transaction that gets added to storage.

- [ ] Unit tests written?

### DIFF
--- a/app/actions/pendingTransactionActions.js
+++ b/app/actions/pendingTransactionActions.js
@@ -16,33 +16,6 @@ const STORAGE_KEY = 'pendingTransactions'
 const MINIMUM_CONFIRMATIONS = 10
 const INVALID_TX_ERROR_MESSAGE = 'Unknown transaction'
 
-type PendingTransactions = {
-  [address: string]: Array<any>,
-}
-
-type PendingTransaction = {
-  vout: Array<{ asset: string, address: string, value: string }>,
-  sendEntries: Array<SendEntryType>,
-  confirmations: number,
-  txid: string,
-  net_fee: string,
-  blocktime: number,
-  type: string,
-}
-
-type ParsedPendingTransaction = {
-  confirmations: number,
-  txid: string,
-  net_fee: string,
-  blocktime: number,
-  to: string,
-  amount: string,
-  asset: {
-    symbol: string,
-    image: string,
-  },
-}
-
 export const parseContractTransaction = async (
   transaction: PendingTransaction,
   net: string,
--- a/app/actions/pendingTransactionActions.js
+++ b/app/actions/pendingTransactionActions.js
@@ -95,7 +95,7 @@ export const parseInvocationTransaction = (
   }))
 }
 
-export const parsePendingContractTxInfo = async (
+export const parseTransactionInfo = async (
   pendingTransactionsInfo: Array<PendingTransaction>,
   net: string,
 ) => {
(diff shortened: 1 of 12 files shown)


app/actions/pendingTransactionActions.js
```diff
@@ -136,15 +136,72 @@ export const pruneConfirmedOrStaleTransaction = async (
...
   }
   await setPendingTransactions(storage)
 }
 
+export const fetchTransactionInfo = async (
+  transactions: PendingTransactions,
+  address: string,
+  net: string,
+) => {
+  if (Array.isArray(transactions[address]) && transactions[address].length) {
```
dvd-schwrtz: is it ever possible that pending transactions is undefined instead of an empty object? this would cause a problem

app/components/Send/SendPanel/ActivityLink/ActivityLink.jsx
```diff
@@ -0,0 +1,23 @@
...
+import { ROUTES } from '../../../../core/constants'
+import styles from './ActivityLink.scss'
+
+type Props = {
+  error: boolean,
+}
+
+export const ActivityLink = ({ error }: Props) => (
+  <div>
+    {error
```
dvd-schwrtz: error && 'your sentence' vs error ? 'your sentence' : ' ', unless you need that space if there is no error

app/containers/App/Sidebar/index.js
```diff
@@ -1,9 +1,19 @@
 // @flow
 import { compose } from 'recompose'
+import { withData } from 'spunky'
 import { withRouter } from 'react-router-dom'
 
 import Sidebar from './Sidebar'
+import { addPendingTransaction } from '../../../actions/pendingTransactionActions'
+
+const mapPendingTransactionsDataToProps = (
+  pendingTransactions: Array<any>,
```
evgenyboxer: shouldn't the type be - `PendingTransactions`?

app/containers/App/Sidebar/index.js
```diff
@@ -1,9 +1,19 @@
...
+) => ({
+  pendingTransactionsCount:
+    (Array.isArray(pendingTransactions) && pendingTransactions.length) || 0,
```
evgenyboxer: perhaps ? `get(pendingTransactions, 'length', 0)`

app/containers/App/Sidebar/Sidebar.jsx
```diff
@@ -53,6 +54,11 @@ const Sidebar = ({ className, theme }: Props) => {
           className={styles.navItem}
           activeClassName={styles.active}
         >
+          {!!pendingTransactionsCount && (
```
evgenyboxer: I actually think `pendingTransactionsCount > 0` reads a bit better.. but perhaps this just a personal preference.

app/actions/pendingTransactionActions.js
comountainclimber: 👍 by default electron JSON storage will always return an empty object for the key value pairing but I will go ahead and add a check in here anyway

Issue #1704, opened 2018-11-17
Cannot copy address for my wallet address on Activity page


osmirnov:
**Wallet version or commit SHA:**
2.0.0 (800d415)

**Environment:**
e.g.: Win 10 Pro (64 bit) Version 10.0.17134 Build 17134

**Prerequisites:**
N/A

**Reproduction steps:**
1. Start Neon Wallet
2. Login
3. Make transaction on your own wallet address
4. Go on Activity page  -> "Copy address" button is unavailable

Expectation: "Copy address" should be available because on "Contact" page we can create a contact for our wallet address. Without an assigned contact the button is shown.

**Screenshots/Video:**
![image](https://user-images.githubusercontent.com/454196/48663934-40873780-ea65-11e8-98b4-769749b8bca6.png)

comountainclimber:
This was by design as this address has already been added to the contacts section... We can alter that to allow for copying anyway if you think thats better 👍

osmirnov:
Hidding "Add Contact" makes sense for because the address already assigned to the contact. However, hiding "Copy Address" means I need to go to "Contacts" to get it from there. I am against extra clicking in general 😉

comountainclimber:
Confirmed this is no longer an issue

PR #1770, opened 2019-01-06
Chore/fix copy address

phetter:
[Cannot copy address for my wallet address on Activity page #1704](https://github.com/CityOfZion/neon-wallet/issues/1704)

Cannot copy address for my wallet address on Activity page

Removed conditional checks for contact existence to always show the copy address button

I tested each transaction address was correctly copied upon pressing the copy address button

No 

No

- [ ] Unit tests written?
No


comountainclimber:
@phetter thank you for your contribution 👍 the work you did in this PR is noteworthy and was taken into consideration when determining points allocated for your other work during this time period

PR #1771, opened 2019-01-06
feature: fixes minor security issue and updates webpack

comountainclimber:
In an effort to resolve the minor security issue identified by GH related to `webpack-dev-server` I ended up trying to upgrade to webpack 4 which seems to be running totally fine both with our dev and prod bundles 🎉 also tested and noticed that HMR is substantially less buggy with the set of updated deps.

screen shot 2019-01-06 at 9 58 57 am

- [ ] Unit tests written?


package.json
```diff
@@ -174,9 +174,9 @@
     "eslint-plugin-prettier": "2.6.1",
     "eslint-plugin-promise": "3.6.0",
     "eslint-plugin-react": "7.9.1",
-    "extract-text-webpack-plugin": "3.0.2",
+    "extract-text-webpack-plugin": "4.0.0-beta.0",
     "flow-bin": "0.80.0",
-    "html-webpack-plugin": "2.30.1",
+    "html-webpack-plugin": "^3.2.0",
```
evgenyboxer: please use exact versioning

dvd-schwrtz: when was the last time we went through and updated the modules?

comountainclimber: Depends on which dependencies you are referring to but this PR represents an upgrade of all our webpack related modules to the latest and greatest...

Issue #1703, opened 2018-11-17
Cannot add contact for my wallet address on Activity page


osmirnov:
**Wallet version or commit SHA:**
2.0.0 (800d415)

**Environment:**
e.g.: Win 10 Pro (64 bit) Version 10.0.17134 Build 17134

**Prerequisites:**
N/A

**Reproduction steps:**
1. Start Neon Wallet
2. Login
3. Make transaction on your own wallet address
4. Go on Activity page  -> "Add Contact" button is unavailable

Expectation: "Add Contact" should be available because on "Contact" page we can create a contact for our wallet address.

**Screenshots/Video:**
![image](https://user-images.githubusercontent.com/454196/48663902-c5258600-ea64-11e8-9b55-9d6c48661979.png)

comountainclimber:
Please award @phetter 100 🚀

lllwvlvwlll:
https://neoscan.io/transaction/6b84fb0383d72a0c6d71766c96ecf8591fa7f7a24914f0df6b86f473d2db8d3b

PR #1772, opened 2019-01-06
chore/add-contact-activity-button

phetter:
[Cannot add contact for my wallet address on Activity page #1703](https://github.com/CityOfZion/neon-wallet/issues/1703)

One can now add their own address to contacts.

Code was added to all transaction types that conditionally shows an add contact button if the contact does not already exist in the contacts.

I tested the buttons for on the activity view.

No

No
- [ ] Unit tests written?
No

### DIFF
--- a/app/components/Blockchain/Transaction/ClaimAbstract.jsx
+++ b/app/components/Blockchain/Transaction/ClaimAbstract.jsx
@@ -2,8 +2,10 @@
 import React, { Fragment } from 'react'
 
 import classNames from 'classnames'
+import Button from '../../Button'
 import styles from './Transaction.scss'
 import ClaimIcon from '../../../assets/icons/claim.svg'
+import ContactsAdd from '../../../assets/icons/contacts-add.svg'
 import CopyToClipboard from '../../CopyToClipboard'
 
 type Props = {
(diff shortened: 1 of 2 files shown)


app/components/Blockchain/Transaction/ClaimAbstract.jsx
```diff
@@ -53,7 +57,16 @@ export default class ClaimAbstract extends React.Component<Props> {
               }
             </Fragment>
           </div>
-          <div className={styles.historyButtonPlaceholder} />
+          {
```
evgenyboxer: I don't think there is a need to wrap the `Button` with the curly braces

app/components/Blockchain/Transaction/ReceiveAbstract.jsx
```diff
@@ -60,9 +60,7 @@ export default class ReceiveAbstract extends React.Component<Props> {
               />
             )}
           </div>
-          {isMintTokens || isGasClaim ? (
-            <div className={styles.transactionHistoryButton} />
-          ) : (
+          {
```
evgenyboxer: same comment as above ^

app/components/Blockchain/Transaction/ClaimAbstract.jsx
phetter: Whoops, thank you!

Issue #1677, opened 2018-11-12
Wallet crashes on dashboard if one of cryptocompare `histohour` responses fail to respond


drptbl:
**Wallet version or commit SHA:**
`dev` branch (ee7f579)

**Environment:**
Win 10 Home (64 bit) Version 1803 Build 17134.376

**Reproduction steps:**
1. Start wallet
2. Set intercepting on `https://min-api.cryptocompare.com/data/histohour?limit=24&fsym=GAS&tsym=USD` or `https://min-api.cryptocompare.com/data/histohour?limit=24&fsym=NEO&tsym=USD`
3. Login to the wallet
4. Pass all requests excluding the one mentioned above

**Description:**
If one of these two requests fail to respond (or both of them):
`https://min-api.cryptocompare.com/data/histohour?limit=24&fsym=GAS&tsym=USD`
`https://min-api.cryptocompare.com/data/histohour?limit=24&fsym=NEO&tsym=USD`
Wallet is crashing on dashboard. This means that currently we're fully relying on cryptocompare API in the wallet (by this I mean that if this endpoint goes down, wallet will crash on login).

**Video:**
![neon_crash](https://user-images.githubusercontent.com/8177587/48353530-f16d8b00-e68f-11e8-8131-e6f0d31f2be2.gif)

There are 3 different console errors in 3 cases:
1. Both of requests fail to respond
2. Request for `histohour` GAS fail to respond
3. Request for `histohour` NEO fail to respond

comountainclimber:
I am unable to reproduce this using the exact steps you outlined on macOS

comountainclimber:
![attempt-to-repro-on-osx](https://user-images.githubusercontent.com/13072035/48359665-abe29b00-e65a-11e8-9b10-a5006ab5db29.gif)

comountainclimber:
Fixed in 2.1.0

PR #1780, opened 2019-01-10
feature: functionality without historical pricing from cryptocompare

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1677

Because of the way the data is being accessed in neon-wallet the app will crash if the data structure does not conform to what the app was expecting. This has been on my TODO for some time and unfortunately it is biting us in production right now.

By conditionally rendering components and using lodash get

By testing locally

screen shot 2019-01-09 at 6 28 28 pm

(also the percentage changes have been removed in the asset panel)

- [ ] Unit tests written?


evgenyboxer:
@comountainclimber this PR seems identical to https://github.com/CityOfZion/neon-wallet/pull/1781 ?

comountainclimber:
Its branched off of `master` instead of `dev` and is identical

PR #1781, opened 2019-01-10
hotfix: fixes bug causing app to crash due to missing historical price info

comountainclimber:
https://github.com/CityOfZion/neon-wallet/pull/1780

- [ ] Unit tests written?

### DIFF
--- a/app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
+++ b/app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
@@ -61,15 +61,17 @@ export default class AssetBalancesPanel extends React.Component<Props> {
               <span className={styles.value} id="neoWalletValue">
                 {this.getFormattedFiatBalance(this.getNEOValue())}
               </span>
-              <span
-                className={classNames(
-                  styles.change,
-                  styles[this.getNEOPriceChangeDirection()]
-                )}
-                id="priceChangeNeo"
-              >
-                {this.getNEOFormattedPriceChange()}
-              </span>
+              {!Number.isNaN(this.props.neoPriceChange.toNumber()) && (
+                <span
+                  className={classNames(
+                    styles.change,
+                    styles[this.getNEOPriceChangeDirection()]
+                  )}
+                  id="priceChangeNeo"
+                >
+                  {this.getNEOFormattedPriceChange()}
+                </span>
+              )}
             </div>
             <div className={styles.asset}>
               <div className={classNames(styles.label, styles.assetName)}>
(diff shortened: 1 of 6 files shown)


app/components/Dashboard/PriceHistoryPanel/PriceHistoryChart.jsx
```diff
@@ -103,7 +105,7 @@ export default class PriceHistoryChart extends React.Component<Props, State> {
   }
 
   tooltipContent = (tooltipProps: Object) =>
-    !!tooltipProps.payload.length && (
+    !!get(tooltipProps, '.payload.length') && (
```
evgenyboxer: I think it should be `get(tooltipProps, 'payload.length')` <- without the leading '.' before `payload`

app/components/Dashboard/PriceHistoryPanel/PriceHistoryChart.jsx
```diff
@@ -55,7 +55,9 @@ export default class PriceHistoryChart extends React.Component<Props, State> {
   render = (): React$Node => {
     const { className, prices, timeKey, priceKey } = this.props
 
-    return (
+    return isEmpty(prices) ? (
```
evgenyboxer: Seems like prices is guaranteed to be an array right? If so, its better to just check the length because `isEmpty` will return true if `prices` is undefined/null.

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -61,15 +61,17 @@ export default class AssetBalancesPanel extends React.Component<Props> {
...
-              <span
-                className={classNames(
-                  styles.change,
-                  styles[this.getNEOPriceChangeDirection()]
-                )}
-                id="priceChangeNeo"
-              >
-                {this.getNEOFormattedPriceChange()}
-              </span>
+              {!Number.isNaN(this.props.neoPriceChange.toNumber()) && (
```
evgenyboxer: How about `!this.props.neoPriceChange.isNaN()`

app/components/Dashboard/AssetBalancesPanel/AssetBalancesPanel.jsx
```diff
@@ -86,15 +88,17 @@ export default class AssetBalancesPanel extends React.Component<Props> {
...
-              <span
-                className={classNames(
-                  styles.change,
-                  styles[this.getGASPriceChangeDirection()]
-                )}
-                id="priceChangeGas"
-              >
-                {this.getGASFormattedPriceChange()}
-              </span>
+              {!Number.isNaN(this.props.gasPriceChange.toNumber()) && (
```
evgenyboxer: and `!this.props.gasPriceChange.isNaN()`

app/components/Dashboard/PriceHistoryPanel/PriceHistoryChart.jsx
comountainclimber: Prices is not an array - currently the response from cryptocompare is malformed and prices is an object...

comountainclimber: I updated `index.js` so that if pricing data is not an array it will fallback to an empty array and I changed this check to look for length like you suggested

Issue #1729, opened 2018-11-27
Infinite loader on MainNet node selection page


drptbl:
**Wallet version or commit SHA:**
`dev` branch (c92b866)

**Environment:**
MacBook Pro (13-inch, 2017), macOS Mojave 10.14.1

**Reproduction steps:**
1. Start Neon Wallet
2. Login to any wallet
3. Click on "Settings" button in the left menu
4. Click on "Node selection" position in settings

**Screenshot/Video:**
https://i.imgur.com/ZD6corV.gifv

comountainclimber:
Please award: 
@ranbena 200 points
@dvd-schwrtz 100 points
@evgenyboxer 100 points

lllwvlvwlll:
https://neoscan.io/transaction/52a8ff627aa7c29e0b398f7e3026818a9e23b9adc1fcf6e2398da5ac029ee0e0

PR #1787, opened 2019-01-10
feature: fixes node select bugs

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1729
BEFORE:
(XHR requests cancelled after a max of 2 seconds no matter what - high latency on the network or any other factors could cause the app to "infinitely load" without any resolution when trying to select a node. This gif shows the app in offline mode)

node-selection-before

AFTER:
(shows the application in offline mode)
node-selection-2
node-selection

There were a number of issues with the way the UX for node select was designed. This PR addresses several of them:
- https://github.com/CityOfZion/neon-wallet/blob/5af8f21a6751ed506f14b2db6dfba83b97bf92eb/app/actions/nodeNetworkActions.js#L40 as you can see, previously if this condition was never met (for any number of reasons) this promise would never resolve causing an infinite spinner via redux
- Leaving the user hanging without anyway to navigate back was very strange UX. This PR puts the loading indicator inside the context of the panel (with the ability to navigate back)
- It was impossible to ever render a proper error message with the ability to retry the action previously.
- Increases the timeout for RPC ping via neon-js settings
- Although not currently being used in this PR, this PR also fixes a long standing linting issue with iterators/generators (for of loops) which I am a huge fan of due to their readability and ease of maintaining. (I had originally committed code to this branch that used an async for of loop)
- Removes several stale and unreachable nodes from our hard coded main net list.

Solved this problem by rethinking how `nodeNetworkActions.js` functions...

Lots of manual testing locally. Happy to report node selection will work even with "slow 3g" network enabled. This PR also enables the dashboard to work with the same network settings 🎉 which previously was impossible due to default ping timeout of 2s

As you can see currently in `dev` branch a user with extremely slow connection speeds is essentially unable to use neon-wallet as ping requests are cancelled in neon-js by default after only 2s. This PR will allow users with even slow 3g connections to use neon to its fullest extent.

Overall this PR should improve performance for those with fast network connections (reduced timeout for ping when running auto select algo) and allow those with extremely limited bandwidth to use the app which was previously impossible

- [ ] Unit tests written?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -72,7 +72,9 @@
     "react/sort-comp": 0,
     "react/button-has-type": 0,
     "flowtype/generic-spacing": 0,
-    "no-param-reassign": ["warn"]
+    "no-param-reassign": ["warn"],
+    "no-restricted-syntax": ["error", "LabeledStatement", "WithStatement"],
+    "no-await-in-loop": 0
   },
   "settings": {
     "import/resolver": {
--- a/app/actions/nodeNetworkActions.js
+++ b/app/actions/nodeNetworkActions.js
@@ -21,6 +21,10 @@ const pingNodes = (nodes: Array<any>) => {
       let url = node.protocol ? `${node.protocol}://${node.url}` : node.url
       url = node.port ? `${url}:${node.port}` : url
 
+      const maxWaitTimer = setTimeout(() => {
+        resolve(responses)
+      }, RPC_PING_OVERRIDE)
+
       const client = new rpc.RPCClient(url)
       client
         .ping()
@@ -33,13 +37,11 @@ const pingNodes = (nodes: Array<any>) => {
             })
           }
           if (responses.length === nodes.length) {
+            clearTimeout(maxWaitTimer)
             resolve(responses)
           }
         })
         .catch(console.error)
-      setTimeout(() => {
-        resolve(responses)
-      }, RPC_PING_OVERRIDE)
     })
   })
 }
--- a/app/actions/nodeNetworkActions.js
+++ b/app/actions/nodeNetworkActions.js
@@ -9,60 +9,51 @@ import {
   TEST_NETWORK_ID,
   NODE_EXLUSION_CRITERIA,
 } from '../core/constants'
+import { raceAll } from '../util/promiseUtils'
 
 const ID = 'nodeNetwork'
-const RPC_PING_OVERRIDE = 5000
+const PING_TIMEOUT_OVERRIDE = 5000
+const DEFAULT_PING_TIMEOUT = settings.timeout.ping
 
-const pingNodes = (nodes: Array<any>) => {
-  settings.timeout.ping = RPC_PING_OVERRIDE
-  const responses = []
-  return new Promise(resolve => {
-    nodes.forEach(node => {
-      let url = node.protocol ? `${node.protocol}://${node.url}` : node.url
-      url = node.port ? `${url}:${node.port}` : url
-
-      const maxWaitTimer = setTimeout(() => {
-        resolve(responses)
-      }, RPC_PING_OVERRIDE)
-
-      const client = new rpc.RPCClient(url)
-      client
-        .ping()
-        .then(latency => {
-          if (client.lastSeenHeight !== 0) {
-            responses.push({
-              url,
-              blockCount: client.lastSeenHeight,
-              latency,
-            })
-          }
-          if (responses.length === nodes.length) {
-            clearTimeout(maxWaitTimer)
-            resolve(responses)
-          }
+const pingNode = node =>
+  new Promise(resolve => {
+    let url = node.protocol ? `${node.protocol}://${node.url}` : node.url
+    url = node.port ? `${url}:${node.port}` : url
+    const client = new rpc.RPCClient(url)
+    client.ping().then(latency => {
+      if (client.lastSeenHeight !== 0) {
+        resolve({
+          url,
+          blockCount: client.lastSeenHeight,
+          latency,
         })
-        .catch(console.error)
+      }
     })
   })
-}
+
+const pingNodes = (nodes: Array<any>) =>
+  raceAll(nodes.map(pingNode), PING_TIMEOUT_OVERRIDE)
 
 export default createActions(ID, ({ networkId }) => async () => {
   let nodes
+  const defaultCase = () =>
+    NODES_MAIN_NET.filter(
+      data =>
+        !NODE_EXLUSION_CRITERIA.some(criteria => data.url.includes(criteria)),
+    )
   switch (networkId) {
     case MAIN_NETWORK_ID:
-      nodes = NODES_MAIN_NET.filter(
-        data =>
-          !NODE_EXLUSION_CRITERIA.some(criteria => data.url.includes(criteria)),
-      )
+      nodes = defaultCase()
       break
     case TEST_NETWORK_ID:
       nodes = NODES_TEST_NET
       break
     default:
-      nodes = NODES_MAIN_NET.filter(
-        data =>
-          !NODE_EXLUSION_CRITERIA.some(criteria => data.url.includes(criteria)),
-      )
+      nodes = defaultCase()
   }
-  return pingNodes(nodes)
+  settings.timeout.ping = PING_TIMEOUT_OVERRIDE
+  const results = await pingNodes(nodes)
+  settings.timeout.ping = DEFAULT_PING_TIMEOUT
+  // filter out the undefined results that did not meet the max time alloted
+  return results.filter(node => node)
 })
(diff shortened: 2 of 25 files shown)


app/actions/nodeNetworkActions.js
```diff
@@ -10,67 +11,60 @@ import {
 } from '../core/constants'
 
 const ID = 'nodeNetwork'
+const MAX_RESPONSE_TIME = 10000
+const RPC_PING_OVERRIDE = 10000
```
dvd-schwrtz: can you explain these 2 variables. the last time I checked there was a timeout at 2 seconds in neon-js unless I am not remembering correctly. It has been a while

dvd-schwrtz: 10 seconds seems very excessive in any case

dvd-schwrtz: Just reread the overview section. maybe use 5 seconds so the user gets a faster notice. I doubt there will be good responses after 5 seconds

dvd-schwrtz: does the 3g work after 5 seconds?

comountainclimber: Sounds good I will test with 5 seconds... I agree that 10 could be a little excessive. Thanks for taking a look @dvd-schwrtz

.eslintrc
```diff
@@ -72,7 +72,9 @@
     "react/sort-comp": 0,
     "react/button-has-type": 0,
     "flowtype/generic-spacing": 0,
-    "no-param-reassign": ["warn"]
+    "no-param-reassign": ["warn"],
+    "no-restricted-syntax": ["error", "LabeledStatement", "WithStatement"],
+    "no-await-in-loop": 0
```
evgenyboxer: 👍

app/components/NodeSelectPanel/NodeSelectPanel.jsx
```diff
@@ -135,8 +143,18 @@ export default class NodeSelect extends React.Component<Props, State> {
...
+      selectedNode,
+      loading,
+      theme,
+      loadNodesData,
+      networkId,
+    } = this.props
+    if (loading && !nodes.length) {
+      return <Loading theme={theme} nobackground />
+    }
+    if (nodes.length) {
```
evgenyboxer: can we simplify the condition?

```suggestion
if (!nodes.length) {
  return loading ? (
    <Loading theme={theme} nobackground />
  ) : (
    <DialogueBox
      icon={<WarningIcon className={styles.warningIcon} />}
      renderText={() => (
        <div>
          Oops! There was an issue retrieving metrics from the network.{" "}
          <a onClick={() => loadNodesData({ networkId })}>Retry?</a>
        </div>
      )}
      className={styles.tokenSalePanelDialogueBox}
    />
  );
}

return ... (other condition)
```

app/components/NodeSelectPanel/NodeSelectPanel.jsx
```diff
@@ -135,8 +143,18 @@ export default class NodeSelect extends React.Component<Props, State> {
...
-    if (nodes) {
+    const {
+      nodes,
+      selectedNode,
+      loading,
+      theme,
+      loadNodesData,
+      networkId,
+    } = this.props
+    if (loading && !nodes.length) {
```
evgenyboxer: can we simplify the condition?

```suggestion
if (!nodes.length) {
  return loading ? (
    <Loading theme={theme} nobackground />
  ) : (
    <DialogueBox
      icon={<WarningIcon className={styles.warningIcon} />}
      renderText={() => (
        <div>
          Oops! There was an issue retrieving metrics from the network.{" "}
          <a onClick={() => loadNodesData({ networkId })}>Retry?</a>
        </div>
      )}
      className={styles.tokenSalePanelDialogueBox}
    />
  );
}

return ... (other condition)
```

app/actions/nodeNetworkActions.js
```diff
@@ -10,67 +11,60 @@ import {
...
-  const responses = []
-  return new Promise(resolve =>
-    promises.forEach(promise =>
-      promise.then(result => {
-        responses.push(result)
-        if (responses.length === total) resolve(responses)
-      }),
-    ),
-  )
+      // if condition on line 58 is never met return the results
```
evgenyboxer: I think its kinda problematic specifying a line number because it may change quickly making this comment irrelevant... Also, its always good to clear the timeout when you got the desired the result.

.eslintrc
```diff
@@ -72,7 +72,9 @@
     "react/sort-comp": 0,
     "react/button-has-type": 0,
     "flowtype/generic-spacing": 0,
-    "no-param-reassign": ["warn"]
+    "no-param-reassign": ["warn"],
```
evgenyboxer: not a fan of disabling this rule... it is a good rule and can help catch some really nasty bugs.

comountainclimber: Ok we can change that back... That was already committed to dev however and was not a change introduced by this PR

app/actions/nodeNetworkActions.js
comountainclimber: 5 seconds works for slow 3g 👍

.eslintrc
comountainclimber: For clarity I am going to revert the changes to this rule in a separate PR

app/actions/nodeStorageActions.js
```diff
@@ -47,6 +49,7 @@ export const getRPCEndpoint = async (
   net: Net,
   excludeCritera: Array<string> = NODE_EXLUSION_CRITERIA,
 ) => {
+  settings.timeout.ping = RPC_PING_OVERRIDE
```
ranbena: Overriding a global setting should better be avoided as it introduces the problematic "side-effect" anti-pattern. Suggestions:
1) Restore original setting on completion.
```js
const originalPing = settings.timeout.ping
settings.timeout.ping = RPC_PING_OVERRIDE // override
// do stuff
settings.timeout.ping = originalPing // restore
```
2) Update `neon-js` api to accept an override in current scope. Something like:
```js
data.client.ping(RPC_PING_OVERRIDE)
```
or
```js
const settings = {timeout: { ping: RPC_PING_OVERRIDE } }
const client = new rpc.RPCClient(url, settings)
```

app/actions/nodeNetworkActions.js
```diff
@@ -10,67 +11,58 @@ import {
...
-)
+export default createActions(ID, ({ networkId }) => async () => {
+  let nodes
+  switch (networkId) {
+    case MAIN_NETWORK_ID:
+      nodes = NODES_MAIN_NET.filter(
+        data =>
+          !NODE_EXLUSION_CRITERIA.some(criteria => data.url.includes(criteria)),
+      )
+      break
```
ranbena: nit: code duplication of switch default state. Suggesting to remove.

app/actions/nodeNetworkActions.js
```diff
@@ -10,67 +11,58 @@ import {
...
+            })
+          }
+          if (responses.length === nodes.length) {
+            clearTimeout(maxWaitTimer)
+            resolve(responses)
+          }
+        })
+        .catch(console.error)
+    })
+  })
```
ranbena: I think the timer method could be optimized with some nifty [Promise enhancements](https://stackoverflow.com/questions/48577702/setting-a-timeout-for-each-promise-within-a-promise-all).
Something like this:

Utils:
```js
Promise.delay = function(t) {
    return new Promise(resolve => {
        setTimeout(resolve, t);
    });
}

Promise.raceAll = function(promises, timeoutTime) {
    return Promise.all(promises.map(p => {
        return Promise.race([p, Promise.delay(timeoutTime)])
    }));
}
```
Usage:
```js
const pingNode = (node) => {
  return new Promise(resolve => {
    let url = node.protocol ? `${node.protocol}://${node.url}` : node.url
    url = node.port ? `${url}:${node.port}` : url
    new rpc.RPCClient(url)
      .ping()
      .then(latency => {
        if (client.lastSeenHeight !== 0) {
          resolve({
            url,
            blockCount: client.lastSeenHeight,
            latency,
          })
        }
      })
  })
}

const pingNodes = (nodes: Array<any>) => {
  settings.timeout.ping = RPC_PING_OVERRIDE
  return Promise.raceAll(nodes.map(pingNode), RPC_PING_OVERRIDE)
}
```
Why better?
* Doesn't rely on individual node ping promises to eventually resolve the accumulated responses.
* Doesn't recreate `maxWaitTimer` for each node ping promise.

comountainclimber: Wow I really like this strategy 🙌 took my a minute to wrap my head around but is much cleaner than is what is currently in the PR. Implementing now 😎

ranbena: Yay! Basically, there each node has a race between the timeout and the ping. The timeout rejects, the ping resolves. Whichever resolves gets added to the eventual response array.

```js
pingNodes(nodes).then((nodesThatPinged: Array<Node>) => {
  ...
});
```

Issue #1788, opened 2019-01-11
Bug: Fix Broken Layout in Receive (dev)


comountainclimber:
At some point the layout in the receive tab broke not sure how or when this popped up but it needs to be addressed before 2.1.0 can go out!

(currently on `dev`  branch)
<img width="1102" alt="screen shot 2019-01-10 at 9 05 58 pm" src="https://user-images.githubusercontent.com/13072035/51012698-b961b780-151b-11e9-9a26-dbf7800343b0.png">

comountainclimber:
please award @phetter 100

lllwvlvwlll:
https://neoscan.io/transaction/fe4c1ab1b0bfacf82ac686e065e5fdef376bbf77bea6c560c4715c8be4ce9ad5

PR #1789, opened 2019-01-11
chore/fix receive panel layout bug

phetter:
[Bug: Fix Broken Layout in Receive (dev) #1788](https://github.com/CityOfZion/neon-wallet/issues/1788)

Please see the screenshot in the above issue. The receive panel header was misaligned.

I adjusted the style sheet for the panel.

I tested it.

No.

I searched commits to find the correct one to provide context as to why this change was applied as the only outcome of the change that I could find was the alignment bug. Line 11 of style.css
https://github.com/CityOfZion/neon-wallet/commit/c03263b26f15770b113b615483d0bf043247cd1e#diff-49e0eb5369f7f556ca2b256f936f5e87R11

- [ ] Unit tests written?
No.

### DIFF
--- a/app/components/Receive/ReceivePanel/styles.scss
+++ b/app/components/Receive/ReceivePanel/styles.scss
@@ -7,9 +7,6 @@
   justify-content: center;
 }
 
-.receivePanel {
-}
-
 .receivePanelContent {
   display: flex;
   padding: 0;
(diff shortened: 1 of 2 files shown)


app/components/Receive/ReceivePanel/styles.scss
```diff
@@ -8,7 +8,7 @@
 }
 
 .receivePanel {
-  height: 100%;
+  // height: 100%;
```
dvd-schwrtz: if you are just going to comment out the height, just remove it completely

comountainclimber:
please award @phetter 100 👍

PR #1808, opened 2019-02-27
chore: fixes minor bugs in release

comountainclimber:
N/A

This PR fixes a problem that was causing this for loop to unnecessarily be iterated over and thus calling `pruneConfirmedOrStaleTransaction()`

```
    for (const transaction of toBePurged) {
      // eslint-disable-next-line
      await pruneConfirmedOrStaleTransaction(address, transaction.txid)
      handleGetPendingTransactionInfo()
    }
```

It also maps the `getPendingTransactionInfo` to props of the transaction history component so that the action can be dispatched after pending transactions are pruned from storage. This was causing issues in the delay between when the count in the sidebar would be removed. Previously the count would only decrement after two confirmations because the action was not being dispatched, now it decrements when the neoscan API returns the tx OR the node reports two confirmations.

Mapping the action to props and fixing the typo

Manual testing

N/A

N/A

- [ ] Unit tests written?

### DIFF
--- a/__tests__/components/__snapshots__/TransactionHistoryPanel.test.js.snap
+++ b/__tests__/components/__snapshots__/TransactionHistoryPanel.test.js.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`TransactionHistoryPanel renders without crashing 1`] = `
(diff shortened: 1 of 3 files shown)


evgenyboxer:
looks good to me!

comountainclimber:
Thanks for taking a look @evgenyboxer it is much appreciated as always 🙏

PR #1816, opened 2019-03-15
Updated TOLL to BRDG token with new supply

ThomasLobker:
Bridge Protocol has updated the NEP5 token from TOLL to BRDG. The TOLL token will not be used anymore. Also BRDG has a lower supply than TOLL because of a token burn event.

https://neonewstoday.com/general/bridge-protocol-to-conduct-token-swap-from-toll-to-brdg/


comountainclimber:
Thanks @ThomasLobker !

PR #1832, opened 2019-04-09
chore: yarn dev issues on Linux

phetter:
When executing 'yarn dev' no indication of progress was showing before. Now the build status is output as it goes.

This PR also addresses an issue with Linux (experienced personally on Ubuntu 18.04) where 'yarn dev' builds but Neon never opens. A note was made on the readme.

The PR resolves the issue of no clear feedback to the developer of build progress.

This PR also addresses an issue with Linux (experienced personally on Ubuntu 18.04) where 'yarn dev' builds but Neon never opens. A note was made on the readme.

I added the flag --progress to the package.json build script named 'dev'.

I tested it and verified it showed progress.

No.

No.

- [ ] Unit tests written?
No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Setup:
 Developing:
 
 - `yarn dev` - Start the application in development mode with hot reloading enabled
+  - There is a known condition that may arise on Linux systems where `yarn dev` builds but Neon never opens. Try using `yarn dev >/dev/null`.
 
 Running (for production):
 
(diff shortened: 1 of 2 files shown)


comountainclimber:
Please award @phetter 100 points

lllwvlvwlll:
https://neoscan.io/transaction/c382c87ce7282664dc849fef5cb8c8de22cf3b596dd1b4e85c87efb194bb2003

PR #1838, opened 2019-04-12
feature: adds neo news today RSS feed to app

comountainclimber:
This PR replaces the dead "voting" tab with a news tab via the RSS feed available at https://neonewstoday.com/feed/

Screen Shot 2019-04-12 at 3 35 21 PM

NOTE:
For security reasons we do not render the feed directly but instead attempt to parse the feed and render our own markup

TODO: 
- Add more automated testing
- Small tweaks and refining of UI elements on the News page


app/containers/News/News.jsx
```diff
@@ -0,0 +1,75 @@
...
+        </Panel>
+      </div>
+    )
+  }
+
+  parseItems() {
+    const { items } = this.props.feed
+    const imgTagRegex = new RegExp('<s*img[^>]*>(.*?)')
+
+    const imageHrefFromImgTags = img =>
```
comountainclimber: Getting the image href is pretty ugly like this but it was hard for me to think of anything cleaner/simpler as there is no image key on the available parsed XML

koko6767:
hi

PR #1849, opened 2019-05-14
added verifySideNav setup

tbeck860:
first PR test, need to add validation. Just seeing if this works :)


tbeck860:
did it wrong

PR #1850, opened 2019-05-14
added verifySideNav

tbeck860:
1st PR - Adds automated testing for the side nav routes and UI

(builds off of the work started in the `feature/end-to-end-tests` branch)

### DIFF
--- a/__e2e__/verifySideNav.e2e.js
+++ b/__e2e__/verifySideNav.e2e.js
@@ -1,0 +1,83 @@
+import { Application } from 'spectron'
+import path from 'path'
+import test from 'ava'
+
+let electronPath = path.join(
+  __dirname,
+  '..',
+  'node_modules',
+  '.bin',
+  'electron',
+)
+if (process.platform === 'win32') {
+  electronPath += '.cmd'
+}
+const appPath = path.join(__dirname, '..')
+
+const app = new Application({
+  path: electronPath,
+  args: [appPath],
+})
+
+test.before(async () => {
+  await app.start()
+  await app.client.waitUntilWindowLoaded()
+})
+
+test.after(async () => {
+  if (app && app.isRunning()) {
+    await app.stop()
+  }
+})
+
+test.serial('should login successfully', async t => {
+  // Go to login page
+  await app.client.waitUntilTextExists('#home', 'Login', 60000)
+  await app.client.click('#react-tabs-2')
+  // await app.client.click('div[aria-label="Private Key"]')
+
+  // // Enter Wif
+  await app.client.setValue(
+    'input[type="password"]',
+    'KxB52D1FGe5xBn6YeezNwj7grhkHZxq7bv2tmaCPoT4rxApMwMvU',
+  )
+  // Click on login btn
+  await app.client.click('#loginButton')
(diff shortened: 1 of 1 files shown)


comountainclimber:
This is awesome @tbeck860 ! A few things can you be sure to add assertions here in addition to running spectron (like the ones that exist in `view-balance.e2e.js`)

I am also wondering if we should encapsulate all of the tests that involved being authenticated in a single file rather than duplicating the code to login

tbeck860:
> This is awesome @tbeck860 ! A few things can you be sure to add assertions here in addition to running spectron and these automated clicks (like the ones that exist in `view-balance.e2e.js`)
> 
> I am also wondering if we should encapsulate all of the tests that involved being authenticated in a single file rather than duplicating the code to login. Let me know your thoughts

@comountainclimber If waitUntillTextExists fails the test will fail but the error is undefined. I am still trying to catch this error (promise<void) and output something more meaningful.

package.json
```diff
@@ -129,7 +129,8 @@
     "sifter": "0.5.3",
     "spunky": "1.3.1",
     "svg-react-loader": "0.4.5",
-    "uuid": "3.2.1"
+    "uuid": "3.2.1",
+    "webdriverio": "^5.8.4"
```
comountainclimber: Why exactly was this added? I dont see it being used in the new test

__e2e__/verifySideNav.e2e.js
```diff
@@ -30,7 +30,7 @@ test.after(async () => {
   }
 })
 
-test.serial('should login successfully', async t => {
+test.serial('should login successfully', async e => {
```
comountainclimber: I prefer that this variable be kept `t` as it is the ava the test instance (not an error)

__e2e__/verifySideNav.e2e.js
```diff
@@ -43,41 +43,56 @@ test.serial('should login successfully', async t => {
...
-
+  await app.client.waitUntilTextExists('//h3', 'Settings')
+  e.pass()
+})
+test.serial('Logout button clicked', async e => {
   //click logout
-  //await app.client.click('#logout')
-  //await app.client.waitUntilWindowLoaded(10000)
+  await app.client.click('#logout')
+  await app.client.waitUntilTextExists('//div[contains(@class,\'_3M2L1lj7oJRu_0KBN3_v87\')]', 'Login')
```
comountainclimber: I am worried about checks on dynamically generated class name hashes like this one... Can we just add an `id` tag to the element you are checking instead?

comountainclimber:
Keep up the good work @tbeck860 I have some feedback for you based on the latest changes

comountainclimber:
@tbeck860 this looks like its ready to go! Nice job... Last bit of feedback and then we can merge: Can you revert your changes to both `package.json` and `package-lock.json` - those _should_ remain unchanged with this PR

comountainclimber:
Please award @tbeck860 200 points

lllwvlvwlll:
https://neoscan.io/transaction/c382c87ce7282664dc849fef5cb8c8de22cf3b596dd1b4e85c87efb194bb2003

Issue #1846, opened 2019-05-13
feature suggestion: offline signing


comountainclimber:
Although seemingly very complicated user experience, neon should support the ability for users to sign transactions offline for future broadcast onto the network. Would like to open up the discussion around this feature in this thread

PR #1867, opened 2019-06-04
feature: sign transactions

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1846

This PR completes the work for generating and signing both contract and invocation transactions using Neon. An additional PR adding an extremely simple "offline UX" will be the icing on the cake and complete this entire set of features. My thoughts were that we would determine if the client is offline using the navigator API and then present a prompt asking the user if they would like to sign a transaction, this would render the import modal added with this PR completing the entire offline signing user experience

- [ ] Unit tests written?

TODO:
- ~Add ledger support for signing (separate PR)~ https://github.com/CityOfZion/neon-wallet/pull/1877
- ~Fully implement support for file upload / creation~ https://github.com/CityOfZion/neon-wallet/pull/1867/commits/ce816a66b339c1f7b03be09403d691077a05689d
- ~Add new UX for offline mode using `navigator` api~ https://github.com/CityOfZion/neon-wallet/pull/1878
- Add testing (separate PR)

sign-transaction


app/components/Modals/GeneratedTransactionModal/GeneratedTransactionModal.jsx
```diff
@@ -119,7 +122,10 @@ export default class GeneratedTransactionModal extends React.Component<
           <div className={baseStyles.section}>
             <div className={baseStyles.sectionTitle}>TRANSACTION OUPUT</div>{' '}
             <div className={styles.transactionOutput}>
-              <pre>{JSON.stringify(this.props.tx, null, 2)}</pre>
+              <JSONPretty
```
comountainclimber: Was having some issues with just using the `<pre>` tag here so added this ultra weight dependency. This library also adds the ability to format JSON with CSS which could be a nice future enhancement

__tests__/components/Send.test.js
```diff
@@ -6,6 +6,7 @@ import Send from '../../app/containers/Send/Send'
 import ZeroAssets from '../../app/components/ZeroAssets/ZeroAssets'
```
comountainclimber: The tests here all had to be updated because they were making assertions with the assumption that only one button existed on the page, there was also a new "panel header" button added

app/modules/transactions.js
```diff
@@ -100,6 +107,18 @@ export const generateBalanceInfo = (
   })
 }
 
+// This adds some random bits to the transaction to prevent any hash collision.
+const attachAttributesForEmptyTransaction = (config: api.apiConfig) => {
```
comountainclimber: Without this function (which I dug out of neon-js) there was validation errors for every single invocation transaction but by adding it here I am able to perform offline signing for any transaction type 🎉

Issue #1863, opened 2019-05-30
Ability to import account using encrypted key


canesin:
Currently it is only possible to import a account using a WIF, that in itself presents a security risk as you have to copy the key or type the key in full. While not a strong protection copying the encrypted key is safer.

This feature is a regression as in NEON prior to v2 was possible to import a encrypted key account.

comountainclimber:
Thanks for the feedback and opening this issue @canesin I will make sure it gets prioritized/fixed

dauTT:
Hi @canesin,
you are referring to this import? You would like to add an option to import encrypted key?

![image](https://user-images.githubusercontent.com/30392990/59152043-2df24100-8a3d-11e9-801a-8fce079373ca.png)

canesin:
That is correct @dauTT !

dauTT:
@comountainclimber , 

FYI, I am working on this issue.

dauTT:
@comountainclimber , we should probably close this issue. :+1:

Issue #1862, opened 2019-05-28
feature suggestion: We should render a QR for encrypted keys (without passwords)


comountainclimber:
For consistency we should offer the ability for users to save a copy of a QR of their encrypted key (exactly how we provide address/private key) this will allow users to take advantage of the work completed here: https://github.com/CityOfZion/neon-wallet/pull/1855

dauTT:
@comountainclimber,
the option to generate the QR (address/private key) is only possible after the **creation** or the **import** of a **wallet**. Where exactly do you want to add the option to generate the QR of encrypted key?

I believe we should add this option after :
1) `creation wallet`
2) `import wallet` (option encrypted key: #1863)
3) `Encrypt a key` (This option is available in the **setting** section)

comountainclimber:
@dauTT that is exactly correct... Those would be the three places. Please note the work to scan the QR has already been completed here: https://github.com/CityOfZion/neon-wallet/pull/1855

dauTT:
I would like to check the code changes that I have made on the `test` net.

How do you usually create a new wallet on the `test` net? When I run 

`yarn dev >/dev/null`

neon wallet is connecting to the `main` net by default and the home page does not have any option to switch between `test` and `main`.

comountainclimber:
@dauTT A wallet address will "work" on both networks if created using neon

comountainclimber:
For development purposes I recommend always using the command `yarn dev`

comountainclimber:
Taken care of in https://github.com/CityOfZion/neon-wallet/pull/1873

PR #1872, opened 2019-06-15
feature suggestion: We should render a QR for encrypted keys (without passwords)

dauTT:

#1862

The PR add the feature to render QR code for the encrypted key.

I followed more or less the logic for rendering the QR code for private key and public address and adapted it for the encrypted key. This logic is essentially stored in these two modules: 

- `DisplayWalletAccountsQrCodes`
- `DisplayWalletAccounts`

- I tested manually the rendering of the encrypted key QR for these two scenarios:
   1) `Create Wallet`
   2) `Encrypt Key` (option available in the setting section.)

- I also did run `yarn test` and all the existing tests are passing.

Yes. We should also include the option to render encrypted key QR code after selecting the option 
`Import Wallet`

This will be done in the next PR addressing issue #1863.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Root/Routes.jsx
+++ b/app/components/Root/Routes.jsx
@@ -22,6 +22,7 @@ import TokenSale from '../../containers/TokenSale'
 import Encrypt from '../../containers/Encrypt'
 import NodeSelect from '../../containers/NodeSelect'
 import News from '../../containers/News'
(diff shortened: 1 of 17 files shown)


dauTT:
I selected the wrong branch. I will close this PR and create a new one. :sweat_smile:

PR #1873, opened 2019-06-15
feature suggestion: We should render a QR for encrypted keys (without passwords)

dauTT:

#1862

The PR add the feature to render QR code for the encrypted key.

I followed more or less the logic for rendering the QR code for private key and public address and adapted it for the encrypted key. This logic is essentially stored in these two modules: 

- `DisplayWalletAccountsQrCodes`
- `DisplayWalletAccounts`

- I tested manually the rendering of the encrypted key QR for these two scenarios:
   1) `Create Wallet`
   2) `Encrypt Key` (option available in the setting section.)

- I also did run `yarn test` and all the existing tests are passing.

Yes. We should also include the option to render encrypted key QR code after selecting the option 
`Import Wallet`

This will be done in the next PR addressing issue #1863.

- [ ] Unit tests written?

### DIFF
--- a/app/components/Settings/EncryptPanel/index.js
+++ b/app/components/Settings/EncryptPanel/index.js
@@ -1,1 +1,28 @@
-export { default } from './EncryptPanel'
+// @flow
+
+import { connect } from 'react-redux'
+import { compose } from 'recompose'
+import { bindActionCreators } from 'redux'
+import {
+  getEncryptedWIF,
+  resetEncryptedWIF,
+} from '../../../modules/generateEncryptedWIF'
+import EncryptPanel from './EncryptPanel'
+
+const mapStateToProps = (state: Object) => ({
+  encryptedWIF: getEncryptedWIF(state),
+})
+
+const actionCreators = {
+  resetEncryptedWIF,
+}
+
+const mapDispatchToProps = dispatch =>
+  bindActionCreators(actionCreators, dispatch)
+
+export default compose(
+  connect(
+    mapStateToProps,
+    mapDispatchToProps,
+  ),
+)(EncryptPanel)
--- a/app/components/Settings/EncryptPanel/index.js
+++ b/app/components/Settings/EncryptPanel/index.js
@@ -1,5 +1,4 @@
 // @flow
-
 import { connect } from 'react-redux'
 import { compose } from 'recompose'
 import { bindActionCreators } from 'redux'
--- a/app/components/Settings/EncryptQR/EncryptQR.jsx
+++ b/app/components/Settings/EncryptQR/EncryptQR.jsx
@@ -15,6 +15,7 @@ import withCopyCanvasToClipboard from '../../../hocs/withCopyCanvasToClipboard'
 
 type Props = {
   encryptedWIF: string,
+  resetEncryptedWIF: Function,
   handleCopy: (?HTMLCanvasElement, string, ?boolean) => Promise<void>,
   handleCreateCanvas: (?HTMLCanvasElement, string) => any,
   copied: boolean,
@@ -29,13 +30,18 @@ class EncryptQR extends Component<Props> {
   }
 
   render() {
+    const { resetEncryptedWIF } = this.props
     return (
       <FullHeightPanel
         headerText="Encrypted QR Code"
         renderInstructions={false}
         headerContainerClassName={styles.headerIconMargin}
         renderHeaderIcon={() => <CheckIcon />}
-        renderCloseButton={() => <CloseButton routeTo={ROUTES.SETTINGS} />}
+        renderCloseButton={() => (
+          <div onClick={resetEncryptedWIF}>
+            <CloseButton routeTo={ROUTES.SETTINGS} />
+          </div>
+        )}
         renderBackButton={() => <BackButton routeTo={ROUTES.ENCRYPT} />}
         iconColor="#F7BC33"
       >
--- a/app/components/Settings/EncryptQR/EncryptQR.jsx
+++ b/app/components/Settings/EncryptQR/EncryptQR.jsx
@@ -20,7 +20,7 @@ type Props = {
   copied: boolean,
 }
 
-class EncryptQR extends Component<Props, State> {
+class EncryptQR extends Component<Props> {
   encryptedCanvas: ?HTMLCanvasElement
 
   componentDidMount() {
--- a/app/components/Settings/EncryptQR/EncryptQR.jsx
+++ b/app/components/Settings/EncryptQR/EncryptQR.jsx
@@ -1,0 +1,83 @@
+// @flow
+import React, { Component } from 'react'
+
+import Button from '../../Button'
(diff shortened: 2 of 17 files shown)


comountainclimber:
Excellent work @dauTT! Will be reviewing your PR later this evening

app/components/Settings/EncryptPanel/index.js
```diff
@@ -1 +1,28 @@
-export { default } from './EncryptPanel'
+// @flow
+
```
comountainclimber: Mostly just a nitpick but can we remove this empty line here (no line break needed after flow annotation)

app/components/Settings/EncryptQR/index.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
```
comountainclimber: Same as comment above

app/containers/Encrypt/index.js
```diff
@@ -1,26 +1,27 @@
 // @flow
+
```
comountainclimber: no line break needed after flow annotation

app/modules/generateEncryptedWIF.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+
```
comountainclimber: no line break needed after flow annotation

app/containers/DisplayWalletAccounts/DisplayWalletAccountsQrCodes/DisplayWalletAccountsQrCodes.jsx
```diff
@@ -95,6 +100,14 @@ class DisplayWalletAccountsQrCodes extends Component<Props, State> {
...
               />
             </div>
+            <div className={styles.qr}>
+              <label> encrypted key </label>
+              <canvas
+                ref={node => {
+                  this.encryptedCanvas = node
+                }}
+              />
+            </div>
```
comountainclimber: Nice job here! Would like it if we could add padding around the div containing the QR codes... It looks a little a little off without any. I think it would be nice to also add the ability for users to copy the encrypted key qr to clipboard like they can for the public address qr
Screen Shot 2019-06-16 at 7 22 10 PM

app/components/Settings/EncryptQR/EncryptQR.jsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  render() {
+    return (
+      <FullHeightPanel
+        headerText="Encrypted QR Code"
+        renderInstructions={false}
+        headerContainerClassName={styles.headerIconMargin}
+        renderHeaderIcon={() => <CheckIcon />}
+        renderCloseButton={() => <CloseButton routeTo={ROUTES.SETTINGS} />}
```
comountainclimber: I think it would be a good idea to invoke `resetEncryptedWIF` on close here. Resetting the state after closing the panel seems to make the most sense from a UX perspective rather than having the user click "Encrypt another private key"

app/components/Settings/EncryptSuccess/EncryptSuccess.jsx
```diff
@@ -39,6 +42,21 @@ export default class EncryptSuccess extends React.Component<Props> {
...
+            <Button renderIcon={AddIcon} primary onClick={this.handlePrint}>
+              Print
+            </Button>
+            <NavLink
+              id="display-encrypted-wif-qr"
+              exact
+              to={ROUTES.DISPLAY_ENCRYPTED_WIF_QR}
+            >
+              <Button primary renderIcon={() => <GridIcon />} type="submit">
+                Generate QR Codes
```
comountainclimber: Can this say "Generate QR Code" instead of "Codes"

app/containers/DisplayWalletAccounts/DisplayWalletAccounts.jsx
```diff
@@ -87,7 +90,10 @@ class DisplayWalletAccounts extends Component<Props> {
               <div key={item.label} className={styles.detailRow}>
                 <div
                   className={classNames(styles.input, {
-                    [styles.reducedInputFontSize]: item.label === 'Private Key',
+                    [styles.reducedInputFontSize]: [
+                      'Private Key',
+                      'Encrypted Key',
+                    ].includes(item.label),
```
comountainclimber: 👍

app/containers/Encrypt/index.js
```diff
@@ -1,26 +1,27 @@
...
-  if (passphrase !== confirmPassphrase) {
-    throw new Error('Passphrases do not match')
-  }
-  if (!validatePassphraseLength(passphrase)) {
-    throw new Error('Please choose a longer passphrase')
-  }
-  if (privateKey && !wallet.isWIF(privateKey)) {
-    throw new Error('The private key is not valid')
-  }
-  return wallet.encrypt(privateKey, passphrase)
```
comountainclimber: 👍

app/containers/DisplayWalletAccounts/DisplayWalletAccountsQrCodes/DisplayWalletAccountsQrCodes.jsx
dauTT: In regards to the padding, Could you set the proper measure? I could not tell whether it is better using the padding or not. :sweat_smile:

comountainclimber: After taking a closer look I think something like this will work just fine

Screen Shot 2019-06-18 at 7 00 54 PM
Screen Shot 2019-06-18 at 7 01 02 PM


Also lets remove the ability to copy the private key code to the clipboard (keep that as it is currently in `master`)

app/containers/DisplayWalletAccounts/DisplayWalletAccountsQrCodes/DisplayWalletAccountsQrCodes.jsx
```diff
@@ -99,6 +112,19 @@ class DisplayWalletAccountsQrCodes extends Component<Props, State> {
                   this.privateCanvas = node
                 }}
               />
+              <Button
```
comountainclimber: Let's actually remove this and only have the ability to copy encrypted and public. Having the private key QR on clipboard could introduce security concerns because unlike copying the text to the clipboard copying the QR in its current format using `withCopyCanvasToClipboard.js` requires us to write the file to disk

comountainclimber: Thanks for addressing this one 😄

comountainclimber:
Please award @dauTT 500 points

lllwvlvwlll:
https://neoscan.io/transaction/589328c622e101065f6717d61a332f93aac71cc68b665e194ef8c874a7e26669

PR #1878, opened 2019-06-19
feature: offline signing prompt

comountainclimber:
This PR is the final piece of the offline signing flow... It presents a prompt to the user if they are disconnected from the internet asking if they would like to sign a transaction

This provides an "offline" UX for users who wish to perform offline signing on an air gapped machine

I solved this problem by adding two key value pairs to state, one that checks whether the user has a connection during authentication and another that checks whether or not the user has been presented with offline prompt

Manual testing of many different permutations of use

- [ ] Unit tests written?

offline-prompt

### DIFF
--- a/app/actions/authActions.js
+++ b/app/actions/authActions.js
@@ -37,7 +37,6 @@ type AccountType = ?{
   isHardwareLogin: boolean,
   isWatchOnly?: boolean,
   hasInternetConnectivity: boolean,
-  internetConnectionPromptPresented: boolean,
 }
 
 export const ID = 'auth'
@@ -51,12 +50,6 @@ export const checkForInternetConnectivity = (): Promise<boolean> =>
       resolve(true)
     })
   })
-
-export const internetPromptPresentedActions = createActions(ID, () => (): {
-  internetConnectionPromptPresented: boolean,
-} => ({
-  internetConnectionPromptPresented: true,
-}))
 
 export const wifLoginActions = createActions(
   ID,
@@ -73,7 +66,6 @@ export const wifLoginActions = createActions(
       address: account.address,
       isHardwareLogin: false,
       hasInternetConnectivity,
-      internetConnectionPromptPresented: false,
     }
   },
 )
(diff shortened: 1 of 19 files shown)


app/actions/authActions.js
```diff
@@ -35,38 +36,53 @@ type AccountType = ?{
...
   isHardwareLogin: boolean,
   isWatchOnly?: boolean,
+  hasInternetConnectivity: boolean,
 }
 
 export const ID = 'auth'
 
+export const checkForInternetConnectivity = (): Promise<boolean> =>
+  new Promise(resolve => {
+    dns.lookup('google.com', err => {
```
comountainclimber: Without adding any sort of external library this seemed like the least bad option

app/actions/internetConnectivityPromptActions.js
```diff
@@ -0,0 +1,13 @@
...
+export const ID = 'internet-connectivity-prompt'
+
+export const internetConnectionPromptPresented = createActions(
+  ID,
+  (bool = false) => (): {
+    internetConnectionPromptPresented: boolean,
+  } => ({
+    internetConnectionPromptPresented: bool,
+  }),
+)
```
comountainclimber: A simple boolean being added to state to determine whether or not the Dashboard should redirect to the offline prompt

app/actions/authActions.js
sfreeman422: `dns.resolve()` may be a better option here due to performance concerns as well as consistently using the network as mentioned in https://nodejs.org/api/dns.html#dns_implementation_considerations

app/actions/authActions.js
```diff
@@ -35,38 +36,53 @@ type AccountType = ?{
   signingFunction?: Function,
   isHardwareLogin: boolean,
   isWatchOnly?: boolean,
+  hasInternetConnectivity: boolean,
```
comountainclimber: Adding this here seemed like the best place for me as we care about performing the check for this particular flow at the point where a user signs in

Issue #1621, opened 2018-11-01
"Import" functionality is missing on wallet start page


drptbl:
Currently you're able to export your wallet details, but to import them you have to create a dummy wallet, login to it, then go to "Settings" to see import functionality. It should be accessible from start page of the wallet without requirement to login.

phetter:
I have a cople of things that have come up regarding this as I was looking at working on it:

1. If you go into settings and recover wallets (via NEP-6 file) it will pull a list of wallets that you can see under Manage Wallets, but you can't switch between them from there. I'd like to add a feature to the Manage Wallets page to switch to another wallet in the list. I propose we add an item to the drill down, and perhaps to the list view of wallets, that lets you switch to that wallet. "Set Active", "Use this wallet", or similar. This button, imo, would just log you out and redirect you to the login with that wallet pre-filled in the login section or popup a login pane. I'm leaning toward the login pane popup.

2. I want to add a button on the main screen that says "import recovery file", or whatever language you think is best, that will let you do so without having to first login. The import would perform the import then redirect you back to the login screen with the dropdown populated with the list of wallets for you to select to login. We would keep the import wallet button as is.

dauTT:
@comountainclimber ,

Do you simple want to add a new tab, `Recover Wallet` on the `import wallet` page? The functionality  of `Recover Wallet` is the same as the one that we have in the `Settings` ?

![image](https://user-images.githubusercontent.com/30392990/60055821-2a1e1a00-96df-11e9-9c01-252ce582a37e.png)

![image](https://user-images.githubusercontent.com/30392990/60055961-9b5dcd00-96df-11e9-81ea-7197883820df.png)

comountainclimber:
<img width="1222" alt="Screen Shot 2019-06-24 at 7 21 00 PM" src="https://user-images.githubusercontent.com/13072035/60058147-4311d580-96b5-11e9-8dd9-029ceace393b.png">

comountainclimber:
@dauTT I was thinking something like this let me know what you think

uvmetal:
I kind of like this, but I still feel like recover and import are confusing. It seems it may be the use of the term "wallet". Importing a NEP6 versus importing a private key, etc, is where it gets ambiguous, imho.

uvmetal:
Maybe some clarification along the lines of import wallet by file versus by key?

dauTT:
[...]

PR #1879, opened 2019-06-21
Add option to import wallet account using encrypted key

dauTT:

#1863

Add option to import wallet account using encrypted key

I have generalized the method `generateNewWalletAccount` to accept a new input parameter `keyoption`. This parameter can assume two value:
-1) WIF
-2) ENCRYPTED_WIF
This parameter is actually active only if the user has chosen the `Import wallet` option. If the user has choosen `Create Wallet`, the parameter is not active and the behavior of the method is exactly the same as before. 

tested manually the code under difference scenarios:

1)  "Create Wallet" : valid and invalid input parameters
2)  "Import Wallet" with key option "PRIVATE KEY" and with valid and invalid input parameters.
3) "Import Wallet" with key option "ENCRPYTED KEY" and with valid and invalid input parameters.

Once #1873 is merged with this PR in `dev`, we will be able to render the encrypted key QR code also after choosing and completing the "Import Wallet" option.

- [x] Unit tests written?


app/containers/CreateWallet/CreateWallet.scss
```diff
@@ -27,3 +27,9 @@
   margin-top: auto;
   margin-bottom: 50px;
 }
+
+.SelectContainer {
```
comountainclimber: For consistency can we stick with the pattern of having class names `camelCased`

app/containers/CreateWallet/CreateWallet.jsx
```diff
@@ -107,14 +136,31 @@ export default class CreateWallet extends React.Component<Props, State> {
...
-                <PasswordInput
-                  value={wif}
-                  label="Private Key"
-                  onChange={e => this.setState({ wif: e.target.value })}
-                  placeholder="Private Key"
-                  autoFocus
-                />
+                <div>
+                  <div className={styles.SelectContainer}>
+                    <StyledReactSelect
```
comountainclimber: Sorry to be so nitpicky but I think for consistency we should actually use our "tabs" component here instead of the select drop down... You can see an example of this on the Login page container. I think this will look better here

app/modules/generateWallet.js
```diff
@@ -183,94 +183,137 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
...
-        onFailure()
-        return dispatchError('A wallet with this name already exists locally')
-      }
+  let wif = ''
+  // If the key is not given, it means that the user has choosen
+  // the option 'Create Wallet'. Therefore isImport = false.
+  // If the key is given, isImport = true.
+  const isImport = key === null
+
+  Promise.resolve()
```
comountainclimber: What exactly is the point of having this promise resolution here?

app/modules/generateWallet.js
```diff
@@ -183,94 +183,137 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
...
+    .then(async () => {
+      wif =
+        keyOption === 'WIF' ? key : await decryptEncryptedWIF(key, passphrase)
+
+      validateInputs(wif, passphrase, passphrase2)
+    })
+    .then(() =>
+      setTimeout(async () => {
```
comountainclimber: Do we need the `setTimeout`?

comountainclimber:
@dauTT if you are interested in taking this on next I think it would compliment the work you have done here (and would look great with the tabbed view) https://github.com/CityOfZion/neon-wallet/issues/1621

app/containers/CreateWallet/CreateWallet.scss
dauTT: Thanks for the advice! I will keep in mind going forward. For this particular case, I do not need `SelectContainer` anymore because I will use Tabs component as you have suggested below.

app/containers/CreateWallet/CreateWallet.jsx
dauTT: I like much better with tabs component. This is really a good improvement on the look and feel. Thanks for the suggestion!

app/modules/generateWallet.js
dauTT: Sorry, I will remove it if it makes confusion.

app/modules/generateWallet.js
dauTT: After removing `setTimeout` everything works. (At least for the few cases that I tested). This particular method was already there before I made my changes. So I not sure if removing it will cause some side effects.

dauTT:
Sure, I am happy to work on #1621. Thanks for suggestion! :+1:

app/containers/CreateWallet/CreateWallet.jsx
dauTT: image

comountainclimber: @dauTT this is looking really good! Would we be able to use all upper case text in tabs (for consistency) and remove the "instructions" like in this mock here (you can leave out the last tab in this PR):
Screen Shot 2019-06-24 at 7 21 00 PM

app/modules/generateWallet.js
```diff
@@ -184,94 +184,133 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
...
+          keyOption === 'WIF' ? wallet.encrypt(WIF, passphrase) : key
+        const storedWallet = await getWallet()
+        if (walletName && walletHasLabel(storedWallet, walletName)) {
+          onFailure()
+          return dispatchError('A wallet with this name already exists locally')
+        }
+
+        if (walletHasKey(storedWallet, encryptedWIF)) {
+          onFailure()
+          return dispatchError('A  already exists locally')
```
comountainclimber: What is this error supposed to represent? I believe its supposed to say something like "A wallet with this encrypted key already exists locally."

app/modules/generateWallet.js
```diff
@@ -184,94 +184,133 @@ export const recoverWallet = (wallet: Object): Promise<*> =>
...
+    .then(async () => {
+      try {
+        const account = new wallet.Account(wif || wallet.generatePrivateKey())
+        const { WIF, address } = account
+        const encryptedWIF =
+          keyOption === 'WIF' ? wallet.encrypt(WIF, passphrase) : key
+        const storedWallet = await getWallet()
+        if (walletName && walletHasLabel(storedWallet, walletName)) {
+          onFailure()
+          return dispatchError('A wallet with this name already exists locally')
```
comountainclimber: Can we add punctuation here `'A wallet with this name already exists locally'.`

comountainclimber:
@dauTT this is looking excellent and is super close... Just some minor feedback and then we can merge

comountainclimber:
Please award @dauTT 300 points

lllwvlvwlll:
https://neoscan.io/transaction/589328c622e101065f6717d61a332f93aac71cc68b665e194ef8c874a7e26669

PR #1882, opened 2019-06-23
feature: offline signing

comountainclimber:
This PR encapsulates the entirety of offline signing UI/UX into Neon. It is composed of a several new routes, container components and modals:

WHATS NEW:
1.) Offline signing prompt presented to users if internet connectivity check detects user is not connected
Screen Shot 2019-06-23 at 11 56 34 AM

2.) "Generate transaction JSON" button in Send container (available in watch only and normal authentication)
Screen Shot 2019-06-23 at 12 09 53 PM

3.) Import "panel header" button in Send container for importing generated transaction JSON and signing. The import modal is also used to broadcast. After signing the modal will produce a signed serialized transaction that can get directly sent to RPC nodes
Screen Shot 2019-06-23 at 12 10 54 PM

4.) "Import" and "Generate" modals both support node js fs module for saving and importing .txt and .json files that neon now generates for transactions
Screen Shot 2019-06-23 at 12 12 36 PM

RELATED PULL REQUESTS:
https://github.com/CityOfZion/neon-wallet/pull/1877
https://github.com/CityOfZion/neon-wallet/pull/1867
https://github.com/CityOfZion/neon-wallet/pull/1859
https://github.com/CityOfZion/neon-wallet/pull/1878

This PR presents an alternative way for users interested in cold storage of private keys to interact with neon and the NEO ecosystem such that users will never have to expose private keys to a machine that is connected to the internet.

By breaking down the problem and feature set into the simplest possible user experience I could imagine

A great deal of manual regression testing around the following paths:
1.) Invocation and Contract transactions using the offline signing components
2.) That these new pieces are fully functional with ledger nano
3.) Existing invocation and contract transactions remain fully functional

I worked very hard to ensure that existing functionality was minimally affected or not affected at all.

Ideally we would have some sort of "confirmation" of exactly what assets are being sent and to who (similar to what exists currently in Send) however I did not feel this 100% necessary for MVP of this feature and release as users can see their exact transaction details in the JSON output

- [x] Unit tests written?

TODO:
- PR to add additional unit and integration tests around the new work in addition to the tests that have already been updated
- Would be nice to conditionally disable buttons like "broadcast" and "sign transaction" until the text area has valid input (although this is not necessary for release)
- Have other users interact with this flow for feedback
- Verify that the instructions / details in text we provide is sufficient for the average user to understand whats going on


jseagrave21:
When I tried to create a raw transaction, it appears the type is incorrect:
image

I was expecting to see `ContractTransaction`

jseagrave21:
NOTE: I am logged in to a watch-only testnet address, but I think this should still work.

When I copied the above raw transaction and then imported it and tried to sign it, I received this error:
image

jseagrave21:
Everything seems to work fine if I am actually logged into my wallet; however, I think you should consider the option of signing the transaction [if on a watch-only address] with the WIF key or Encrypted Key plus Passphrase.

My thinking comes from my `RawTransaction` class in neo-python where you can build a tx and view it with `ToJson` without signing it. Then sign it with the WIF or Encrypted Key as described above without ever being logged in to any wallet. (see https://github.com/jseagrave21/neo-python/blob/f75cb5dc58bedbfc12ea1a6afbca5332088f8ca6/neo/Core/TX/RawTransaction.py#L451)

jseagrave21:
Also, since `SimplePolicyPlugin` is mandatory now, how do you provide feedback to the user if a fee is required or not? I think this feedback could apply to normal sending tx but have you considered displaying the size of the tx in the raw output? Or simply not allowing them to build [or send] the tx without the required fee and displaying what that fee is. The latter is how my `RawTransaction` class works which you can see here https://github.com/jseagrave21/neo-python/blob/f75cb5dc58bedbfc12ea1a6afbca5332088f8ca6/neo/Core/TX/RawTransaction.py#L541.

A version of this was just implemented in neo-python here https://github.com/CityOfZion/neo-python/pull/964/files#diff-87c9c1e2a861910053534dc0bad32f6bR608

jseagrave21:
As an additional TODO, I think adding the ability to build and sign raw Claim transactions would be good. You can see how I do it here: https://github.com/jseagrave21/neo-python/blob/f75cb5dc58bedbfc12ea1a6afbca5332088f8ca6/neo/Core/TX/RawTransaction.py#L346-L374

jseagrave21:
Considering logging in while offline, I think it would be good to switch to the generic 'loading' screen (similar to when you first open Neon wallet) while waiting for the connectivity test to complete because I waited 51 seconds after entering my passphrase (I timed it) before seeing the "offline" screen. The first time, I thought something had gone wrong.

jseagrave21:
There seems to be no way to send a transaction generated from a multi-sig address. I can generate the raw transaction by viewing the address in `watch-only`. And then I can sign it once. However, my multi-sig address requires 2 signatures and I am unable to sign the transaction again.

Recommended TODO: Add multi-sig support for raw transactions.

comountainclimber:
@jseagrave21 thanks so much for taking such a comprehensive look man!

1.) In regards to your first comment (`ContractTransaction` vs `InvocationTransaction`)  neon-js is actually producing these numbers instead and the serialized transaction works exactly as expected so at this point I am not super concerned although am confused as to whats going on there


2.)
> When I copied the above raw transaction and then imported it and tried to sign it, I received this error:

> Everything seems to work fine if I am actually logged into my wallet; however, I think you should consider the option of signing the transaction [if on a watch-only address] with the WIF key or Encrypted Key plus Passphrase.

That is definitely a bug as users will not be able to sign transactions in watch only. I do not plan on supporting that UX at all as it is unnecessarily complicated. Thank you for pointing that out... My plan is to simply disable signing in watch only however a user can still use watch only to "broadcast"

3.) Regarding the `SimplePolicyPlugin` I think this would eliminate a great deal of confusion however I am not extremely familiar with the new specification I will take a look at the code you and try and include some sort of validation that looks for a fee based on size. Thank you for this suggestion 🙏 

4.)

>  I waited 51 seconds after entering my passphrase (I timed it) before seeing the "offline" screen

No idea what is going on here but I will take a look... The only check being run is this:

```
export const checkForInternetConnectivity = (): Promise<boolean> =>
  new Promise(resolve => {
    dns.resolve('google.com', 'A', err => {
      if (err) {
        return resolve(false)
      }
      return resolve(true)
    })
  })
```

On my machine when not connected this resolves instantly 😕 . Perhaps there is a more efficient way to determine whether a user has a connection

5.) With all of the new features included here I do not plan to support multi sig signing in this initial release of this feature although it is something that I plan on iterating towards using the new components. Ideally we would release the ability to create multi sig addresses in neon at the same time as well

jseagrave21:
@comountainclimber thank you for the feedback. This is the original PR where I worked on implementing the `SimplePolicyPlugin`: https://github.com/CityOfZion/neo-python/pull/869
Please see the discussion there.
This is the key logic:
```
if wallet_tx.Size() > settings.MAX_FREE_TX_SIZE:
                req_fee = Fixed8.FromDecimal(settings.FEE_PER_EXTRA_BYTE * (wallet_tx.Size() - settings.MAX_FREE_TX_SIZE))
                if req_fee < settings.LOW_PRIORITY_THRESHOLD:
                    req_fee = settings.LOW_PRIORITY_THRESHOLD
                if net_fee < req_fee:
                    net_fee = req_fee
```
You can reference the key values from the PR links.
The similar version found in my `RawTransaction` class may be more usable for you since our application is similar:
```
        if not self._network_fee:
            self._network_fee = Fixed8.Zero()
        fee = self._network_fee
        if self.Size() > settings.MAX_FREE_TX_SIZE and not self.Type == b'\x02':  # Claim Transactions are High Priority
            req_fee = Fixed8.FromDecimal(settings.FEE_PER_EXTRA_BYTE * (self.Size() - settings.MAX_FREE_TX_SIZE))
            if req_fee < settings.LOW_PRIORITY_THRESHOLD:
                req_fee = settings.LOW_PRIORITY_THRESHOLD
            if fee < req_fee:
                raise TXFeeError(f'The tx size ({self.Size()}) exceeds the max free tx size ({settings.MAX_FREE_TX_SIZE}).\nA network fee of {req_fee.ToString()} GAS is required.')
```
(ref https://github.com/jseagrave21/neo-python/blob/f75cb5dc58bedbfc12ea1a6afbca5332088f8ca6/neo/Core/TX/RawTransaction.py#L533-L541)

vncoelho:
I do know if this helps, but we have raw example (not organized) about advanced signing (cosigners) of invokes and also multi-sig in neocompiler eco.

I think it is worth a brief looking, my friend:

https://github.com/NeoResearch/neocompiler-eco/blob/master/public/js/eco-scripts/signatures_manager.js

https://github.com/NeoResearch/neocompiler-eco/blob/master/public/js/eco-scripts/rawMultiSigConversions.js

https://github.com/NeoResearch/neocompiler-eco/blob/master/public/js/eco-scripts/transfers_NeonJS.js

comountainclimber:
@jseagrave21 was able to implement the fee logic for `SimplePolicyPlugin` here: https://github.com/CityOfZion/neon-wallet/pull/1882/commits/593a1a3bdbf8deae1697887dae25274f30f22f80 thanks again for all your help here

vncoelho:
Sometime ago we thought about a RPC server just for sharing offline signatures (@igormcoelho).

Each user registers the multisig, the server receives the signed parts related to a specific hash. Those that have that hash registered in their wallet will receive the already signed part.
This would be an easy way to share signatures for smart accounts.

I believe there is not to worry much about security. This could be a good service offered by Light Wallet with an extra RPC server.

jseagrave21:
:tada:

Issue #1899, opened 2019-07-30
Error when sending tokens from a balance with more than 16 significant digits


hal0x2328:
**Wallet version or commit SHA:**
2.3.0

**Environment:**
OSX Mojave

**Prerequisites:**
Have greater than 10 million NEP-5 tokens with a 8-digit fraction (e.g. 10000000.00000001)

**Reproduction steps:**
1. Start Neon Wallet
2. Try to send any amount of the tokens where you have >10M in the wallet

Nothing happens in the GUI but the console log shows:
```
Uncaught (in promise) Error: gt() number type has more than 15 significant digits: 247716140.63257673
    at $ (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:26:85512)
    at new I (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:26:83742)
    at I.Y.greaterThan.Y.gt (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:26:91729)
    at t.validateTransactionBeforeSending (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:21:9886)
    at n (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:21:10163)
    at Array.find (native)
    at t.validateTransactionsBeforeSending (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:21:10175)
    at file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:55:232528
    at p (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:66:57807)
    at Generator._invoke (file:///Applications/Neon.app/Contents/Resources/app.asar/app/dist/bundle.js:66:57595)
```

dauTT:
Thank you @hal0x2328  for reporting! I will have a look at this issue.

dauTT:
@hal0x2328 ,


On Ubuntu, I did a transaction on test and didn't find have any problem the first time:

https://neoscan-testnet.io/transaction/F4B35EB82E2842E3122752D5726F6B33358DD527E024B30E9C8EB290E14F8249

![image](https://user-images.githubusercontent.com/30392990/62413784-551b5800-b613-11e9-8c6b-5ca40ea24567.png)

The second time, when I did the same transaction in the opposite direction I can't:

[...]

PR #1901, opened 2019-08-03
Fix sending transaction from an account with balance more than 15 significant digits.

dauTT:

#1899

When sending a transaction from an account with balance more than 15 significant digits, neon-wallet is unable to process the transaction because the balance can't be used as input to create a `bignumber` object. This is a limitation of the package bignumber.js version 5.0.0.

We did a workaround and used an alredy existing method, `toBigNumber`, to generate a bignumber object from an input whose type is a number and with more 15 significant digits.

Tested manually :
https://neoscan-testnet.io/transaction/e0d08eda33891e16700819a62ae83a04aa73a5bdd4daa24fa05ad2b1ee26f7b4
image

- [X] Unit tests written?

### DIFF
--- a/__tests__/core/wallet.test.js
+++ b/__tests__/core/wallet.test.js
@@ -1,0 +1,14 @@
+import { validateTransactionBeforeSending } from '../../app/core/wallet'
+
+describe('validateTransactionBeforeSending tests', () => {
+  test('balance of type number and with more than 15 significant digts', () => {
+    const sendEntry = {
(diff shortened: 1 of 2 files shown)


comountainclimber:
Nice job @dauTT 💪

comountainclimber:
Award @dauTT 175 points

lllwvlvwlll:
https://neoscan.io/transaction/d82da2040b11e60522f6d409ad8f96d6842f527feb13941ce2fa787997ca06f7

Issue #1903, opened 2019-08-13
bug: Gas claims should be disabled in "Watch" Mode


comountainclimber:
**Wallet version or commit SHA:**
2.3.0

**Environment:**
Any

**Reproduction steps:**
e.g.:
1. Start Neon Wallet
2. Log in using "watch mode" to any address that has held NEO (will have the ability to claim GAS)
3. Attempt to claim GAS

Because claiming GAS without the ability to sign the transaction is impossible it will always fail. The button should be disabled with a tooltip that reads "Gas claims are unavailable in Watch mode." Or something to that effect

**Screenshots/Video:**
![claim_bug](https://user-images.githubusercontent.com/13072035/62966668-b89c4700-bdc4-11e9-84ac-a5fc693d2aff.gif)

dauTT:
FYI, I am working on this task.

PR #1907, opened 2019-08-20
Fix gas claim bug in Watch mode

dauTT:

#1903 

Neon wallet user is able to click on the claim gas button in Watch mode login and this is not good as this feature is supposed to be applicable only to the user who owns the private key.

1) Disable the option to claim gas in Watch mode.
2) Provide an appropriate tooltip text to the user to inform him about point 1).

Screencast-from-21 08 2019-00_20_06

tested the code manually under Watch mode and Normal scenario. In the Normal scenario, I own the private key and I was able to claim gas. In the Watch Mode scenario, the claim button is disabled and a tooltip warning appeared informing the user of this fact.

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Claim/Claim.jsx
+++ b/app/containers/Claim/Claim.jsx
@@ -12,23 +12,20 @@ type Props = {
   doGasClaim: Function,
   disableClaimButton: boolean,
   claimAmount: string,
+  isWatchOnly?: boolean,
 }
 
 export default class Claim extends Component<Props> {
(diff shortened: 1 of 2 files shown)


comountainclimber:
Please award @dauTT 175 points

lllwvlvwlll:
https://neoscan.io/transaction/d82da2040b11e60522f6d409ad8f96d6842f527feb13941ce2fa787997ca06f7

PR #1912, opened 2019-09-09
[Feature] Split-key wallet import

hal0x2328:
None

Adds the ability to import a private key directly derived from combining two other private keys.

Added a new import form, and a corresponding controller to implement EC split-key derivation math.

Manual testing and added a test case

Added the user's public key to the dashboard view so it can be easily shared with a third party for multisig or split-wallet purposes. 

A couple of new dependencies were added to package.json but these were already child dependencies of other packages used by neon-wallet so there shouldn't be any increase in overall code size except for the lines of code in the PR. 

The node-sass package had to be updated to a newer version or the package would not build on my system (OSX Mojave)

- [X] Unit tests written?


comountainclimber:
@hal0x2328 If you run the command `yarn` it will update our lock file which will fix the broken CI build...

I am still reviewing the code here and will leave more feedback shortly

comountainclimber:
So I have reviewed the PR and run through the split key import locally... Awesome job @hal0x2328 this is a pretty neat feature!

I do have some suggestions in terms of enhancing the user experience. I think it would be beneficial to break the form for importing a split key into two separate views as well as some sort of additional explanation on the pages to inform users what is going on. There are simply too many inputs on this one page in my opinion:

CURRENTLY:
Screen Shot 2019-09-10 at 7 52 22 PM

PROPOSED:
split-key-1

split-key-2

There are numerous places in the app where we have a similar layout... Let me know your thoughts. I also mostly am just using the text there as placeholder but think it is valuable to have some sort of brief explanation

hal0x2328:
> @hal0x2328 This is looking really good... I am ready to merge contingent on one small defect I found (again related to UI) was hoping we could "fix" the button spacing on step 2 such that the buttons match other pages that have this same "two button" layout
> 
> CURRENTLY:
> Screen Shot 2019-09-16 at 9 25 06 AM
> 
> DESIRED:
> Screen Shot 2019-09-16 at 9 26 12 AM
> 
> Sorry to be a pain, I prefer to have all of these sorts of issues resolved before merging. Thanks

No problem at all, was a simple fix. I just wish I had noticed the other two-button layout before, would have saved me a lot of time fiddling with the CSS!

comountainclimber:
@hal0x2328 any chance you have bandwidth to get this last styling fix in? Otherwise I can make a PR against your fork with the changes

hal0x2328:
> @hal0x2328 any chance you have bandwidth to get this last styling fix in? Otherwise I can make a PR against your fork with the changes

I made that change in https://github.com/CityOfZion/neon-wallet/pull/1912/commits/c5cd1697e783c94006838ed6936ecaf194b91cf2 - was it insufficient?

comountainclimber:
Oops never saw that commit until just now... Thanks! 🎉

PR #1933, opened 2020-01-24
feature: Internationalization

comountainclimber:
https://github.com/CityOfZion/neon-wallet/issues/1935

Sets up the application to consume JSON translations...

* [X] Create list of translations in english to pass on to translators and for the application to consume

* [X] Create a wrapper component that can dynamically update the current language based on props from application state

* [X] Update the settings page with an input that allows users to chose desired language that persists to local storage and global state.

* [X] Update the application so that all english translations are rendered via react-intl and are consuming our JSON

* [X] Add JSON files for additional translations (see linked issue below with attached google doc)  

* [X]  Fix any UI defects introduced by new language support (overflowing content etc)

* [ ]  Translate all remaining untranslated strings and get finished versions of Spanish, Chinese and Portuguese


app/components/Root/IntlWrapper.jsx
```diff
@@ -0,0 +1,25 @@
...
+    const { children, lang } = this.props
+    return (
+      <IntlProvider locale="en" messages={english}>
+        {children}
+      </IntlProvider>
+    )
+  }
+}
+
+export default IntlWrapper
```
comountainclimber: The `IntlWrapper` component will allow us to inject the language desired into the application dynamically via props from our redux implementation. (Right now hard coded to use "en")

comountainclimber: NOTE: updated to consume language from global state and persisted settings

PR #1978, opened 2020-08-02
Fix french language translation

fvilers:

There's no issue, I just found the bug while using the software.

This fix the translation of the word french into french (Francés is wrong, should be Français).

By changing the translation to the right one.

It's just a string.

No.

No.

- [ ] Unit tests written?

No.


fvilers:
Looks like some tests are failing and I don't have access to CircleCI to see which ones. I forked the `dev` branch, I suppose there was already some failing tests. Let me know if I must do something else to fix this PR.

comountainclimber:
Thanks @fvilers !

comountainclimber (approved):

fvilers:
It's not much, but I'm glad to help.

PR #2002, opened 2020-10-27
fix wallet export/import

ixje:
None, I didn't create an issue

#1998 updated the electron version from ancient (`2.x`) to recent (`8.5.2`). With the new electron version the [showSaveDialog()](https://github.com/electron/electron/blob/v8.5.2/docs/api/dialog.md#dialogshowsavedialogbrowserwindow-options) no longer accepts a callback as argument, but returns a promise. As a result wallet exporting currently does nothing (meaning; the dialog shows, but the JSON file is never written to disk). 

-edit-
same story for `showOpenDialog`. One place immediately related place I found was in importing the wallet I just exported. Didn't work. Now it does

Update code to be based on promises

manually tested

no

we got this 😆 

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Settings/Settings.jsx
+++ b/app/containers/Settings/Settings.jsx
@@ -76,6 +76,34 @@ export const loadWalletRecovery = async (
   if (canceled || !filePaths) return
 
   const filepath = filePaths[0]
+  fs.readFile(filepath, 'utf-8', async (err, data) => {
+    if (err) {
+      showErrorNotification({
+        message: `An error occurred reading the file: ${err.message}`,
+      })
+      return
+    }
+    const walletData = JSON.parse(data)
+    const recoveryData = await recoverWallet(walletData).catch(err => {
+      showErrorNotification({
+        message: `An error occurred recovering wallet: ${err.message}`,
+      })
+    })
+
+    if (recoveryData) {
+      showSuccessNotification({ message: 'Recovery was successful.' })
+      setAccounts(recoveryData.accounts)
+    }
+  })
+}
+  showSuccessNotification: Object => any,
+  showErrorNotification: Object => any,
+  setAccounts: (Array<Object>) => any,
+) => {
+  const { canceled, filePaths } = await dialog.showOpenDialog()
+  if (canceled || !filePaths) return
+
+  const filepath = filePaths[0]
   fs.readFile(filepath, 'utf-8', (err, data) => {
     if (err) {
       showErrorNotification({
--- a/app/containers/Settings/Settings.jsx
+++ b/app/containers/Settings/Settings.jsx
@@ -67,37 +67,34 @@ type State = {
   selectedLanguage: Language,
 }
 
-export const loadWalletRecovery = (
+export const loadWalletRecovery = async (
   showSuccessNotification: Object => any,
   showErrorNotification: Object => any,
   setAccounts: (Array<Object>) => any,
 ) => {
-  dialog.showOpenDialog(fileNames => {
-    // fileNames is an array that contains all the selected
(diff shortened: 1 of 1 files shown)


comountainclimber:
Thanks @ixje there may be a few more of these subtle issues lingering... I am going to take a close look at your refactor later today. Thanks again 🙏

ixje:
I've seen some, but since I didn't know yet how to trigger those I kept the PR small.

app/containers/Settings/Settings.jsx
```diff
@@ -67,36 +67,42 @@ type State = {
...
         showErrorNotification({
-          message: `An error occurred reading the file: ${err.message}`,
+          message: `An error occurred recovering wallet: ${err.message}`,
         })
-        return
-      }
-      const walletData = JSON.parse(data)
+      })
+  })
+}
```
comountainclimber, line 99: Tiny refactor suggestion for readability:

```javascript

export const loadWalletRecovery = async (
  showSuccessNotification: Object => any,
  showErrorNotification: Object => any,
  setAccounts: (Array<Object>) => any,
) => {
  const { canceled, filePaths } = await dialog.showOpenDialog()
  if (canceled || !filePaths) return

  const filepath = filePaths[0]
  fs.readFile(filepath, 'utf-8', async (err, data) => {
    if (err) {
      showErrorNotification({
        message: `An error occurred reading the file: ${err.message}`,
      })
      return
    }
    const walletData = JSON.parse(data)
    const recoveryData = await recoverWallet(walletData).catch(err => {
      showErrorNotification({
        message: `An error occurred recovering wallet: ${err.message}`,
      })
    })

    if (recoveryData) {
      showSuccessNotification({ message: 'Recovery was successful.' })
      setAccounts(recoveryData.accounts)
    }
  })
}
```

ixje: Can you make it a code suggestion, so I can just accept/commit? It is the button left of `H B I` in the editor. Right now I'm trying to eye ball the differences

app/containers/Settings/Settings.jsx
```diff
@@ -67,36 +67,42 @@ type State = {
   selectedLanguage: Language,
 }
 
-export const loadWalletRecovery = (
+export const loadWalletRecovery = async (
```
comountainclimber, line 70:
```suggestion
export const loadWalletRecovery = async (
  showSuccessNotification: Object => any,
  showErrorNotification: Object => any,
  setAccounts: (Array<Object>) => any,
) => {
  const { canceled, filePaths } = await dialog.showOpenDialog()
  if (canceled || !filePaths) return

  const filepath = filePaths[0]
  fs.readFile(filepath, 'utf-8', async (err, data) => {
    if (err) {
      showErrorNotification({
        message: `An error occurred reading the file: ${err.message}`,
      })
      return
    }
    const walletData = JSON.parse(data)
    const recoveryData = await recoverWallet(walletData).catch(err => {
      showErrorNotification({
        message: `An error occurred recovering wallet: ${err.message}`,
      })
    })

    if (recoveryData) {
      showSuccessNotification({ message: 'Recovery was successful.' })
      setAccounts(recoveryData.accounts)
    }
  })
}
```

app/containers/Settings/Settings.jsx
ixje: Somehow Github failed to replace the old code with your new suggestion and instead inserted it. I added a commit that fixed it. Should be good to go now

comountainclimber:
Awesome! thanks again for taking the initiative to fix this

comountainclimber (approved):

PR #2127, opened 2021-05-06
feature: n3 support

comountainclimber:


app/modules/transactions.js
```diff
@@ -178,145 +179,230 @@ export const sendTransaction = ({
...
+    ? new Promise(async (resolve, reject) => {
+        try {
+          /*
+            TODO:
+              - Ability to send multiple assets in a single transaction
+              - Ability to attach custom fees
+              - Ledger support
+              - Support for test AND main net
+              - Node url should come from settings
+          */
```
comountainclimber, line 214: Each of these should be broken into their own issues ☝️

Issue #2137, opened 2021-06-23
[Contacts] Implement improvements to Contact/Address registration process so it properly validates and persists Legacy and N3 records


thiagocbalducci:
As of this moment, the system has no means of properly storing and differentiating Neo Legacy and Neo 3 contacts, even though we already have a version toggle available before logging in.

Our address validation process is also not prepared to validate Neo 3 addresses, currently validating only Neo Legacy addresses, no matter which value is set for our version toggle (Neo 3 or Neo Legacy).

We need to improve our Contact/Address registration process so that, depending on which value is currently set for the Neo Version Toggle it can validate and store those entities properly.

- If the version toggle is set to "Neo Legacy", contacts and addresses should be persisted in separated storage, dedicated to this Neo version. When adding an address, this address should be validated as per Neo Legacy standards.
- If the version toggle is set to "Neo 3", contacts and addresses should be persisted in separated storage, dedicated to this Neo version. When adding an address, this address should be validated as per Neo 3 standards.

All behaviors that currently exist in these processes should be kept as is (in terms of behavior and UI/UX), with exception of those changes mentioned in this issue.

Issue #2138, opened 2021-06-23
[Contacts] Implement improvements to Contact/Address listing so that it only presents records related to the Neo Version that is currently set in the app toggle


thiagocbalducci:
As of this moment, the system has no means of properly differentiating Neo Legacy and Neo 3 contacts and addresses, even though we already have a version toggle available before logging in. Because of this, the contact page is always listing Neo Legacy records, even if the toggle is set to Neo 3.

We need to improve our Contact/Address listing process so that, depending on which value is currently set for the Neo Version Toggle it can read and display only those records that are related to that version of Neo.

- If the version toggle is set to "Neo Legacy", only Neo Legacy contacts and addresses should be displayed.
- If the version toggle is set to "Neo 3", only Neo 3 contacts and addresses should be displayed.
- All behaviors that currently exist in these processes should be kept as is (in terms of behavior and UI/UX), with exception of those changes mentioned in this issue.

PR #2140, opened 2021-06-29
correction in contacts screen to know when it is neo 3 or neoLegacy

endkeyCoder:
…2138 #2137

- [ ] Unit tests written?

### DIFF
--- a/app/actions/contactsActions.js
+++ b/app/actions/contactsActions.js
@@ -12,8 +12,10 @@ export type TChains = { contactKey: string, chain: string }[]
 const STORAGE_KEY_CHAIN = 'chainBook'
 const STORAGE_KEY = 'addressBook'
 
-export const getChains = async (): Promise<TChains> =>
-  getStorage(STORAGE_KEY_CHAIN)
+export const getChains = async (): Promise<TChains> => {
+  const chains = await getStorage(STORAGE_KEY_CHAIN)
+  return Array.isArray(chains) ? chains : []
+}
 
 const setChainContact = async (contactKey: string, chain: string) => {
   const chains = await getChains()
@@ -42,6 +44,8 @@ const updateChainContact = async (
       chainContacts[chainContacts.indexOf(chainContactFound)].chain = chain
 
       await setStorage(STORAGE_KEY_CHAIN, chainContacts)
+    } else {
+      await setChainContact(contactKey, chain)
     }
   }
 }
@@ -72,7 +76,8 @@ type Contacts = {
   [name: string]: string,
 }
 
-const getContacts = async (): Promise<Contacts> => getStorage(STORAGE_KEY)
+export const getContacts = async (): Promise<Contacts> =>
+  getStorage(STORAGE_KEY)
 
 const setContacts = async (contacts: Contacts): Promise<any> =>
   setStorage(STORAGE_KEY, contacts)
--- a/app/actions/contactsActions.js
+++ b/app/actions/contactsActions.js
@@ -18,11 +18,12 @@ const STORAGE_KEY = 'addressBook'
 const setChainInContactsLessChain = async () => {
   const contacts: Contacts = await getStorage(STORAGE_KEY)
   const chains: TChains = await getStorage(STORAGE_KEY_CHAIN)
+  const contactChains = Array.isArray(chains) ? chains : []
   await Promise.all(
     Object.entries(contacts).map(async ([name]) => {
-      if (!chains.find(chain => chain.contactKey === name)) {
-        chains.push({ contactKey: name, chain: 'neo2' })
-        await setStorage(STORAGE_KEY_CHAIN, chains)
+      if (!contactChains.find(chain => chain.contactKey === name)) {
+        contactChains.push({ contactKey: name, chain: 'neo2' })
+        await setStorage(STORAGE_KEY_CHAIN, contactChains)
       }
     }),
   )
--- a/app/actions/contactsActions.js
+++ b/app/actions/contactsActions.js
@@ -5,11 +5,72 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
 
 import { getStorage, setStorage } from '../core/storage'
 
+export const ID = 'contacts'
+
+export type TChains = { contactKey: string, chain: string }[]
+
+const STORAGE_KEY_CHAIN = 'chainBook'
+const STORAGE_KEY = 'addressBook'
+
+export const getChains = async (): Promise<TChains> =>
+  getStorage(STORAGE_KEY_CHAIN)
+
+const setChainContact = async (contactKey: string, chain: string) => {
+  const chains = await getChains()
+  const chainContacts: TChains = Array.isArray(chains) ? chains : []
+  if (contactKey && chain) {
+    chainContacts.push({ contactKey, chain })
+    await setStorage(STORAGE_KEY_CHAIN, chainContacts)
+  }
+}
+
+const updateChainContact = async (
+  contactKey: string,
+  chain: string,
+  oldContactKey: string,
+) => {
+  const chains = await getChains()
+  const chainContacts: TChains = Array.isArray(chains) ? chains : []
+  if (contactKey && chain && oldContactKey) {
+    const chainContactFound = chainContacts.find(
+      chainContact => chainContact.contactKey === oldContactKey,
+    )
+    if (chainContactFound) {
+      chainContacts[
+        chainContacts.indexOf(chainContactFound)
+      ].contactKey = contactKey
+      chainContacts[chainContacts.indexOf(chainContactFound)].chain = chain
+
+      await setStorage(STORAGE_KEY_CHAIN, chainContacts)
+    }
+  }
+}
+
+const deleteChainContact = async (contactKey: string) => {
+  const chains = await getChains()
+  const chainContacts: TChains = Array.isArray(chains) ? chains : []
+  if (contactKey) {
+    const newChainContacts = chainContacts.filter(
+      chainContact => chainContact.contactKey !== contactKey,
+    )
+
+    await setStorage(STORAGE_KEY_CHAIN, newChainContacts)
+  }
+}
+
+export const getContactChainAction = createActions(
+  ID,
+  (): Promise<TChains> => getChains(),
+)
+
+export const cleanContacts = async () => {
+  await setStorage(STORAGE_KEY, {})
+  await setStorage(STORAGE_KEY_CHAIN, [])
+}
+
 type Contacts = {
   [name: string]: string,
 }
-
-const STORAGE_KEY = 'addressBook'
 
 const getContacts = async (): Promise<Contacts> => getStorage(STORAGE_KEY)
 
@@ -26,13 +87,17 @@ const validateContact = (name: string, address: string) => {
   }
 }
 
-export const ID = 'contacts'
-
 export const addContactActions = createActions(
   ID,
-  ({ name, address }: { name: string, address: string }) => async (): Promise<
-    Contacts,
-  > => {
+  ({
+    name,
+    address,
+    chain,
+  }: {
+    name: string,
+    address: string,
+    chain: string,
+  }) => async (): Promise<Contacts> => {
     validateContact(name, address)
 
     const contacts = await getContacts()
@@ -43,7 +108,7 @@ export const addContactActions = createActions(
 
     const newContacts = { ...contacts, [name]: address }
     await setContacts(newContacts)
-
+    await setChainContact(name, chain)
     return newContacts
   },
 )
@@ -54,10 +119,12 @@ export const updateContactActions = createActions(
     oldName,
     newName,
     newAddress,
+    chain,
   }: {
     oldName: string,
     newName: string,
     newAddress: string,
+    chain: string,
   }) => async (): Promise<Contacts> => {
     validateContact(newName, newAddress)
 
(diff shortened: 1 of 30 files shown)


endkeyCoder:
corrections to issues #2137 and #2138

app/components/Contacts/ContactsPanel/ContactsPanel.jsx
```diff
@@ -98,6 +119,14 @@ const getContactsInGroups = (
 export default class ContactsPanel extends React.Component<Props, State> {
   state = {
     sorting: SORTING_OPTIONS[0],
+    contactChains: [],
+  }
+
+  componentDidMount() {
```
comountainclimber, line 125: nitpick: I think something like -

```javascript
async componentDidMount() {
      const chains = await getChains()
      this.setState(prevState => ({ ...prevState, contactChains: chains }))
}
```
is a little cleaner

app/components/Contacts/ContactsPanel/ContactsPanel.jsx
```diff
@@ -67,15 +70,33 @@ const SORTING_OPTIONS = [
...
+  contactChains: TChains,
+  chain: string,
 ) => {
   /* $FlowFixMe */
-  const contactsArray: Array<Contacts> = Object.entries(contacts).map(
-    ([name, address]) => ({
-      name,
-      address,
-    }),
+  const contactsSpecifcChain = contactChains.filter(
```
comountainclimber, line 77: typo here and I think `const chainSpecificContacts` makes a little more sense

app/actions/contactsActions.js
```diff
@@ -5,12 +5,73 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
...
 
+export const ID = 'contacts'
+
+export type TChains = { contactKey: string, chain: string }[]
+
+const STORAGE_KEY_CHAIN = 'chainBook'
+const STORAGE_KEY = 'addressBook'
+
+export const getChains = async (): Promise<TChains> =>
+  getStorage(STORAGE_KEY_CHAIN)
```
comountainclimber, line 16: I dont think we should be relying on storage here but rather the value passed from the `withChainData` HOC

app/actions/contactsActions.js
```diff
@@ -5,12 +5,73 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
...
+
+const setChainContact = async (contactKey: string, chain: string) => {
```
comountainclimber, line 18: If we already have chain as a param why do we call `getChains` on the next line?

comountainclimber:
When spinning up this branch the contacts screen crashes the application:


Screen Shot 2021-06-29 at 10 32 11 AM
Screen Shot 2021-06-29 at 10 32 01 AM


I am also a little confused on the general approach with this solution... Why not just add a new key to storage and follow the exact same logic already in place switching based on chain? something like this:




```javascript
const STORAGE_KEY = 'addressBook'
const N3_STORAGE_KEY = 'n3AddressBook'

const getContacts = async (chain): Promise<Contacts> =>getStorage( chain === 'neo3' ? N3_STORAGE_KEY : STORAGE_KEY)
```

app/components/Contacts/ContactsPanel/ContactsPanel.jsx
```diff
@@ -20,6 +20,7 @@ import CopyToClipboard from '../../CopyToClipboard'
 import LogoWithStrikethrough from '../../LogoWithStrikethrough'
```
comountainclimber, line 20: I think ideally this file would actually not be updated at all in this PR 👀

endkeyCoder: I agree

endkeyCoder:
> When spinning up this branch the contacts screen crashes the application:
> 
> Screen Shot 2021-06-29 at 10 32 11 AM
> 
> Screen Shot 2021-06-29 at 10 32 01 AM
> 
> I am also a little confused on the general approach with this solution... Why not just add a new key to storage and follow the exact same logic already in place switching based on chain? something like this:
> 
> ```js
> const STORAGE_KEY = 'addressBook'
> const N3_STORAGE_KEY = 'n3AddressBook'
> 
> const getContacts = async (chain): Promise<Contacts> => getStorage(chain === 'neo3' ? N3_STORAGE_KEY : STORAGE_KEY)
> ```
> 
> (And then follow the exact same structure for all the functions/methods)

basically if you put conditionals inside the code to validate which chain should use, if at some point there is a need to implement new chains, there will be a need to put new conditionals to validate this, if you persist the chain in storage and use a "filter" function to display your data, no matter how many are added, the behavior of the code remains the same

I'll see what's going on with the contact rendering logic, but I think this is the best approach

comountainclimber:
Ok makes sense! With the new data structure will be relatively simple to keep the same code and support new chains. Lets get that crashing bug sorted out and then we can merge

endkeyCoder:
> Ok makes sense! With the new data structure will be relatively simple to keep the same code and support new chains. Lets get that crashing bug sorted out and then we can merge

This is amazing! Let with me, I'll fix it

endkeyCoder:
@comountainclimber everything has been changed according to your suggestions, as soon as you can let me know if you agree or need to make further changes

comountainclimber:
When I login using an n3 address I am seeing all of the contacts from my neo legacy contact list we need to fix this before merging... To reproduce login with a neo2 address, create a contact and then login with an n3 address

comountainclimber (changes requested):
See comment above about contact list chain issue

endkeyCoder:
> its no longer crashing but when I login using an n3 address I am seeing all of the contacts from my neo legacy contact list we need to fix this before merging... To reproduce login with a neo2 address, create a contact and then login with an n3 address

you can see why these contacts were created in a version that didn't separate the strings, if you edit the contact and save it persists that contact in the current configuration, n3 or neo legacy, you want me to do a function that when starting the app put all pre-existing contacts as neo legacy?

comountainclimber:
☝️  You should not be able to see Neo legacy contacts when logging in with an n3 address (and also the opposite should be true)

endkeyCoder:
> ☝️ You should not be able to see Neo legacy contacts when logging in with an n3 address (and also the opposite should be true)

ok, so should I create a function that check the contacts created in back versions and set this as n3 or neo legacy depending on address?

comountainclimber:
Not sure... there are probably many different ways to solve this but I would do your best to leave the legacy code functioning as it does currently so not to introduce any regressions

endkeyCoder:
> Not sure... there are probably many different ways to solve this but I would do your best to leave the legacy code functioning as it does currently so not to introduce any regressions

from what you told me, this should solve it, so the pre-existing contacts will be validated in the same way as the new ones.

I'll do it and you test it and tell me if it's as expected, ok?

app/actions/contactsActions.js
```diff
@@ -5,13 +5,79 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
...
+    chainContacts.push({ contactKey, chain })
+    await setStorage(STORAGE_KEY_CHAIN, chainContacts)
+  }
+}
+
+const updateChainContact = async (
+  contactKey: string,
+  chain: string,
+  oldContactKey: string,
+) => {
```
comountainclimber, line 33: Where is the validation of n3 addresses here?

comountainclimber:
I think we may be overcomplicating things a little bit but either way n3 addresses need the same validation and flow we are currently applying to neo legacy addresses... 

The goal here is to recreate the exact functionality that exists in the neo legacy contacts screen

app/actions/contactsActions.js
```diff
@@ -5,13 +5,79 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
 
 import { getStorage, setStorage } from '../core/storage'
 
+export const ID = 'contacts'
+
+export type TChains = { contactKey: string, chain: string }[]
```
comountainclimber, line 10: Dont we need name and address in here too 🤔

endkeyCoder: don't is necessary, the addressBook storage just is a "auxiliary table"

app/actions/contactsActions.js
endkeyCoder: it's in chain parameter

endkeyCoder:
> I think we may be overcomplicating things a little bit but either way n3 addresses need the same validation and flow we are currently applying to neo legacy addresses...
> 
> The goal here is to recreate the exact functionality that exists in the neo legacy contacts screen

I understand your thought, but I think that this is a best aprroach to future maintenance

comountainclimber:
The latest issue I can see:
Screen Shot 2021-07-01 at 9 15 48 AM

lllwvlvwlll:
@endkeyCoder this is meant to be a short term implementation so we can ship N3 support quickly. The contacts list behavior will be very different in the next major release. It will align more closely with the mobile wallet approach.

comountainclimber:
👆 definitely... There is actually no need to add any additional code/logic other than the bare minimum needed to support N3 I am worried about making things overly complex here when the changes necessary could be substantially smaller than what this PR is currently doing

comountainclimber:
For more context, I was expecting something along these lines:
```javascript
// @flow
import { createActions } from 'spunky'
import { wallet } from '@cityofzion/neon-js'
import { wallet as n3Wallet } from '@cityofzion/neon-js-next'
import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'

import { getStorage, setStorage } from '../core/storage'
import { getSettings } from './settingsActions'

type Contacts = {
  [name: string]: string,
}

const STORAGE_KEY = 'addressBook'
const N3_STORAGE_KEY = 'n3AddressBook'

const getContacts = async (chain: string): Promise<Contacts> =>
  chain === 'neo3' ? getStorage(N3_STORAGE_KEY) : getStorage(STORAGE_KEY)

const setContacts = async (contacts: Contacts, chain: string): Promise<any> =>
  setStorage(chain === 'neo3' ? N3_STORAGE_KEY : STORAGE_KEY, contacts)

const validateContact = (name: string, address: string, chain: string) => {
  if (isEmpty(name)) {
    throw new Error('Name cannot be empty.')
  }
  if (
    chain === 'neo3' ? !n3Wallet.isAddress(address) : !wallet.isAddress(address)
  ) {
    throw new Error(`Invalid address ${address}.`)
  }
}

export const ID = 'contacts'
export const addContactActions = createActions(
  ID,
  ({
    name,
    address,
    chain,
  }: {
    name: string,
    address: string,
    chain: string,
  }) => async (): Promise<Contacts> => {
    validateContact(name, address, chain)

    const contacts = await getContacts(chain)

    if (has(contacts, name)) {
      throw new Error(`Contact "${name}" already exists.`)
    }

    const newContacts = { ...contacts, [name]: address }
    await setContacts(newContacts, chain)

    return newContacts
  },
)

export const updateContactActions = createActions(
  ID,
  ({
    oldName,
    newName,
    newAddress,
    chain,
  }: {
    oldName: string,
    newName: string,
    newAddress: string,
    chain: string,
  }) => async (): Promise<Contacts> => {
    validateContact(newName, newAddress, chain)

    const contacts = await getContacts(chain)
    const names = keys(contacts)
    const addresses = values(contacts)
    const index = indexOf(names, oldName)

    if (index === -1) {
      throw new Error(`Contact "${oldName}" does not exist.`)
    }

    const newContacts = zipObject(
      [...names.slice(0, index), newName, ...names.slice(index + 1)],
      [...addresses.slice(0, index), newAddress, ...addresses.slice(index + 1)],
    )
    await setContacts(newContacts, chain)

    return newContacts
  },
)

export const deleteContactActions = createActions(
  ID,
  ({ name, chain }: { name: string, chain: string }) => async (): Promise<
    Contacts,
  > => {
    const contacts = await getContacts(chain)

    if (!has(contacts, name)) {
      throw new Error(`Contact "${name}" does not exist.`)
    }

    const newContacts = omit(contacts, name)
    await setContacts(newContacts, chain)

    return newContacts
  },
)

export default createActions(ID, () => async (): Promise<Contacts> => {
  const settings = await getSettings()
  const { chain } = settings
  return getContacts(chain)
})
```

app/actions/contactsActions.js
```diff
@@ -5,13 +5,93 @@ import { has, isEmpty, keys, values, indexOf, zipObject, omit } from 'lodash-es'
...
+  (): Promise<TChains> => getChains(),
+)
+
+export const cleanContacts = async () => {
+  await setStorage(STORAGE_KEY, {})
+  await setStorage(STORAGE_KEY_CHAIN, [])
+}
+
+export const getContacts = async (): Promise<Contacts> =>
+  getStorage(STORAGE_KEY)
```
comountainclimber, line 94: without chain data this is always going to return the neo legacy contacts 👆

comountainclimber:
https://github.com/CityOfZion/neon-wallet/pull/2140#discussion_r662422431 this last bug needs to be fixed before we can merge

app/actions/contactsActions.js
endkeyCoder: about it function, you prefer to return already filtered by chain contacts, is that it? because in contact screen I make the filter and show only contacts that own a specific chain

comountainclimber: We will want the redux state tree to always have the right contacts in place so we need to filter one level higher, I am afraid filtering at the component level will lead to more work and unexpected bugs

endkeyCoder: ok, I understood, so I gonna the filter in high level

comountainclimber:
@endkeyCoder @melanke in the future *lets always avoid* merging broken commits into `dev` `master` any other working branches

PR #2141, opened 2021-07-01
feature: n3 contacts support

comountainclimber:
ezgif com-gif-maker


app/components/Modals/ShowQrForExportModal/ShowQrForExportModal.scss
```diff
@@ -17,10 +17,6 @@ canvas {
   min-height: 250px;
 }
 
-button {
-  margin-top: 12px;
-}
-
```
comountainclimber, line 23: Not quite sure why but this was breaking layout on the Send screens

PR #2153, opened 2021-07-20
feature: n3 support

comountainclimber:

- [ ] Unit tests written?


app/core/formatters.js
```diff
@@ -24,6 +24,15 @@ export const formatToRoundedShortNumber = (value: number): string => {
     .toString()
 }
 
+export const convertToArbitraryDecimals = (
+  num: number,
+  decimals: number = 8,
+): string => {
+  // eslint-disable-next-line
+  const multiplier = 1 / Math.pow(10, decimals)
+  return (num * multiplier).toFixed(decimals)
```
mhuggins, line 33: I'd suggest using the bignumber.js package (already part of this repo) for this math to avoid any loss of precision.  There are some helpful functions in [math.js](https://github.com/CityOfZion/neon-wallet/blob/aef2ec4d59a6f7f1590d30a94974b6af3d92e982/app/core/math.js) that could be used here.

```js
export const convertToArbitraryDecimals = (num: number, decimals: number = 8): string =>
  toBigNumber(num)
    .mul(multiplier)
    .div(Math.pow(10, decimals))
    .toFixed(decimals)
```

app/modules/claim.js
```diff
@@ -117,6 +125,58 @@ const getUpdatedClaimableAmount = async ({
...
+  let endpoint = await getNode(net)
+  if (!endpoint) {
+    endpoint = await getRPCEndpoint(net)
+  }
+
+  const CONFIG = {
+    account: FROM_ACCOUNT,
+    rpcAddress: endpoint,
+    // TODO: this will have to by dynamic based on test/mainnets
+    networkMagic: 844378958,
```
mhuggins, line 149: Is this the correct mainnet number?  Just making sure this isn't a missed TODO prior to mainnet usage.

app/containers/Send/Send.jsx
```diff
@@ -191,21 +208,97 @@ export default class Send extends React.Component<Props, State> {
...
-      const totalSendableAssets = toBigNumber(sendableAssets[asset].balance)
+      let totalSendableAssets = toBigNumber(sendableAssets[asset].balance)
+
+      if (asset === 'GAS') {
+        const existingGasAmounts =
+          Number(this.calculateRowAmounts(asset, index)) - MIN_EXPECTED_GAS_FEE
+
+        totalSendableAssets = minusNumber(
+          totalSendableAssets,
+          MIN_EXPECTED_GAS_FEE * rowsWithAsset.length,
```
mhuggins, line 287: With this being a floating point number, I'd suggest using `multiplyNumber` here & storing `MIN_EXPECTED_GAS_FEE` as a string value instead of a floating point number.

```js
const MIN_EXPECTED_GAS_FEE = '0.072'

totalSendableAssets = minusNumber(
  totalSendableAssets,
  multiplyNumber(MIN_EXPECTED_GAS_FEE, rowsWithAsset.length)
)
```

Issue #2147, opened 2021-07-06
NW - [N3] - Ledger Integration


lllwvlvwlll:
This enhancement implements Ledger Nano S and X support for Neon desktop using a similar workflow as with Neo Legacy.  This feature is blocked by the Ledger Support initiative which is currently in-progress.

## Terms:
* Supported Hardware Wallets - Refers to the Ledger Nano S and Ledger Nano X hardware wallets
* Application - The Neon Wallet Desktop application

## Feature User Needs:
* UN01 - The user shall have the ability to authenticate into a Neo N3 account using the supported hardware wallets.
* UN02 - The authentication interface shall provide users with feedback on their authentication status.
* UN03 - The authentication interface shall provide users with their address upon application interfacing with the hardware wallet.
* UN04 - The authentication interface shall provide users with the option to create an additional address derived from the hardware wallet.
* UN05 - The application shall allow users to successfully sign and relay transactions for publishing on the Neo N3 blockchain.

### Authentication:
  <b>Test Begin State:</b> New ledger configuration without assets

 * [ ] 1. Navigate to the N3 side of the application and select "ledger" for the login method
 * [ ] 2. The "Connect and unlock your ledger device" has a spinner and login is disabled
 * [ ] 3. On the ledger, unlock the device with your pin
 * [ ] 4. On Neon Wallet, the authentication workflow should progress to the next step with a spinner
 * [ ] 5. Open the "Neo N3" app on the ledger device
 * [ ] 6. The "public address" field on Neon should be populated with an address
 * [ ] 7. Record the address
 * [ ] 8. Clicking the "dropdown chevron" presents the user with a single address
 * [ ] 9. Click the "fetch additional addresses" icon
 * [ ] 10. When clicking the dropdown chevron, the user is presented with a list of address options
 * [ ] 11. The login button is clickable
 * [ ] 12. Click the login button
 * [ ] 13. The user is taken into the application and the same address which was logged is presented to the user

### Transaction Signing
  <b>Test Begin State:</b>: Ledger Wallet using the same bip39 mnemonic as referenced in the Authentication test with N3 application installed and a non-zero GAS balance on the address logged in the previous test case.

 * [ ] 1. Navigate to the N3 side of the application and select "ledger" for the login method
 * [ ] 2. The "Connect and unlock your ledger device" has a spinner and login is disabled
 * [ ] 3. On the ledger, unlock the device with your pin
[...]

PR #2215, opened 2021-09-01
Ledger n3

ixje:
https://github.com/CityOfZion/neon-wallet/issues/2147

not having n3 ledger support

add support

I haven't yet. Only parts are tested

no

- [ ] Unit tests written?

### DIFF
--- a/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
+++ b/app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react'
 import { wallet } from '@cityofzion/neon-js'
+import { wallet as n3Wallet } from '@cityofzion/neon-js-next'
 import { progressValues } from 'spunky'
 import classNames from 'classnames'
 import { FormattedMessage, FormattedHTMLMessage } from 'react-intl'
@@ -13,6 +14,10 @@ import ConfirmIcon from '../../assets/icons/confirm.svg'
 import RefreshIcon from '../../assets/icons/refresh.svg'
 import styles from '../Home/Home.scss'
 import { getPublicKeys, MESSAGES } from '../../ledger/neonLedger'
+import {
+  getPublicKeys as n3GetPublicKeys,
+  MESSAGES as N3MESSAGES,
+} from '../../ledger/n3NeonLedger'
 import DialogueBox from '../../components/DialogueBox'
 import WarningIcon from '../../assets/icons/warning.svg'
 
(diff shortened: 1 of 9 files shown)


comountainclimber:
The build is failing because `package.json` was updated _but_ the lock file was not updated... If you run `yarn` it should automatically update the lock file and fix the build

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -278,7 +281,11 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
             {this.getStatusIcon(OPEN_APP)}
           </div>
           <div className={styles.ledgerStageText}>
-            <FormattedHTMLMessage id="auth.ledger.navigateToNeoApp" />
+            {chain === 'neo2' ? (
+              <FormattedHTMLMessage id="auth.ledger.navigateToNeoApp" />
+            ) : (
+              <FormattedHTMLMessage id="auth.ledger.navigateToNeo3App" />
```
comountainclimber: Can we just use the string `'Navigate to the <strong>NEO3 app</strong> on your device'` here - reason being is that without the other translation files updated users will just see `auth.ledger.navigateToNeo3App` when not using English

ixje:
@comountainclimber Known issue(s):
* open Ledger tab in N2 mode, open the app such that we get an address, now switch to N3 mode. The data will not be refreshed. Same happens if you start with N3 and go back to N2.
* open ledger tab in N3 mode, it will say "Connect and unlock your Ledger device", do that. It will stay on this message until you actually do the second step (open the N3 app). I don't know where to look for the trigger to go to the next step.

comountainclimber:
If you revert these changes then the snapshot tests will pass => https://github.com/CityOfZion/neon-wallet/pull/2215/files#diff-5dcaf0cf2e58b16f97b0d3730d6174b58883b456eaf922f936843ffa5c0e8b93R16

I am still trying to figure out what the other issue breaking the test suite is

ixje:
champ!

comountainclimber (approved):

app/containers/LoginLedgerNanoS/LoginLedgerNanoS.jsx
```diff
@@ -278,7 +281,7 @@ export default class LoginLedgerNanoS extends React.Component<Props, State> {
             {this.getStatusIcon(OPEN_APP)}
           </div>
           <div className={styles.ledgerStageText}>
-            <FormattedHTMLMessage id="auth.ledger.navigateToNeoApp" />
+            Navigate to the <strong>NEO3 app</strong> on your device
```
comountainclimber, line 284: Now this component always says NEO3 app 😞

comountainclimber: Fixed with https://github.com/CityOfZion/neon-wallet/commit/e2a3773d150ffd8fff635fb38c4a17f77e334d47

PR #2265, opened 2021-09-22
resolve neoscan gas claim issue

lllwvlvwlll:
This PR resolves the GAS claim issue correlated to the GAS generation rate change on Neo Legacy.  It has been verified against a NEP2 and Ledger wallet.


comountainclimber:
Looks good to me lets fix the linting errors and then we can merge

comountainclimber (approved):

PR #2266, opened 2021-09-24
update polling approach for claims

lllwvlvwlll:
This PR updates the claim polling mechanism to resolve issues associated with GAS claim on Neo Legacy.  Note that the node is hardcoded to use a COZ node for the activity.

### DIFF
--- a/app/modules/claim.js
+++ b/app/modules/claim.js
@@ -1,6 +1,6 @@
 // @flow
 import { api, type Claims } from '@cityofzion/neon-js'
-import { api as apiLatest } from '@cityofzion/neon-js-legacy-latest'
+import { api as apiLatest, rpc } from '@cityofzion/neon-js-legacy-latest'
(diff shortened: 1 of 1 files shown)


app/modules/claim.js
```diff
@@ -31,7 +31,7 @@ import { getNode, getRPCEndpoint } from '../actions/nodeStorageActions'
 // Constants
 export const DISABLE_CLAIM = 'DISABLE_CLAIM'
 const POLL_ATTEMPTS = 30
-const POLL_FREQUENCY = 10000
+const POLL_FREQUENCY = 4000
```
comountainclimber, line 34: Why do we need to double the polling frequency?

lllwvlvwlll: Arbitrary UX enhancement.

comountainclimber (approved):

PR #2304, opened 2021-11-02
Feat/2256

lllwvlvwlll:
## General
* Updates the address history features to use dora  via `@cityofzion/dora-ts`
* Enhances the activity tab to use the new txfull endpoint and include all contract invocations to support walletconnect
* Updates the history export feature to work with N3


lllwvlvwlll:
@comountainclimber This is ready to review.  There is also an outstanding defect associated with the new svg elements and the themes.

comountainclimber:
The build is failing because of this error `error Your lockfile needs to be updated,` looks like you need to run the command `yarn` on this branch

app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -48,12 +51,14 @@ export default class Transaction extends React.Component<Props> {
...
       tx: { type },
+      chain,
       className,
       isPending,
     } = this.props
     return (
       <div className={classNames(styles.transactionContainer, className)}>
-        {this.renderAbstract(type)}
+        {chain !== 'neo3' && this.renderAbstract(type)}
+        {chain === 'neo3' && this.renderAbstractN3()}
```
comountainclimber: nitpick: Probably would be better to just have ternary here: `chain === 'neo3 ?  this.renderAbstractN3() : this.renderAbstract(type)`

app/components/Blockchain/Transaction/Transaction.jsx
```diff
@@ -88,7 +93,15 @@ export default class Transaction extends React.Component<Props> {
...
     const { networkId, explorer, tx, chain } = this.props
-    const { txid } = tx
+    let txid
+    switch (chain) {
+      case 'neo3':
+        txid = tx.hash.substring(2)
+        break
+      default:
+        ;({ txid } = tx)
+    }
```
comountainclimber, line 102: This reads and looks a little strange maybe we could just have:
```javascript
  let txid = tx.txid
  if (chain === 'neo3') txid = tx.hash.substring(2)
```

comountainclimber (changes requested):
Other than a few small suggestions looks good to me nice work @lllwvlvwlll

app/components/Blockchain/Transaction/Transaction.jsx
lllwvlvwlll: resolved in 
eac430b

app/components/Blockchain/Transaction/Transaction.jsx
lllwvlvwlll: resolved in 
8063583

comountainclimber:
I _think_ this is still a problem => https://github.com/CityOfZion/neon-wallet/pull/2304#issuecomment-960282558

If running `yarn` does not resolve it just let me know

app/components/Blockchain/Transaction/N3ContractInvocationAbstract.jsx
```diff
@@ -0,0 +1,37 @@
+// @flow
+import React from 'react'
+import classNames from 'classnames'
+import { injectIntl } from 'react-intl'
+import styles from './Transaction.scss'
+import InvokeIcon from '../../../assets/icons/raw-invoke-tx.svg'
+
+type Props = {
+  contract_name: string,
```
comountainclimber, line 9: Looks like eslint is failing here and below:

```
root/neon-wallet/app/components/Blockchain/Transaction/N3ContractInvocationAbstract.jsx
  17:13  error  Identifier 'contract_name' is not in camel case  camelcase
  30:12  error  Identifier 'contract_name' is not in camel case  camelcase
  ```

lllwvlvwlll: This is a raw field name coming straight out of the `dora` api.  I've set the linter to ignore it instead of changing the name in the action for simplicity sake.

comountainclimber (approved):

PR #2314, opened 2021-11-08
closes #2285

### LINKED ISSUE #2285: Neon cannot send total amount of GAS
There is a calculation issue when determining the maximum amount of GAS that can be sent after subtracting the required fee. The "Max" button will always show an amount plus fee less 0.06078673 GAS (and you can't manually override this maximum amount). So when the user sends the transaction after hitting the "Max" button they will always be left with 0.06078673 GAS in the wallet that cannot be transferred.

lllwvlvwlll:
Updates `Max` button behavior for N3 to calculate the max gas using the explicit transaction fee.

### DIFF
--- a/app/containers/Send/Send.jsx
+++ b/app/containers/Send/Send.jsx
@@ -268,8 +268,6 @@ export default class Send extends React.Component<Props, State> {
   calculateMaxValue = (asset: string, index: number = 0) => {
     const { sendableAssets, chain } = this.props
 
-    const MIN_EXPECTED_GAS_FEE = 0.072
-
     if (chain === 'neo2') {
       if (sendableAssets[asset]) {
         const rows = [...this.state.sendRowDetails]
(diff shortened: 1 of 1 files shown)


comountainclimber:
When I pulled down this branch it did not seem to be working correctly - I needed to click the MAX button twice...  Pretty sure that is because `this.state.n3Fees.x` is `undefined` the first time

comountainclimber:
Screen Shot 2021-11-08 at 3 01 11 PM


The first time ☝️  and after clicking it again (the correct amount) 👇

Screen Shot 2021-11-08 at 3 01 15 PM

comountainclimber (changes requested):
See other comments

lllwvlvwlll:
We can try to figure out a workaround.  If a user does not populate a value or address first, the transaction fee is not correctly calculated because its not a valid invocation.  

I have another version locally where the calculate max method will deep copy and autofill the `bad` fields with mock values in order to calculate the correct fees, but that operation is asynchronous and the parent method is not so it introduces a lot of overhead.  I can look at making this approach more graceful and agree with you that it is not clear what is happening.

comountainclimber:
One possible solution is to delay rendering the screen when the container mounts with a loading indicator until we have performed the gas fee check (maybe we just use a default address OR the current users address) after that we should be able to appropriately saturate the gas fee object in state which would prevent the "double click" bug

PR #2322, opened 2021-11-15
Stale issues workflow

ixje:
As discussed on Slack, this adds a Github bot to deal with stale issues. Unlike what I said on Slack this version does not seem to be able to exclude issues created by members, but only by labels. I would argue that's actually not a bad thing.

Currently it will run once every day. It will exclude issues labelled with: Bug,Blocked,Blocker,Design,enhancement
so we have to make sure that we apply any of those if we wish to avoid being reminded on an issues going stale.

There is a rate limit of 30 issues per day it can close (something todo with github APIs). This means it will initially take a couple of days to go through all open issues once merged.

### DIFF
--- a/.github/workflows/stale-issue-reminder-and-closing.yml
+++ b/.github/workflows/stale-issue-reminder-and-closing.yml
@@ -1,0 +1,23 @@
+name: Handle stale issues
+on:
+  schedule:
+    - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


comountainclimber:
Thanks @ixje !!

comountainclimber (approved):

PR #2331, opened 2021-11-22
add GHOST logos

vincentgeneste:

Fixing that: 

Screen Shot 2021-11-21 at 7 14 58 PM

adding logos


app/assets/nep5/svg/index.js
```diff
@@ -47,6 +47,7 @@ import WWB from './WWB.svg'
 import XQTA from './XQTA.svg'
 import ZPT from './ZPT.svg'
 import NOS from './NOS.svg'
+import GHOST from './GHOST.svg'
 
 export const imageMap = {
   NEO,
```
comountainclimber: If you are not also exporting the asset here this code is noop... Have you verified that this PR works? If it does I am super confused as to how its working

comountainclimber (changes requested):
See comment above

app/assets/nep5/svg/index.js
vincentgeneste: haerm, my bad it's a miss on the PR, will fix now.

comountainclimber (approved):

PR #2359, opened 2022-01-04
add GM logo

vincentgeneste:

adding missing NEP17 logo


comountainclimber (approved):

comountainclimber:
@vincentgeneste There are now two different `GHOST` logos... Which one should we use in the app?

vincentgeneste:
> @vincentgeneste There are now two different `GHOST` logos... Which one should we use in the app?

Those are different. GHOST is our NEP11, GM is our NEP17. Both have been added :)

PR #2366, opened 2022-01-18
updates the walletconnect interface to use rpc nodes for the abi

### LINKED ISSUE #2364: WalletConnect: Problems calling `invokeFunction`
## Description
This problem was reported for more than one person, after connecting the dapp with the wallet, when trying to make an `invokeFunction` the NeonWallet fails, throwing the following error:
> Objects are not valid as a React child (found: object with keys {script, state, gasconsumed, exception, stack, isTest}). If you meant to render a collection of children, use an array instead.

PS.: This error doesn't occur on Aero

## Invocation Code
unknown

## WcSdk version:
```
"@cityofzion/wallet-connect-sdk-core": "^1.6.0",
"@walletconnect/client": "^2.0.0-beta.17",
"@walletconnect/qrcode-modal": "^2.0.0-alpha.20",
"@walletconnect/types": "^2.0.0-beta.17",
"@walletconnect/utils": "^2.0.0-beta.17",
```

lllwvlvwlll:
resolves #2364

This enhancement transitions the wallet connect interface from dora to the node rpc method `getcontractstate` for ABI requests. In doing so, we further decentralize the wallet and also ensure that the manifest is always representative of the smart contract being invoked.

### DIFF
--- a/app/containers/ConnectDapp/ConnectDapp.jsx
+++ b/app/containers/ConnectDapp/ConnectDapp.jsx
@@ -1,7 +1,7 @@
 // @flow
 import React, { useEffect, useState } from 'react'
 import classNames from 'classnames'
-import { wallet } from '@cityofzion/neon-js-next'
+import { wallet, rpc } from '@cityofzion/neon-js-next'
 import axios from 'axios'
 
 import CloseButton from '../../components/CloseButton'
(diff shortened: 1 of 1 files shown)


comountainclimber (approved):
Nice work @lllwvlvwlll looks good to me 👍

PR #2373, opened 2022-01-25
feature: fix network switch logic

comountainclimber:
This PR fixes a bug that required the user to refresh the dashboard after switching from Mainnet => testnet (or vice versa). The update will now occur if and when the user toggles the network switch

### DIFF
--- a/app/containers/App/Sidebar/NetworkSwitch/index.js
+++ b/app/containers/App/Sidebar/NetworkSwitch/index.js
@@ -1,6 +1,6 @@
 // @flow
 import { compose } from 'recompose'
-import { withActions, type Actions } from 'spunky'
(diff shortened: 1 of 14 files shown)


app/containers/App/Sidebar/NetworkSwitch/index.js
```diff
@@ -15,11 +15,12 @@ const mapActionsToProps = (actions: Actions): Object => ({
 })
 
 const mapAccountActionsToProps = (actions, props) => ({
-  loadWalletData: () =>
+  loadWalletData: (net: string) =>
```
comountainclimber, line 18: Possibly a bug in spunky but even though the state tree was being updated succesfully the actions below were being called with the previous state 😅 this change maps the change event from the select input to the invocation of these actions

Issue #2404, opened 2022-03-30
Use Dora as the default chain explorer instead of ancient Neoscan


gsmachado:
**Wallet version or commit SHA:**
2.12.7

**Environment:**
MacOS 12.3

**Prerequisites:**
None.

**Reproduction steps:**
1. Start Neon Wallet
2. "Log in" to your wallet, either a saved one or using Ledger -- it doesn't matter
3. On the top, where the address is shown, click on the address
4. A browser window should open pointing to the Neoscan explorer

The problem here is: Neoscan is ancient and should be replaced by Dora as the **default** chain explorer.

This is a super simple fix and should be done as soon as possible for usability.

lllwvlvwlll:
This is most likely a result of user settings.  I am under the impression that new installations configure the default explorer to dora. @comountainclimber can you confirm this?

kinmya:
Confirmed that it opens to https://dora.coz.io/address/neo3/mainnet/<address>/assets.

Issue #2402, opened 2022-03-19
QR code reader not loading address codes


hal0x2328:
**Wallet version or commit SHA:**
2.12.7

**Environment:**
MacOS 12.1

**Reproduction steps:**
e.g.:
1. Start Neon Wallet
2. Log in
3. Click on "Send" tab
4. Click on "Enter QR code" button
5. Try to scan a QR code from the Neon Wallet Mobile app

The code does not get processed - the camera periodically attempts to take pictures but never successfully reads an address from the code.

kinmya:
Scan of QR code on the desktop app successful when using it as the "RECIPIENT ADDRESS".  Not functional if the user sent a Receive QR Code from mobile wallet (utilizing the "Request Tokens" function).

kinmya:
Request token function option now working as expected.  v1.0.29-2i

Issue #2397, opened 2022-03-02
Cannot recover N3 wallet using json file.


cypher619:
Just installed wallet version 2.12.6 after clean install of Windows 10.  When trying to recover my N3 wallet with json recovery file, I can select a file but nothing happens after that.  My json file was greated with an 2.10.0 of the wallet back on 11/05/21 when I did my N3 conversion.

cypher619:
same issue as #2334 which has been closed due to inactivity.

cypher619:
bump so it doesn't die due to inactivity

cypher619:
bump so it doesn't die due to inactivity

kinmya:
I don't have Windows 10 access, but this functioned as expected on mac version of app.

Issue #2396, opened 2022-02-28
QR login to N3 results in Legacy address


hal0x2328:
**Wallet version or commit SHA:**
2.12.5

**Environment:**
Mac OS 12.1

**Prerequisites:**
A WIF private key saved as QR code

**Reproduction steps:**
1. Start Neon Wallet
2. Switch Login toggle to "N3"
3. Click on "Private" tab
4. Click on "Scan QR" button
5. Show QR code to camera

After the QR code is read and the wallet is logged in, Neon shows the N3 logo, but with a legacy address

**Screenshots/Video:**
https://www.dropbox.com/s/te2nutlep22r258/Screen%20Shot%202022-02-28%20at%202.00.02%20PM.png?dl=0

hal0x2328:
A follow-up comment from the original reporting user: _I noticed another issue with the QR reader. It's not working properly with other functions either. Send and receive will not populate the fields. Looks like the only way right now is to manually enter the keys every single time._

kinmya:
4 different QR codes with private key all show as invalid. "Invalid private key entered".

kinmya:
Working as expected.  The private key QR scan resulted in expected result of N3 wallet labeled as such in the app.

PR #2409, opened 2022-04-29
feature: fixes qr features and wallet recovery

comountainclimber:
Numerous defects listed below:

https://github.com/CityOfZion/neon-wallet/issues/2404

https://github.com/CityOfZion/neon-wallet/issues/2402

https://github.com/CityOfZion/neon-wallet/issues/2397

- ☝️ this one is slightly tricky to test... Navigate to `~/Library/Application Support/Neon/storage` and rename the file `n3UserWallet.json`, restart the app and attempt to import your recovery file 😄 

https://github.com/CityOfZion/neon-wallet/issues/2396

- Added the ability to scan QR codes on the contact screen
Screen Shot 2022-04-29 at 1 43 10 PM

- Additionally I rewrote the "Scan QR" UX on the send screen such that it simply looks for an address and adds it to the current send row (as opposed to the somewhat strange nep9 standard  which i ket in the legacy flow AND which seems to never get used 😅 )

- Also added the ability to login via QR scan on the "Watch only" tab
Screen Shot 2022-04-29 at 1 43 38 PM


kinmya:
Scan QR code visible for each appropriate login type.

PR #2410, opened 2022-05-02
feature: fixes bugs found in `2.12.8`

comountainclimber:

- [ ] Unit tests written?

### DIFF
--- a/app/actions/sendModalActions.js
+++ b/app/actions/sendModalActions.js
@@ -7,9 +7,10 @@ export const ID = 'send'
 
 export const getRecipientData = createActions(
   ID,
-  ({ url, chain }) => (): RecipientData => {
+  ({ url, chain, net }) => async (): Promise<RecipientData> => {
     try {
-      if (chain === 'neo2') return parseQRCode(url)
+      if (chain === 'neo2' || url.includes('neo:'))
+        return parseQRCode({ url, net, chain })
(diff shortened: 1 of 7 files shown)


app/actions/sendModalActions.js
```diff
@@ -7,9 +7,10 @@ export const ID = 'send'
 
 export const getRecipientData = createActions(
   ID,
-  ({ url, chain }) => (): RecipientData => {
+  ({ url, chain, net }) => async (): Promise<RecipientData> => {
     try {
-      if (chain === 'neo2') return parseQRCode(url)
+      if (chain === 'neo2' || url.includes('neo:'))
+        return parseQRCode({ url, net, chain })
```
comountainclimber, line 13: We will attempt to parse the nep9 QR code if it includes protocol otherwise we will just assume it is an address

app/containers/LoginPrivateKey/LoginPrivateKey.jsx
```diff
@@ -58,10 +58,10 @@ export default class LoginPrivateKey extends React.Component<Props, State> {
             <React.Fragment>
               <div className={styles.scannerContainer}>
                 <QrCodeScanner
-                  callback={() =>
+                  callback={key =>
```
comountainclimber, line 61: This is what was breaking the login flow 🙈

PR #2418, opened 2022-08-05
feature: dependency overhaul

comountainclimber:

- [ ] Unit tests written?


config/webpack.config.prod.js
```diff
@@ -2,7 +2,8 @@ const path = require('path')
 const webpack = require('webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
-const ExtractTextPlugin = require('extract-text-webpack-plugin')
```
comountainclimber, line 5: `ExtractTextPlugin` is deprecated and incompatible with node `16.14.2`

package.json
```diff
@@ -189,13 +191,13 @@
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-0": "6.24.1",
     "cross-env": "5.1.1",
-    "css-loader": "0.28.7",
-    "electron": "^10.0.0",
-    "electron-builder": "22.11.11",
+    "css-loader": "4.2.0",
+    "electron": "19.0.8",
```
comountainclimber, line 196: 👏

package.json
```diff
@@ -222,15 +224,14 @@
     "identity-obj-proxy": "3.0.0",
     "jest": "22.4.2",
     "json-loader": "0.5.7",
-    "node-sass": "4.13.1",
```
comountainclimber, line 225: `node-sass` is another deprecated project incompatible with electron 19 and node 16

app/core/storage.js
```diff
@@ -1,9 +1,20 @@
+import { ipcRenderer } from 'electron'
```
comountainclimber, line 1: The storage module needed to be refactored for some reason the default path it was using changed during the electron update. Its working exactly as expected with these changes

melanke (approved):

main.js
```diff
@@ -85,6 +85,7 @@ app.on('ready', () => {
       contextIsolation: true,
       webPreferences: {
         enableRemoteModule: true,
+        contextIsolation: false,
```
comountainclimber, line 88: This was already turned off in our previous builds but the setting has changed to defaulting to true in new versions of electron thus requiring this explicit flag

PR #2421, opened 2022-08-12
feature: replaces instances of dialog with new bridged version

comountainclimber:
[As stated in the docs](https://www.electronjs.org/docs/latest/api/dialog) dialog is now only available in the `main` process... This is a security measure to prevent renderer access to this module. 

This PR replaces all instances in the app with a new version using a bridge to main via `ipcRenderer`


melanke:
Task linked: [CU-2pjkxn3 *Backup and Recover Wallet under Settings does nothing on click](https://app.clickup.com/t/2pjkxn3)

melanke (approved):

PR #2425, opened 2022-08-30
CU-2pjkxp6 - Generate Encrypt a key - copying in a entry under Create…

raulduartep:
… a passphrase did not result in a clickable generate button

- [ ] Unit tests written?

### DIFF
--- a/app/components/Settings/EncryptForm/EncryptForm.jsx
+++ b/app/components/Settings/EncryptForm/EncryptForm.jsx
@@ -39,7 +39,7 @@ export default class EncryptForm extends React.Component<Props, State> {
       privateKeyError: '',
       passphraseError: '',
       confirmPassphraseError: '',
-      isDisabled: false,
+      isDisabled: true,
       privateKey: '',
       passphrase: '',
       confirmPassphrase: '',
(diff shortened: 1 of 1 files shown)


lllwvlvwlll:
Task linked: [CU-2pjkxp6 *Generate “Encrypt a key” - copying in a entry under “Create a passphrase” did not result in a clickable generate button (for example, someone might do this if using a password manager)   If you type in a passphrase, it works fine.  It also worked when I added one character to end of the passphrase.  I was able to remove the added one and it became functional.  (not a good workaround)](https://app.clickup.com/t/2pjkxp6)

app/components/Settings/EncryptForm/EncryptForm.jsx
```diff
@@ -198,10 +210,9 @@ export default class EncryptForm extends React.Component<Props, State> {
     const { privateKey, passphrase, confirmPassphrase } = this.state
 
     const validInput = this.validate(privateKey, passphrase, confirmPassphrase)
-    if (validInput) {
-      this.setState({ isDisabled: true })
-      onSubmit(privateKey, passphrase, confirmPassphrase)
-      this.setState({ isDisabled: false })
-    }
+
```
comountainclimber, line 213: nitpick: I think this reads a little better

```suggestion
    if (validInput) return onSubmit(privateKey, passphrase, confirmPassphrase)
```

comountainclimber (approved):
Looks good to me thanks for fixing this!

PR #2426, opened 2022-09-01
CU-2pjkxye - No visible QR code

raulduartep:

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-2pjkxye *On next step to view the Encrypted QR code, resulted in no visible QR code to copy and no results when selecting copy and trying to paste.](https://app.clickup.com/t/2pjkxye)

app/hocs/withCopyCanvasToClipboard.js
```diff
@@ -45,7 +45,7 @@ export default function withCopyCanvasToClipboard(Component) {
       // $FlowFixMe
       const url = ref.toDataURL('image/jpg', 0.8)
       const base64Data = url.replace(/^data:image\/png;base64,/, '')
-      const storagePath = storage.getDefaultDataPath()
+      const storagePath = storage.getDataPath()
```
comountainclimber: I am pretty sure this is not going to give you the path we use everywhere else in the application... We should be using `const path = await ipcRenderer.invoke('getStoragePath')` here

app/containers/Encrypt/Encrypt.jsx
```diff
@@ -1,61 +1,50 @@
...
+    <FullHeightPanel
+      className={styles.encrypt}
+      headerText={intl.formatMessage({ id: 'encryptPanelHeader' })}
+      renderCloseButton={() => (
+        <CloseButton onClick={resetEncryptedWIF} routeTo={ROUTES.SETTINGS} />
+      )}
+      renderHeaderIcon={HeaderIcon}
+      renderInstructions={
+        encryptedWIF
+          ? () => intl.formatMessage({ id: 'encryptInstructions' })
```
comountainclimber, line 37: We should skip the instructions part OR make the instructions say something different than "Choose a passphrase to encrypt an existing key" like "please save your information in a safe place" once they have successfully encrypted a key (at this point they have already chosen a passphrase)

Screen Shot 2022-09-01 at 10 28 35 AM

comountainclimber:
Two minor things I noticed when reviewing this... Nice work @raulduartep !! 💪

app/hocs/withCopyCanvasToClipboard.js
raulduartep: All right. I was a bit confused on how to get the storage path. Thanks for the alert, I will fix this.

app/containers/Encrypt/Encrypt.jsx
raulduartep: Ok. You are right. I will fix.

comountainclimber:
@raulduartep please let me know when the feedback has been addressed and merge conflicts are resolved 👀

comountainclimber (approved):

PR #2427, opened 2022-09-08
eslint and If Contact list is greater than the size of the visible it…

endkeyCoder:
…ems in the list, the list is cut off. It has a scrollbar, but it doesn't go down to the item on the bottom. Resize and toggle between order of sort (A-Z, Z-A) results in fix of sort. Video attempted. | #2u0ckdj

- [ ] Unit tests written?

### DIFF
--- a/app/components/Contacts/ContactsPanel/ContactsPanel.scss
+++ b/app/components/Contacts/ContactsPanel/ContactsPanel.scss
@@ -3,6 +3,7 @@
 .contactsPanel {
   width: 100%;
   height: 100%;
+  overflow: scroll;
 
   .name {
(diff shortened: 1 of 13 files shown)


lllwvlvwlll:
Task linked: [CU-2u0ckdj If Contact list is greater than the size of the visible items in the list, the list is cut off.  It has a scrollbar, but it doesn't go down to the item on the bottom.  Resize and toggle between order of sort (A-Z, Z-A) results in fix of sort.  Video attempted.](https://app.clickup.com/t/2u0ckdj)

comountainclimber (approved):

PR #2442, opened 2022-10-13
feature: Fixes error handling in approveRequest and fixes logic in reset func

comountainclimber:
Screen Shot 2022-10-13 at 10 15 11 AM


melanke:
Task linked: [CU-34axt6h Fix problem that is causing NWD to return a "Failed or Rejected Request" to Dapps when the user doesn't have sufficient GAS for invocations](https://app.clickup.com/t/34axt6h)

PR #2445, opened 2022-10-21
feature: Fix network config tooltip position

comountainclimber:

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-35rqj6c Clicking on a menu item is often obstructed by the network data popover (check the video on the attachments)](https://app.clickup.com/t/35rqj6c)

PR #2448, opened 2022-11-09
add GFUND logo

vincentgeneste:
adding new logo


comountainclimber:
@vincentgeneste thanks for the PR 🙌 , this wont work as intended unless you also add the svg and update the svg index... 

Let me know if that is unclear or if you need help doing that

vincentgeneste:
are you sure? our GM logo is on neon only with png and it works

comountainclimber:
@vincentgeneste "it works" but the svgs are used for QR code token request generation (although this is currently disabled for n3 we plan on being able to support it in the future) and its also just annoying to not have static files added to the application in a consistent manner.

I must have missed the mistake when the GM logo was added 😓

vincentgeneste:
ok :) added svg then!

comountainclimber:
Looks like there are some merge conflicts 👀

comountainclimber:
@vincentgeneste can you merge `dev` into your fork and resolve these so we can merge thank you 🙏

PR #2450, opened 2022-11-10
feature: Allow users to connect to dapps before logging in

comountainclimber:

- [ ] Unit tests written?

### DIFF
--- a/app/containers/App/App.jsx
+++ b/app/containers/App/App.jsx
@@ -1,5 +1,5 @@
 // @flow
-import React, { useEffect } from 'react'
+import React, { useCallback, useEffect } from 'react'
 import { wallet } from '@cityofzion/neon-js-next'
 
 import { ROUTES, DEFAULT_AUTOACCEPT_METHODS } from '../../core/constants'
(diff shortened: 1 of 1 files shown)


melanke:
Task linked: [CU-376p5gx If I use the WalletConnect deep link when I am not logged in it should show the connection modal after I login](https://app.clickup.com/t/376p5gx)

app/containers/App/App.jsx
```diff
@@ -98,19 +99,40 @@ const App = ({
...
-  useEffect(
-    () => {
-      ipc.on('link', (event, url) => {
-        const { uri } = parseQuery(decodeURI(url))
-        if (uri) {
+  useEffect(() => {
+    ipc.on('link', (event, url) => {
+      const { uri } = parseQuery(decodeURI(url))
+      if (uri) {
+        if (store.getState()?.spunky?.auth?.data?.address) {
```
comountainclimber, line 106: We check to see if there is a loaded address in the state tree, if there is we immediately perform the route change otherwise render the notification and queue the route

melanke (approved):

PR #2452, opened 2022-11-16
Updates package.json

comountainclimber:

- [ ] Unit tests written?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Neon",
-  "version": "2.16.21",
+  "version": "2.16.22",
   "main": "./main.js",
   "description": "Light wallet for NEO blockchain",
   "homepage": "https://github.com/CityOfZion/neon-wallet",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -90,7 +90,8 @@
     "win": {
       "target": "nsis",
       "icon": "build/icon.ico",
-      "artifactName": "${productName}.${version}.${ext}"
+      "artifactName": "${productName}.${version}.${ext}",
+      "verifyUpdateCodeSignature": false
```
comountainclimber, line 94: Pretty sure this was the source of an error being thrown on OTA for windows 👀

comountainclimber:
I have verified that at least moving forward, OTA updates will work on every subsequent release after the one that includes this (only relates to windows)

melanke:
Task linked: [CU-3c6y798 Over the air updates are not working (at least on windows)](https://app.clickup.com/t/3c6y798)

PR #2456, opened 2023-01-03
CU-864dmfjy9 - Upgrade Ghost Market API to v2

raulduartep:

- [ ] Unit tests written?

### DIFF
--- a/app/actions/nftGalleryActions.js
+++ b/app/actions/nftGalleryActions.js
@@ -9,18 +9,10 @@ const NFT_PROVIDER = GHOST_MARKET
 
 export type NftGalleryItem = {
   metadata: {
-    attributes: any,
     description: string,
-    image: string,
-    media_type: string,
+    mediaUri: string,
+    mediaType: string,
     name: string,
-  },
-  series: {
-    chain: string,
-    contrract: string,
-    creator: string,
-    current_supply: number,
-    description: string,
   },
   tokenId: string,
   contract: string,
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-864dmfjy9 Upgrade Ghost Market API to v2](https://app.clickup.com/t/864dmfjy9)

app/actions/nftGalleryActions.js
```diff
@@ -39,39 +31,44 @@ const DEFAULT_NFT_GALLERY_RESULTS = { results: [], page: 0, count: 0 }
...
+        tokenId,
+        contract: contract.hash,
+        collection: {
+          name: collection.name,
+        },
+      }),
+    )
+
+    // eslint-disable-next-line no-console
+    console.log(results)
```
comountainclimber: We should probably remove this log

app/actions/nftGalleryActions.js
```diff
@@ -39,39 +31,44 @@ const DEFAULT_NFT_GALLERY_RESULTS = { results: [], page: 0, count: 0 }
...
-          contract: asset.nft.contract,
-          collection: asset.nft.collection,
-        }
-        return parsed
-      })
+    const { assets, total: count } = response?.data
 
-      return { results: previousResults.concat(results), page, count }
-    }
+    if (!assets || !assets.length || !count) throw new Error('No assets found')
```
comountainclimber: I think keeping the previous pattern of returning an empty payload makes more sense than returning an error here as there is no error that occurred - no NFT's is not really an error

app/actions/nftGalleryActions.js
raulduartep: I forgot to remove this log. Solved

app/actions/nftGalleryActions.js
raulduartep: Ok. Solved

comountainclimber (approved):

PR #2467, opened 2023-02-15
CU-864dz6z38 - Wc: When not sending a signature scope the wallet shou…

raulduartep:
…ld use the default (CalledByEntry, not None. None would be if the developer explicitly says None)

- [ ] Unit tests written?

### DIFF
--- a/app/components/ConnectDapp/ApproveTransaction.js
+++ b/app/components/ConnectDapp/ApproveTransaction.js
@@ -416,7 +416,7 @@ const ApproveTransaction = ({
             ])}
           >
             <label>signature scope</label>
-            {!!request.params.request.params.signers.length && (
+            {request.params.request.params.signers.length ? (
               <div>
                 {
                   WITNESS_SCOPE[
@@ -427,6 +427,8 @@ const ApproveTransaction = ({
                   String(request.params.request.params.signers[0]?.scopes)
                 ] === 'Global' && <WarningIcon />}
               </div>
+            ) : (
+              <div>{WITNESS_SCOPE['1']}</div>
             )}
           </div>
         )}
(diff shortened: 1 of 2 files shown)


lllwvlvwlll:
Task linked: [CU-864dz6z38 Wc: When not sending a signature scope the wallet should use the default (CalledByEntry, not None. None would be if the developer explicitly says None)](https://app.clickup.com/t/864dz6z38)

app/components/ConnectDapp/ApproveTransaction.js
```diff
@@ -427,6 +427,8 @@ const ApproveTransaction = ({
                   String(request.params.request.params.signers[0]?.scopes)
                 ] === 'Global' && <WarningIcon />}
               </div>
+            ) : (
+              <div>{WITNESS_SCOPE['1']}</div>
```
comountainclimber, line 431: Shouldn't this be `WITNESS_SCOPE[1]`? Why is this a string

app/context/WalletConnect/helpers.js
```diff
@@ -330,7 +330,9 @@ class N3Helper {
       account: account.scriptHash,
     })
 
-    signer.scopes = signerEntry && signerEntry.scopes
+    signer.scopes = signerEntry
+      ? signerEntry.scopes
+      : tx.WitnessScope.CalledByEntry
```
comountainclimber, line 335: 👍

app/components/ConnectDapp/ApproveTransaction.js
raulduartep: Because the key for this const is a string. if you look on the line 28, the WITNESS_SCOPE is a const.

PR #2468, opened 2023-02-18
feature: adds custom network support

comountainclimber:
Kapture 2023-02-18 at 13 33 35

FEATURES:
- Upgrade to React 18 🚀 
- Adds Chakra UI (without implementing any theming, rewriting components etc) to the application for quicker UI development
- New settings UI, layout and implementation
- Ability and support for custom/private networks via the settings tab

REFACTORING:
- Sets up a path forward for completely removing Spunky from the application: if you follow the logic of the refactor in this branch as it applies to the slice of state related to `settings` you can see that spunky actions have been swapped out for vanilla React context
- Removes a bunch of unnecessary higher order components
- Adds a `withSettingsContext` HOC as a drop in replacement for some of the others

TODO: 
- [ ] Implement toast notifications when network or other settings update
- [x] Add translations back to settings screens
- [x] Disable activity, NFT tab etc
- [x] Wallet connect testing??
- [x] More thorough manual regression testing, this massive PR touches almost every feature in the application
- [x] Link to settings from network activity tab and from the custom network label
- [x] Implement links for recover and encrypt key
- [x] Network configuration tooltip needs to route => settings
- [x] There needs to still be a way to navigate to to node select panel if a user is on private/main nets
- [x] Switching chains needs to reset network settings
- [x] Fix label for when user is on anything other than the default
- [x] Remove ability to add private net on legacy
- [x] Add the ability to save different networks with names

QUESTIONS:
- ~Do we want to disable some of the navigation tabs when on a private network?~
- ~Should we remove block explorer as a choice in settings? (Dora offers support for both chains)~


config/webpack.config.dev.js
```diff
@@ -142,6 +142,7 @@ module.exports = {
           loader: 'file-loader',
         },
       },
+      { test: /\.mjs$/, include: /node_modules/, type: 'javascript/auto' },
```
comountainclimber, line 145: This change is in order to support ChakraUI

Issue #2424, opened 2022-08-29
Gas claim in activity tab is shown as 0 NEO transfer


ixje:
**Wallet version or commit SHA:**
Version 2.12.11 (2.12.11.8902)

**Environment:**
OSX

**Reproduction steps:**
1. claim gas in the `wallet` tab
2. look at the entry in the `activity` tab

The activity tab seems to be able to identify "vote" and "transfer" events, but not a gas claim event. This is probably because a claim action creates a transaction with 2 notifications
1. a 0 neo transfer from self to self (which is what the activity row shows)
2. a `x` GAS transfer from `null` to self

We know the exact conditions for a "claim" event, so I think we should identify this and show the amount claimed with a "claim" type instead of what we have now.

jseagrave21:
@ixje I would like to work on this but I have not been able to install neon-wallet using WSL. Do you have any tips? If I can get neon-wallet installed, I think I could resolve this issue.

jseagrave21:
Can this be closed?

PR #2470, opened 2023-03-12
Make N3 Gas Claims look like Gas Claims on the Activity Page

jseagrave21:
https://github.com/CityOfZion/neon-wallet/issues/2424

Makes N3 Gas Claims look like Gas Claims on the Activity page

Updated `transactionHistoryActions` to look for a Gas Claim, then it updates the `metadata` so the current code sees it as a Gas Claim. Then I added `N3ClaimAbstract` so it looks right on the Activity Page.

`yarn dev` testing and verification with the unit test.

No.

This change does not fully update the tx `metadata`. I only updated enough to make the tx look like it should on the Activity page.

- [x] Unit tests written?

### DIFF
--- a/app/components/Blockchain/Transaction/N3ClaimAbstract.jsx
+++ b/app/components/Blockchain/Transaction/N3ClaimAbstract.jsx
@@ -1,46 +1,21 @@
 // @flow
 import React from 'react'
 import classNames from 'classnames'
-import { FormattedMessage, IntlShape, injectIntl } from 'react-intl'
-import Button from '../../Button'
+import { injectIntl } from 'react-intl'
 import styles from './Transaction.scss'
 import ClaimIcon from '../../../assets/icons/claim.svg'
-import ContactsAdd from '../../../assets/icons/contacts-add.svg'
-import CopyToClipboard from '../../CopyToClipboard'
 
 type Props = {
   amount: number,
   image: string,
   isPending: boolean,
-  findContact: (address: string) => React$Node | null,
-  from: string,
-  intl: IntlShape,
-  showAddContactModal: (to: string) => void,
   symbol: string,
   txDate: React$Node,
 }
 
 class N3ClaimAbstract extends React.Component<Props> {
   render = () => {
-    const {
-      amount,
-      image,
-      isPending,
-      findContact,
-      from,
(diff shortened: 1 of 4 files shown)


jseagrave21:
@comountainclimber I'm not sure how to get the checks to run remotely. Everything passes locally.

app/components/Blockchain/Transaction/N3ClaimAbstract.jsx
```diff
@@ -0,0 +1,86 @@
...
+      findContact,
+      from,
+      intl,
+      showAddContactModal,
+      symbol,
+      txDate,
+    } = this.props
+    const logo = image && (
+      <img
+        src={image.replace('ipfs://', 'https://ipfs.io/ipfs/')}
```
comountainclimber: This logic is in place in other Transfer components to render NFT images, so it can be removed

app/components/Blockchain/Transaction/N3ClaimAbstract.jsx
```diff
@@ -0,0 +1,86 @@
...
+            </div>
+            <div className={styles.txAmountContainerN3}>{amount}</div>
+          </div>
+          <div className={styles.txSubjectContainerN3}>
+            <p>{contactFrom}</p>
+            <CopyToClipboard
+              className={styles.copy}
+              text={contactFrom}
+              tooltip={intl.formatMessage({ id: 'copyAddressTooltip' })}
+            />
```
comountainclimber: I dont think a user is going to need to add themselves as a contact on gas claims 👀

comountainclimber:
@jseagrave21 this is awesome thanks for contributing!! I has some minor feedback, basically I think we can get away with making the `N3ClaimAbstract.jsx` component much less verbose.

No idea whats going on with CI but hopefully when you push up another commit it will trigger the test suite.

NICE WORK 🤩

jseagrave21:
@comountainclimber thank you! I agree. I followed the example from the `N3VoteAbstract` and trimmed away as much as I think makes sense. Let me know what you think.
Gas Claim

comountainclimber (approved):

ixje:
🎉 nice!

PR #2472, opened 2023-03-15
CU-864dj4759 - Accept traverseIterator Wc requests on NWD

raulduartep:

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-864dj4759 Accept traverseIterator Wc requests on NWD](https://app.clickup.com/t/864dj4759)

PR #2489, opened 2023-05-08
CU-864ej4ey7 - Implement new method on NWD to answer getNetworkVersio…

raulduartep:
…n Wc's request

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-864ej4ey7 Implement new method on NWD to answer `getNetworkVersion` Wc's request](https://app.clickup.com/t/864ej4ey7)

melanke (approved):

PR #2502, opened 2023-06-13
feature: contacts refactor/rewrite

comountainclimber:
This PR rewrites the contacts UX and data structure to support multiple addresses per individual contact. It also adds NNS support to all "address inputs" as well as contacts.

NOTE: I wrote this feature to be compatible with the previous data structure/formatting of contacts in the app

Screenshot 2023-06-13 at 12 51 35 PM
Screenshot 2023-06-13 at 12 52 00 PM
Screenshot 2023-06-13 at 12 52 34 PM

- [ ] Unit tests written?


app/components/Contacts/ContactFormRefactor/ContactFormRefactor.jsx
```diff
@@ -0,0 +1,311 @@
...
+    if (existingNames.includes(event.target.value)) {
+      setErrorMapping({
+        ...errorMapping,
+        name: intl.formatMessage({ id: 'errors.contact.nameDupe' }),
+      })
+    }
+    setName(event.target.value)
+  }
+
+  async function handleChangeAddress(event, index) {
```
melanke: This method is very big, do you think it's interesting to break it in smaller methods? Specially to be easier to have a general understanding on the multiple steps inside it

melanke (changes requested):

app/components/Contacts/ContactFormRefactor/ContactFormRefactor.jsx
comountainclimber: Broke things down into a couple different functions, thanks again was just focused on getting the feature wrapped up

melanke (approved):

melanke (approved):
I read a few "TODO" comments, how we are dealing with them on NWD? Should we create issues?

comountainclimber:
> I read a few "TODO" comments, how we are dealing with them on NWD? Should we create issues?

@melanke There is just one TODO added with this PR, I dont think creating issues is necessary. Its a minor translation issue but there are a lot of missing translations at this point 😅 

https://github.com/CityOfZion/neon-wallet/pull/2502/files#diff-a13bbabe098fcf64c06d39435201a7daf4315e380d8a87519d4a34a54b5040b2R308

melanke:
@comountainclimber 
> There is just one TODO added with this PR

True, the other TODO already existed

> I dont think creating issues is necessary

Ok, but how do you handle "TODO"s on NWD? Simply leaving a comment might be forgotten. I believe this might not be important enough due to the backlog we have, I am just wondering what we could do.

BTW, You can merge the PR, I already approved, this is just a thought

PR #2503, opened 2023-06-14
feature: remove infinite scroll on activity screen

comountainclimber:
Kapture 2023-06-14 at 12 29 55

- [ ] Unit tests written?

### DIFF
--- a/app/actions/transactionHistoryActions.js
+++ b/app/actions/transactionHistoryActions.js
@@ -218,7 +218,6 @@ export default createActions(
     if (chain === 'neo3') {
       const network = net === 'MainNet' ? 'mainnet' : 'testnet'
       const data = await NeoRest.addressTXFull(address, page, network)
-      // console.log({ data })
       count = data.totalCount
       parsedEntries = await computeN3Activity(data, address, net)
     } else {
--- a/app/actions/transactionHistoryActions.js
+++ b/app/actions/transactionHistoryActions.js
@@ -213,10 +213,13 @@ export default createActions(
     }
 
     let parsedEntries = []
+    let count = 0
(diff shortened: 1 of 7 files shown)


app/containers/TransactionHistory/TransactionHistory.scss
```diff
@@ -1,8 +1,8 @@
 .transactionHistory {
   display: flex;
   flex-direction: column;
-  height: 100%;
-  max-height: calc(100vh - 120px) !important;
+  height: calc(100vh - 195px) !important;
+  // max-height: calc(100vh - 120px) !important;
```
melanke: Usually it's a bad idea leave commented code because we never know if this code should be removed or the comment was only to test and commited by mistake. If you have the intention to uncomment, it's better to put a "TODO" comment above it.

app/actions/transactionHistoryActions.js
```diff
@@ -213,10 +213,13 @@ export default createActions(
     }
 
     let parsedEntries = []
+    let count = 0
 
     if (chain === 'neo3') {
       const network = net === 'MainNet' ? 'mainnet' : 'testnet'
       const data = await NeoRest.addressTXFull(address, page, network)
+      // console.log({ data })
```
melanke: same thing here

melanke (changes requested):

app/containers/TransactionHistory/TransactionHistory.scss
comountainclimber: Totally agree... This was meant to be removed 🙈

melanke (approved):

PR #2506, opened 2023-07-03
CU-864efmq36 - Change NWD to use wallet-connect-sdk-wallet-react - Fix deeplink

raulduartep:
This new PR will fix deeplink connection using wallet-connect-sdk-wallet-react lib.

### DIFF
--- a/app/components/ConnectDapp/ConnectionUrlForm/ConnectionUrlForm.jsx
+++ b/app/components/ConnectDapp/ConnectionUrlForm/ConnectionUrlForm.jsx
@@ -1,5 +1,5 @@
 // @flow
-import React, { useState } from 'react'
+import React, { useRef, useState } from 'react'
 
 import { ROUTES } from '../../../core/constants'
 import CloseButton from '../../CloseButton'
@@ -15,16 +15,16 @@ type Props = {
 
 const ConnectionUrlForm = ({ onURI }: Props) => {
   const [url, setUrl] = useState('')
-  const [buttonDisabled, setButtonDisabled] = useState(false)
+  const buttonDisabled = useRef(false)
 
-  async function handleSubmit() {
-    if (buttonDisabled) return
+  const handleSubmit = async (uri: string) => {
+    if (buttonDisabled.current) return
+    buttonDisabled.current = true
 
-    setButtonDisabled(true)
     try {
-      await onURI(url)
+      await onURI(uri)
     } finally {
-      setButtonDisabled(false)
+      buttonDisabled.current = false
     }
   }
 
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-864efmq36 Change NWD to use wallet-connect-sdk-wallet-react](https://app.clickup.com/t/864efmq36)

package.json
```diff
@@ -22,7 +22,7 @@
     "create-release:major": "node scripts/create-release.js major",
     "release": "yarn assets && yarn electron-builder --publish always",
     "release:mac": "yarn assets && yarn electron-builder --publish always --mac --x64 --arm64",
-    "dist": "yarn assets && dotenv yarn electron-builder",
+    "dist": "yarn assets && dotenv yarn electron-builder --win",
```
melanke: I think you should remove the `--win`

melanke (changes requested):

package.json
raulduartep: Sorry, I was testing something and I forgot to remove it.

melanke (approved):

melanke:
@comountainclimber can you review this?

app/components/ConnectDapp/ConnectionUrlForm/ConnectionUrlForm.jsx
```diff
@@ -15,16 +15,16 @@ type Props = {
 
 const ConnectionUrlForm = ({ onURI }: Props) => {
   const [url, setUrl] = useState('')
-  const [buttonDisabled, setButtonDisabled] = useState(false)
+  const buttonDisabled = useRef(false)
```
comountainclimber, line 18: What is the point of this change? seems more simple to use the `useState` hook

raulduartep: The point is performance, this variable is just to control not executing the function more than once at a time. The useState hook causes a rerender in the component and there is no reason to do that, since it has no visual effect for the disabled button, while the useRef hook does not rerender and serves only to control the execution of the function.

comountainclimber: "there is no reason to do that, since it has no visual effect for the disabled button" it does have a reason, the button in the UI should be disabled if the form is invalid 😅 (for instance if there is no input)

raulduartep: I agree, it should have a state of disabled in the ui, but before my changes the button doesn't disable, so I didn't change that, since my issue was just to change the functionality and not the design.

comountainclimber (approved):

PR #2511, opened 2023-07-25
CU-864eu7dad - The network data popover is unreachable (at least on w…

raulduartep:
…indows)

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-864eu7dad The network data popover is unreachable (at least on windows)](https://app.clickup.com/t/864eu7dad)

comountainclimber:
It looks like you removed the drop shadow (at least from the light theme) on the "popover":

production:
Screenshot 2023-07-25 at 9 21 36 AM

your branch:
Screenshot 2023-07-25 at 9 19 58 AM


Probably worth double checking that there is not also shadow or styles missing from the dark theme

comountainclimber (approved):

PR #2514, opened 2023-07-26
feature: fixes notification bug in ApproveConnection.jsx

comountainclimber:
Kapture 2023-07-26 at 15 06 30

- [ ] Unit tests written?

### DIFF
--- a/app/components/ConnectDapp/ApproveConnection/ApproveConnection.jsx
+++ b/app/components/ConnectDapp/ApproveConnection/ApproveConnection.jsx
@@ -66,19 +66,13 @@ const ApproveConnection = ({
     }
   }
 
-  const rejectSession = async (error?: string) => {
+  const rejectSession = async () => {
     try {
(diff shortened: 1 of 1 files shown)


app/components/ConnectDapp/ApproveConnection/ApproveConnection.jsx
```diff
@@ -66,19 +66,13 @@ const ApproveConnection = ({
     }
   }
 
-  const rejectSession = async (error?: string) => {
```
comountainclimber, line 69: This function was never intentionally being invoked in this component with the error param 👀

melanke: I checked the `dev` branch and the reported error is not happening anymore. I think this PR is safe to be merged if you indeed checked that no one is using the `error` param.

melanke (approved):

PR #2516, opened 2023-08-23
CU-86a0b89zf - Wrong text on sign message confirmation window

luc10921:

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-86a0b89zf Wrong text on sign message confirmation window](https://app.clickup.com/t/86a0b89zf)

melanke (approved):

PR #2545, opened 2023-10-19
CU-86a129e02 - Initial Steps Flow Implementation

raulduartep:


lllwvlvwlll:
Task linked: [CU-86a129e02 Initial Steps Flow Implementation](https://app.clickup.com/t/86a129e02)

lllwvlvwlll:
Task linked: [CU-86a129e02 Initial Steps Flow Implementation](https://app.clickup.com/t/86a129e02)

src/renderer/src/hooks/useBlockchainActions.ts
```diff
@@ -0,0 +1,156 @@
...
+  index?: number
+}
+
+export type TWalletToCreate = {
+  name: string
+  walletType: TWalletType
+  mnemonic?: string
+}
+
+const colors = {
```
melanke:
```suggestion
const accountColors = {
```

src/renderer/src/pages/Dashboard/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const DashboardPage = () => {
+  const isFirstTimeRef = useSelectorRef(state => state.settings.isFirstTime)
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    if (isFirstTimeRef.current) {
+      navigate('/welcome')
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
melanke: Why is this happening? Maybe we can change the lint rules

src/renderer/src/store/account/Account.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.accountType = data.accountType
+    this.idWallet = data.idWallet
+    this.name = data.name
+    this.blockchain = data.blockchain
+    this.backgroundColor = data.backgroundColor
+    this.encryptedKey = data.encryptedKey
+  }
+
+  deserialize() {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
melanke: https://stackoverflow.com/a/69006568

src/renderer/src/hooks/useBlockchainActions.ts
```diff
@@ -0,0 +1,156 @@
...
+        walletType,
+        id,
+        encryptedMnemonic,
+      })
+
+      dispatch(walletReducerActions.saveWallet(newWallet.deserialize()))
+
+      return newWallet
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
melanke: why this is happening? Let's discuss how to avoid this comments, maybe we can change the lint rules

tailwind.config.js
```diff
@@ -34,10 +35,11 @@ module.exports = {
       white: {
         DEFAULT: '#FFFFFF',
       },
-      grey: {
+      gray: {
         black: '#13191B',
         darker: '#293139',
         dark: '#333D46',
+        dark_mid: '#42525C',
```
melanke: replace all this names by numbers to have a better meaning. Otherwise it will lead to questions like "Which is darker: `dark` or `dark_mid`"? Tailwind itself knows about this problem and uses number by default: https://tailwindcss.com/docs/customizing-colors#color-object-syntax

melanke (changes requested):

src/renderer/src/store/account/Account.ts
melanke: You forgot to remove the comment and change to `_` here?

src/renderer/src/pages/Dashboard/index.tsx
melanke: Could we add the ref on the dependencies here too?

PR #2549, opened 2023-10-26
CU-86a16b4wx - NEON3 - Dashboard Basic Structure + Side Menu + Dumb H…

raulduartep:
…eader


lllwvlvwlll:
Task linked: [CU-86a16b4wx NEON3 - Dashboard Basic Structure + Side Menu + Dumb Header](https://app.clickup.com/t/86a16b4wx)

src/renderer/src/contexts/ModalRouter.tsx
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
melanke: why?

src/renderer/src/contexts/ModalRouter.tsx
```diff
@@ -0,0 +1,137 @@
...
+        return setNonRepeatedHistoryElement(routeElement)
+      }
+
+      setHistory(prevState => prevState.slice(0, name))
+      return
+    },
+    [setNonRepeatedHistoryElement, routes]
+  )
+
+  const getElementByHistory = (historyElement: THistory) => {
```
melanke: I think this can be a callback, right?

melanke (changes requested):

src/renderer/src/contexts/ModalRouter.tsx
raulduartep: It was a miss code, probably when I was implementing it, I was too lazy to type something and wrote it to test. Sorry

src/renderer/src/contexts/ModalRouter.tsx
raulduartep: Since we are only using it in the rendering phase, there is no need to create a callback for it as no react hook depends on it. Furthermore, React doesn't let you call a callback inside a loop https://react.dev/reference/react/useCallback#i-need-to-call-usememo-for-each-list-item-in-a-loop-but-its-not-allowed.

PR #2559, opened 2023-11-06
CU-8678g209y - nwd: encrypt, decrypt and decryptFromArray WC's methods

endkeyCoder:

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-8678g209y nwd: encrypt, decrypt and decryptFromArray WC's methods ](https://app.clickup.com/t/8678g209y)

PR #2561, opened 2023-11-06
CU-8678g209y - nwd: encrypt, decrypt and decryptFromArray WC's methods

melanke:


lllwvlvwlll:
Task linked: [CU-8678g209y nwd: encrypt, decrypt and decryptFromArray WC's methods ](https://app.clickup.com/t/8678g209y)

PR #2562, opened 2023-11-06
CU-86a1baw6y - Neon2 - Fix issue that is hiding the backup feature fr…

endkeyCoder:
…om the user interface

- [ ] Unit tests written?

### DIFF
--- a/app/containers/Settings/NewSettings.jsx
+++ b/app/containers/Settings/NewSettings.jsx
@@ -7,6 +7,7 @@ import fs from 'fs'
 
 import { get } from 'lodash-es'
 import { useWalletConnectWallet } from '@cityofzion/wallet-connect-sdk-wallet-react'
+import moment from 'moment'
 import { getStorage } from '../../core/storage'
 import { recoverWallet } from '../../modules/generateWallet'
 import { useSettingsContext } from '../../context/settings/SettingsContext'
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-86a1baw6y Neon2 - Fix issue that is hiding the backup feature from the user interface](https://app.clickup.com/t/86a1baw6y)

app/containers/Settings/NewSettings.jsx
```diff
@@ -35,9 +36,48 @@ import Plus from '../../assets/icons/add.svg'
...
+  showErrorNotification,
+  showSuccessNotification,
+  chain,
+}: {
+  showErrorNotification: Object => any,
+  showSuccessNotification: Object => any,
+  chain: string,
+}) => {
+  const path = await ipcRenderer.invoke('getPath', 'documents')
+  console.warn(path)
```
raulduartep: why do we need this log ?

app/core/backup.js
```diff
@@ -0,0 +1,24 @@
...
+    fs.writeFile(path, JSON.stringify(accounts), errorWriting => {
+      if (errorWriting && callbackError) {
+        callbackError(errorWriting.message)
+      } else if (callbackSuccess) {
+        callbackSuccess()
+      }
+    })
+  } catch (error) {
+    throw error
+  }
```
raulduartep: It's redundant, you're catching an error to throw it, it's the same thing not catching it. So I recommend removing the try catch block

raulduartep (approved):

Issue #2560, opened 2023-11-06
Release notes in the wallet settings page is not up to date


edgedlt:
**Wallet version or commit SHA:**
v2.21.3

**Screenshots/Video:**
![image](https://github.com/CityOfZion/neon-wallet/assets/21316801/cde90742-9569-4da5-81da-24387b91d484)

PR #2569, opened 2023-11-08
Update Release Notes

jseagrave21:
Addresses https://github.com/CityOfZion/neon-wallet/issues/2560

Brings the release notes up to date with the most recent version.
Additionally, I fixed the "PatchDark" image so it appears in Dark theme

I included releases I thought were pertinent.

live testing

No

No

- [ ] Unit tests written?


jseagrave21:
Closing due to lack of interest.  Will re-open if we want to update the release notes in the future.

jseagrave21:
@lllwvlvwlll @raulduartep do you have any guidance for updating the Release Notes?  I was trying to address #2560 but I see that it has closed after becoming stale.

If this is useful then I will open another PR and follow any guidance you have.

PR #2570, opened 2023-11-10
86a1cjzyw - Neon2 - Improve encrypt/decrypt error handling so to show…

ricardoak25:
… only for users with at least one contact

- [ ] Unit tests written?


lllwvlvwlll:
Task linked: [CU-86a1cjzyw Neon2 - Improve encrypt/decrypt error handling so to show only for users with at least one contact](https://app.clickup.com/t/86a1cjzyw)

thiagocbalducci (approved):

PR #2576, opened 2023-11-14
CU-86a16bduq - NEON3 - Contact Management Implementation (-NNS)

yumiuehara:

- [ ] Unit tests written?

### DIFF
--- a/src/renderer/src/helpers/UtilsHelper.ts
+++ b/src/renderer/src/helpers/UtilsHelper.ts
@@ -1,3 +1,4 @@
+import * as uuid from 'uuid'
 export abstract class UtilsHelper {
   static getRandomNumber(max: number) {
     return Math.floor(Math.random() * Math.floor(max))
@@ -56,5 +57,13 @@ export abstract class UtilsHelper {
 
       return aValue < bValue ? 1 : -1
     })
+  }
+
+  static copyToClipboard(text: string) {
+    navigator.clipboard.writeText(text)
+  }
+
+  static uuid() {
+    return uuid.v4()
   }
 }
--- a/src/renderer/src/layouts/Modal.tsx
+++ b/src/renderer/src/layouts/Modal.tsx
@@ -1,5 +1,5 @@
 import { cloneElement, ComponentProps, ReactNode } from 'react'
-import { MdClose } from 'react-icons/md'
+import { MdClose, MdKeyboardBackspace } from 'react-icons/md'
 import { IconButton } from '@renderer/components/IconButton'
 import { Separator } from '@renderer/components/Separator'
 import { StyleHelper } from '@renderer/helpers/StyleHelper'
@@ -11,15 +11,31 @@ type TProps = {
   headingIcon?: JSX.Element
   headingIconFilled?: boolean
   contentClassName?: string
+  withBackButton?: boolean
 } & ComponentProps<'div'>
 
-export const ModalLayout = ({ heading, children, headingIcon, headingIconFilled = true, contentClassName }: TProps) => {
+export const ModalLayout = ({
+  heading,
+  children,
+  headingIcon,
+  headingIconFilled = true,
+  contentClassName,
+  withBackButton = false,
+}: TProps) => {
   const { modalNavigateWrapper } = useModalNavigate()
 
   return (
     <div className="w-[18.75rem] bg-gray-800 h-full px-4 text-white text-xs flex flex-col">
       <header className="flex justify-between items-center py-2.5">
-        <div className="flex gap-x-2.5">
+        {withBackButton && (
+          <IconButton
+            icon={<MdKeyboardBackspace className="fill-white" />}
+            size="md"
+            compacted
+            onClick={modalNavigateWrapper(-1)}
+          />
+        )}
+        <div className="flex items-center gap-x-2.5">
           {headingIcon &&
             cloneElement(headingIcon, {
               className: StyleHelper.mergeStyles('w-6 h-6', {
--- a/src/renderer/src/routes/modals/AddAddress/Step1.tsx
+++ b/src/renderer/src/routes/modals/AddAddress/Step1.tsx
@@ -1,0 +1,102 @@
+import { FormEvent, useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { TbPlus } from 'react-icons/tb'
+import { TBlockchainServiceKey } from '@renderer/@types/blockchain'
+import { TContactAddress } from '@renderer/@types/store'
+import { BlockchainIcon } from '@renderer/components/BlockchainIcon'
+import { Button } from '@renderer/components/Button'
+import { Input } from '@renderer/components/Input'
+import { Separator } from '@renderer/components/Separator'
+import { useBsAggregatorSelector } from '@renderer/hooks/useBlockchainSelector'
+import { useModalLocation, useModalNavigate } from '@renderer/hooks/useModalRouter'
+import { ModalLayout } from '@renderer/layouts/Modal'
+
+type TLocationState = {
+  contactName: string
+  handleAddAddress: (address: TContactAddress) => void
+}
+
+export const AddAddressModalStep1 = () => {
+  const { t } = useTranslation('modals', { keyPrefix: 'addAddressStep1' })
+  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
+  const { t: blockchainT } = useTranslation('common', { keyPrefix: 'blockchain' })
+  const location = useModalLocation<TLocationState>()
+  const { bsAggregatorRef } = useBsAggregatorSelector()
+  const { modalNavigate } = useModalNavigate()
+
+  const [selectedBlockchain, setSelectedBlockchain] = useState<TBlockchainServiceKey>()
+  const [search, setSearch] = useState<string | null>(null)
+
+  const contactName = location.state.contactName
+
+  const filteredBlockchain = useMemo(() => {
+    const blockchainServices = Object.keys(bsAggregatorRef.current.blockchainServicesByName) as TBlockchainServiceKey[]
+
+    if (search) {
+      return blockchainServices.filter(blockchain =>
+        blockchain.toLocaleLowerCase().includes(search.toLocaleLowerCase() as string)
+      )
+    }
+
+    return blockchainServices
+  }, [search, bsAggregatorRef])
+
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault()
+
+    modalNavigate('addAddressStep2', {
+      state: {
+        contactName: contactName,
+        contactBlockchain: selectedBlockchain,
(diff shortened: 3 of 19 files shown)


lllwvlvwlll:
Task linked: [CU-86a16bduq NEON3 - Contact Management Implementation (-NNS)](https://app.clickup.com/t/86a16bduq)

src/renderer/src/helpers/UtilsHelper.ts
```diff
@@ -57,4 +57,24 @@ export abstract class UtilsHelper {
...
+  }
+
+  static uuid() {
+    let dt = new Date().getTime()
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, c => {
+      const r = (dt + Math.random() * 16) % 16 | 0
+      dt = Math.floor(dt / 16)
+      return (c === 'x' ? r : (r & 0x3) | 0x8).toString(16)
+    })
+  }
```
raulduartep: You don't need this. You can use the "uuid" lib.
```suggestion

import * as uuid from 'uuid'
export class UtilsHelper {
   ...methods

  static uuid() {
    return uuid.v4()
  }
}
```

src/renderer/src/layouts/Modal.tsx
```diff
@@ -11,15 +11,31 @@ type TProps = {
   headingIcon?: JSX.Element
   headingIconFilled?: boolean
   contentClassName?: string
+  goBackButton?: boolean
```
raulduartep: I think "withBackButton" is a better name. What do you think ?

src/renderer/src/helpers/UtilsHelper.ts
```diff
@@ -57,4 +57,24 @@ export abstract class UtilsHelper {
...
+
+  static getNameInitials(name: string) {
+    return name
+      .split(' ')
+      .map(word => word[0])
+      .join()
+  }
```
raulduartep: It seems to be something that is private to your screen and will not be used by anyone else. I think it's better to be a method of the screen you use.

src/renderer/src/model/Contact.ts
raulduartep, line 1: You don't need this class, use the interface directly. See how we are using the IWalletState or IAccountState

src/renderer/src/routes/modals/AddAddress/Step1.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
+  const { t: blockchainT } = useTranslation('common', { keyPrefix: 'blockchain' })
+  const location = useModalLocation<TLocationState>()
+  const contactName = location.state.contactName
+  const { bsAggregatorRef } = useBsAggregatorSelector()
+  const { modalNavigate } = useModalNavigate()
+  const blockchainServices = Object.keys(bsAggregatorRef.current.blockchainServicesByName) as TBlockchainServiceKey[]
+  const [selectedBlockchain, setSelectedBlockchain] = useState(location.state?.address?.blockchain)
+  const [search, setSearch] = useState<string | null>(null)
+  const addressCopy = location.state.address
```
raulduartep: Try to separate the useState, useMemo, hooks and simple const with space, it turns the code easier to read.

```suggestion
  const { t } = useTranslation('modals', { keyPrefix: 'addAddressStep1' })
  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
  const { t: blockchainT } = useTranslation('common', { keyPrefix: 'blockchain' })
  const location = useModalLocation<TLocationState>()
  const { bsAggregatorRef } = useBsAggregatorSelector()
  const { modalNavigate } = useModalNavigate()
  
  const [selectedBlockchain, setSelectedBlockchain] = useState(location.state?.address?.blockchain)
  const [search, setSearch] = useState<string | null>(null)
  
    const blockchainServices = Object.keys(bsAggregatorRef.current.blockchainServicesByName) as TBlockchainServiceKey[]
  const addressCopy = location.state.address
    const contactName = location.state.contactName
```

Sometimes you will need a different separation, because some state are using some const, so you will need to declare the const before the state. But always try to separate

src/renderer/src/routes/modals/AddAddress/Step1.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const filteredBlockchain = useMemo(() => {
+    if (search) {
+      return blockchainServices.filter(blockchain =>
+        blockchain.toLocaleLowerCase().includes(search.toLocaleLowerCase() as string)
+      )
+    }
+
+    return blockchainServices
+  }, [search, blockchainServices])
```
raulduartep: blockchainServices is a simple const, so it will always update that will always call this useMemo. You need to declare the blockchainService inside the useMemo and remove it from dependencies array.

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { contactReducerActions } from '@renderer/store/reducers/ContactReducer'
+import { cloneDeep } from 'lodash'
+
+type TFormData = {
+  name: string
+  addresses: TContactAddress[]
+}
+
+type TLocationState = {
+  contact?: IContactState | null
```
raulduartep: You don't need the null

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export const PersistContactModal = () => {
+  const { t } = useTranslation('modals', { keyPrefix: 'persistContactModal' })
+  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
+
+  const { modalNavigate, modalNavigateWrapper } = useModalNavigate()
+  const location = useModalLocation<TLocationState>()
+  const dispatch = useAppDispatch()
+
+  const contact = useMemo(() => location.state?.contact || null, [location.state?.contact])
```
raulduartep: You can use a simple const here

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                    onClick={() => openAddAddressModal(address)}
+                    label={address.address}
+                    leftIcon={<BlockchainIcon blockchain={address.blockchain} type="white" />}
+                    flat
+                    variant="outlined"
+                    className="truncate"
+                    type="button"
+                  />
+                  <IconButton icon={<TbTrash />} compacted type="button" onClick={() => handleDeleteAddress(address)} />
+                </div>
```
raulduartep, line 141: Transform the div into a button instead of create a button inside the div

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                  <Button
+                    clickableProps={{ className: 'border-none' }}
+                    onClick={() => openAddAddressModal(address)}
+                    label={address.address}
+                    leftIcon={<BlockchainIcon blockchain={address.blockchain} type="white" />}
+                    flat
+                    variant="outlined"
+                    className="truncate"
+                    type="button"
+                  />
```
raulduartep: It does not need to be a button. I think its not possible to edit an added address

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const filterAddress = addresses.filter(address => address !== addressToDelete)
+
+    setAddresses(filterAddress)
+  }
+
+  const openAddAddressModal = (selectedAddress?: TContactAddress) => {
+    if (!contactName) {
+      form.setError('name', { message: t('invalidName') })
+      return
+    }
```
raulduartep, line 71: You don't need to check this, let the user add a new address even if he hasn't added a name, name validation is already in handleSubmit

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const [addresses, setAddresses] = useState<TContactAddress[]>(cloneDeep(contact?.addresses) || [])
+  const [showAddressError, setShowAddressError] = useState(false)
+
+  const form = useForm<TFormData>({
+    defaultValues: {
+      name: contact?.name || '',
+    },
+  })
+
+  const contactName = form.watch('name')
```
raulduartep: Remove it and use form.getValues("name") inside the function that is using it

src/renderer/src/routes/modals/PersistContact/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const { t } = useTranslation('modals', { keyPrefix: 'persistContactModal' })
+  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
+
+  const { modalNavigate, modalNavigateWrapper } = useModalNavigate()
+  const location = useModalLocation<TLocationState>()
+  const dispatch = useAppDispatch()
+
+  const contact = useMemo(() => location.state?.contact || null, [location.state?.contact])
+  const [addresses, setAddresses] = useState<TContactAddress[]>(cloneDeep(contact?.addresses) || [])
+  const [showAddressError, setShowAddressError] = useState(false)
```
raulduartep: Use react-hook-form to control it.

src/renderer/src/routes/modals/AddAddress/Step1.tsx
```diff
@@ -0,0 +1,117 @@
...
+                  <BlockchainIcon blockchain={service} type="white" />
+                  {blockchainT(service)}
+                </div>
+                <Input
+                  type="checkbox"
+                  className="accent-neon w-4 h-4 outline-none after:ring-neon ring-neon ring-inset"
+                  containerClassName="w-fit"
+                  {...form.register('blockchain')}
+                  onChange={() => handleCheckboxChange(service)}
+                  checked={isChecked(service)}
```
raulduartep, line 92: You are already using react-hook-form, so you shouldn't need this

src/renderer/src/routes/modals/AddAddress/Step2.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const AddAddressModalStep2 = () => {
+  const { t } = useTranslation('modals', { keyPrefix: 'addAddressStep2' })
+  const { t: commonT } = useTranslation('common', { keyPrefix: 'general' })
+  const location = useModalLocation<TLocationState>()
+  const contactName = location.state.contactName
+  const blockchain = location.state.contactAddress.blockchain
+  const { modalNavigate } = useModalNavigate()
+  const { bsAggregatorRef } = useBsAggregatorSelector()
+  const [isAddressValid, setIsAddressValid] = useState<boolean>(false)
+  const [address, setAddress] = useState(location.state.contactAddress.address)
```
raulduartep: You are using react-hook-form, so you shouldn't need this

PR #2589, opened 2023-12-08
CU-86a1jhbyg - NEON3 - Implement Dapp Connection Process

raulduartep:


lllwvlvwlll:
Task linked: [CU-86a1jhbyg NEON3 - Implement Dapp Connection Process](https://app.clickup.com/t/86a1jhbyg)

src/renderer/src/libs/walletConnectSDK.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (!account.encryptedKey) throw new Error('Key not found')
+
+    const key = await window.api.decryptBasedEncryptedSecret(account.encryptedKey, encryptedPassword)
+    if (!key) throw new Error('Error to decrypt key')
+
+    return key
+  }
+
+  async getWalletInfo(): Promise<any> {
+    // TODO: Implement this method when ledger is supported
```
melanke: every TODO comment needs an issue code, otherwise it will be forgotten

melanke: If you think this is not a real problem, ask Mirella and Lucas, but warning: they might scream at you LOL JK

melanke (changes requested):

PR #2601, opened 2024-01-05
CU-86a16bduq - NEON3 - Contact Management Implementation (-NNS)

luc10921:

- [ ] Unit tests written?

### DIFF
--- a/src/renderer/src/routes/pages/Contacts/index.tsx
+++ b/src/renderer/src/routes/pages/Contacts/index.tsx
@@ -1,8 +1,9 @@
 import { useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { BiSolidDownArrow, BiSolidSortAlt, BiSolidUpArrow } from 'react-icons/bi'
 import { FiSend } from 'react-icons/fi'
+import { MdOutlineContentCopy } from 'react-icons/md'
 import { TbPencil, TbPlus } from 'react-icons/tb'
-import { TbCopy } from 'react-icons/tb'
 import { IContactState } from '@renderer/@types/store'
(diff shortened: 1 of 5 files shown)


lllwvlvwlll:
Task linked: [CU-86a16bduq NEON3 - Contact Management Implementation (-NNS)](https://app.clickup.com/t/86a16bduq)

src/renderer/src/routes/pages/Contacts/index.tsx
```diff
@@ -70,18 +73,39 @@ export const ContactsPage = () => {
...
-  const getNameInitials = (name: string) => {
+  const sortByBlockchain = useMemo(() => {
+    if (sortBlockchain === null) {
+      return (_a: TContactAddress, _b: TContactAddress) => 0
+    } else if (sortBlockchain) {
+      return (a: TContactAddress, b: TContactAddress) => a.address.localeCompare(b.address)
+    } else {
+      return (a: TContactAddress, b: TContactAddress) => b.address.localeCompare(a.address)
+    }
+  }, [sortBlockchain])
```
raulduartep: Instead of returning a sorting function, it would be better to use this memo to actually do the sorting, because you don't need to sort for every re-render, as it stands now, React will run the sort function every time the component re-renders.

```suggestion
  const sortedSelectedContactAddresses = useMemo(() => {
    if (!selectedContact) return
  
    if (sortBlockchain === null) {
      return selectedContact.addresses
    }
    
    if (sortBlockchain) {
      return selectedContact.addresses.sort(a, b) => a.address.localeCompare(b.address)
    }
    
      return  selectedContact.addresses.sort(a, b) => b.address.localeCompare(a.address)
 
  }, [selectedContact, sortBlockchain])
```

PR #2605, opened 2024-01-17
86a1xyc69 - NEON3 - SEND Tokens - Implement Input Screen + Wallet Balance

ricardoak25:

PR #2606, opened 2024-01-18
86a1xyq3y - NEON3 - SEND Tokens - Implement Target Address input or Contact

ricardoak25:


lllwvlvwlll:
Task linked: [CU-86a1xyq3y NEON3 - SEND Tokens - Implement Target Address input or Contact](https://app.clickup.com/t/86a1xyq3y)

ricardoak25:
It's necessary to first perform the CR and merge for:
https://github.com/CityOfZion/neon-wallet/pull/2605

src/renderer/src/components/Contact/ContactList.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+type TProps = {
+  onContactSelected: (contact: IContactState | null) => void
+  selectFirst: boolean
+  showSelectedIcon: boolean
+  children?: React.ReactNode
+}
+
+export const ContactList = ({ onContactSelected, selectFirst, showSelectedIcon, children }: TProps) => {
+  const { t: contactT } = useTranslation('pages', { keyPrefix: 'contacts' })
```
raulduartep: It would be better if this translation was in components instead of pages, since it was changed to be a component

src/renderer/src/routes/pages/Send/Recipient.tsx
```diff
@@ -13,16 +26,30 @@ export const Recipient = () => {
...
+          onClick={modalNavigateWrapper('select-contact', {
+            state: {
+              handleSelectContact: handleSelectContact,
+            },
+          })}
+        >
           <TbUsers className="w-5 h-5 text-neon mr-1" />
           <span className="text-neon">{t('contacts')}</span>
-        </div>
+        </button>
```
raulduartep: What do you think about creating a new <Button> variant called text for it ? It would be better because we could reuse it elsewhere. Follow the same structure as we have for the other button variants

Issue #2602, opened 2024-01-06
Incorrect Transaction Totals Displayed on Activity Display for N3 Wallet


jseagrave21:
**Wallet version or commit SHA:**
2.23.10

**Environment:**
Linux Mint 21.2 Cinnamon

**Prerequisites:**
Must have a N3 wallet with multiple transactions.

~I am not sure how many.  In my example, it seems that the total displayed is capped at 188.~

Update:  You need to have non-NEP-17 transfers.

**Reproduction steps:**
1. Open the N3 Wallet
2. Navigate to the Activity tab
3. Load more transactions

In my example N3 wallet, only 188 of the total 230 transactions are displayed as the total number of available transactions.  Then when the total transactions are loaded, the number of available transactions goes to 0 and the option to load more transactions remains displayed.

This seems to be a N3 problem only.  The N2 wallet shows all transactions and when all are loaded, the displayed total matches the total number of available transactions and the option to load more is no longer available.

**Screenshots/Video:**

- N3 wallet with more transactions displayed than available
![not all txs displayed](https://github.com/CityOfZion/neon-wallet/assets/40873301/f0b36b3d-07a5-4788-ba56-d825bd67f959)

- N3 wallet with all transactions displayed
![all txs loaded](https://github.com/CityOfZion/neon-wallet/assets/40873301/44c64a0e-e9e5-4bed-b039-dcb28c297e88)

- N2 wallet with some transactions loaded
![Partial Load of Legacy](https://github.com/CityOfZion/neon-wallet/assets/40873301/2beb0630-e564-4668-a2b0-0a059e2485a8)

- N2 wallet with all transactions displayed
![Legacy](https://github.com/CityOfZion/neon-wallet/assets/40873301/63e4d08c-09c2-4c70-81d2-621d26826fa0)

jseagrave21:
While trying to research a solution to this problem, I discovered that neon-wallet is using a deprecated method from the rest API.  Specifically, we are using `addressTXFull`, which is annotated as being deprecated [here](https://github.com/CityOfZion/dora-sdk-ts/blob/develop/src/api/neoN3/rest.ts#L56).  However, if you trying to use `addressTransactions` as directed, the code breaks because none of the `metadata` is available in the new method.  Additionally, the `totalCount` displayed for the same address referenced above using `addressTransactions` is 181 which is still wrong.  @lllwvlvwlll @ixje

ixje:
[...]

PR #2612, opened 2024-01-28
Remove Number of Transactions & Total Transactions from Activity Page

jseagrave21:
https://github.com/CityOfZion/neon-wallet/issues/2602

This fixes the problem of displaying an incorrect total number of transactions for a wallet address.  This problem stems from inconsistencies with `totalCount`.  Refer to the issue for more details.

I removed the number of transactions (`transactions.length`) and the total number of transactions (`count`) from the Activity page.  Then I added logic to ensure the Load More button is disabled when there are no more transactions to load.  Specifically, now we check to see if there is a next page after we load each page from Dora.  If there isn't, the `count` is set to the total number of transactions loaded (`entries.length`).  The new logic does not affect the Neo Legacy logic, so it still disables the Load More button accurately using `count` as before.

Live testing with a Neo Legacy address,and two Neo3 addresses, one of which only had a single page of transactions.

Nothing additional.

No.


jseagrave21:

Issue #2618, opened 2024-02-03
Images Not Displaying in Dark Theme


jseagrave21:
**Wallet version or commit SHA:**
2.23.10

**Environment:**
Linux Mint 21.3 Cinnamon

**Prerequisites:**
None.

**Reproduction steps:**
1. Start Neon Wallet
2. Select Dark Theme
3. Select Settings
4. Navigate to Release Notes
5. Scroll down

Observe that some images do not load and display artifacts instead.

**Screenshots/Video:**
![Light](https://github.com/CityOfZion/neon-wallet/assets/40873301/82c5d3a8-9313-4601-88be-c7562c12c0fd)
![Dark](https://github.com/CityOfZion/neon-wallet/assets/40873301/be393dbb-bfa6-4667-af2b-30da6398df3b)

jseagrave21:
Up

PR #2619, opened 2024-02-03
Fix Release Notes Images

jseagrave21:
#2618 

This fixes/replaces the images that were not displaying correctly in Release Notes.

Updated one of the `.svg` file and replaced two `.svg` files with `.png` files because the `.svg` dark image would not display correctly.

Live Testing

No

No


jseagrave21:

Issue #2575, opened 2023-11-13
NEP-5 balances not displaying for some accounts


hal0x2328:
Legacy NEP-5 balance display is broken under some conditions - one condition is if the user has ever provided liquidity to a legacy Flamingo pool.

The issue is related to https://github.com/CityOfZion/neon-js/issues/921 - when the token info call fails due to neon-js' inability to decode the totalSupply (even on a LP pair token with no remaining balance), Neon no longer continues trying to retrieve token balances for the rest of the NEP-5 tokens in the wallet.

If possible, Neon should try and handle the token info error gracefully and just skip any it can't decode (I don't believe totalSupply is even used anywhere in Neon)

Here's an example for reproduction:
1. Import a watch account for the address AKEX765mtnPUPj4L5n5C8KcauGkRkuTuxQ
2. Neon shows the following error in the debug tools console:
```
bundle.js:formatted:525160 Error: Expected a hexstring but got 33119936322
    at S (bundle.js:formatted:39878:27)
    at M (bundle.js:formatted:39863:17)
    at Array.O (bundle.js:formatted:69284:57)
    at bundle.js:formatted:42405:36
    at Array.map (<anonymous>)
    at f.parse (bundle.js:formatted:42404:30)
    at bundle.js:formatted:41279:27
```
3. No NEP-5 token balances are displayed, despite this address having a balance of nNEO and fUSDT

jseagrave21:
@hal0x2328 I started working on this problem and was able to resolve the issue with Legacy `nep-5` transactions not showing up on the Activity tab. However, balances are still not correctly displayed and I think that is because the master [token list](https://github.com/CityOfZion/neo-tokens) has not been up-to-date.  I checked the address you listed and NeoTube shows a balance of fUSDT at the time of writing this, which is not included in the master list of tokens.  I am not sure who is maintaining that list anymore but it looks like @comountainclimber was the last person to merge an update to it.

PR #2630, opened 2024-02-17
Fix NEP-5 Tokens Not Displaying on Legacy Wallets

jseagrave21:
https://github.com/CityOfZion/neon-wallet/issues/2575

This PR bypasses a problem with `neon-js` (described in the referenced issue) and allows the token to be displayed correctly in the Activity tab.
This PR also fixes a problem where some NEP-5 token balances are not displayed.

For the first problem, if the original query `getToken` throws an error, we now query Dora to return the token symbol.
For the second problem, we first load all the tokens from Dora. Then we check the originally determined balance and check it against the balance from Dora.  Any additional tokens are added to the `balances` property manually.  I had to use this method because supplementing the [token list](https://github.com/CityOfZion/neo-tokens) with the results from Dora does not work.  Several of the Flamingo tokens (e.g. `fUSDT`) break the code.

Live Testing.  See screenshots below.

1. Dashboard Before.  Note there is no `fUSDT` displayed.
Dashboard Before

2. Dashboard After.
Dashboard After

3. Activity Before.  Note there is no `FLP-nNEO-fUSDT` displayed.
Activity Before

4. Activity After.
Activity After

No

No


jseagrave21:

PR #2644, opened 2024-02-29
CU-86drpnt0d - NEON3 - Adjust frontend implementation to be responsiv…

yumiuehara:
…e horizontaly

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drpnt0d NEON3 - Adjust frontend implementation to be responsive horizontaly](https://app.clickup.com/t/86drpnt0d)

PR #2648, opened 2024-03-04
CU-86drt3wmd - NEON3 - Contact Management - Fix issue where contact l…

yumiuehara:
…ist is not being shown

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drt3wmd NEON3 - Contact Management - Fix issue where contact list is not being shown](https://app.clickup.com/t/86drt3wmd)

PR #2649, opened 2024-03-05
CU-86drt3vcm - NEON3 - Import - Switch to imported wallet automatically

ricardoak25:


melanke:
Task linked: [CU-86drt3vcm NEON3 - Import - Switch to imported wallet automatically](https://app.clickup.com/t/86drt3vcm)

PR #2650, opened 2024-03-05
CU-86drt3y63 - NEON3 - Transaction List - Add a column to indicate to which blockchain that transaction is related

ricardoak25:


melanke:
Task linked: [CU-86drt3y63 NEON3 - Transaction List - Add a column to indicate to which blockchain that transaction is related](https://app.clickup.com/t/86drt3y63)

PR #2653, opened 2024-03-07
CU-86drtv15c - NEON3 - Contact Management - Design Enhancements and N…

yumiuehara:
…NS/ENS support

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drtv15c NEON3 - Contact Management - Design Enhancements and NNS/ENS support](https://app.clickup.com/t/86drtv15c)

PR #2656, opened 2024-03-13
CU-86drupk52 - NEON3 - Onboarding - Update flow based on revised desi…

raulduartep:
…gns (+onboarding import process)

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drupk52 NEON3 - Onboarding - Update flow based on revised designs (+onboarding import process)](https://app.clickup.com/t/86drupk52)

PR #2665, opened 2024-03-14
CU-86drw54ma - NEON2 - Ledger & WC - Investigate issue with getWallet…

raulduartep:
…Info returning isLedger: false when ledger is connected

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drw54ma NEON2 - Ledger & WC - Investigate issue with getWalletInfo returning isLedger: false when ledger is connected](https://app.clickup.com/t/86drw54ma)

melanke (approved):

PR #2670, opened 2024-03-15
CU-86drvtekh - NEON3 - Wrong mandatory token color

Thacryba:

- [ ] Unit tests written?

### DIFF
--- a/src/renderer/src/helpers/UtilsHelper.ts
+++ b/src/renderer/src/helpers/UtilsHelper.ts
@@ -112,20 +112,26 @@ export class UtilsHelper {
     return mandatoryColors[symbol] ?? `#${(0x1000000 + Math.random() * 0xffffff).toString(16).substr(1, 6)}`
   }
 
+  static normalizeHash(hash: string) {
+    return hash.replace('0x', '').toLowerCase()
+  }
+
   static generateTokenColor(hash: string) {
     if (hash.length === 0) throw new Error('Invalid hash')
+
+    const normalizedHash = this.normalizeHash(hash)
 
     const mandatoryColors: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


melanke:
Task linked: [CU-86drvtekh NEON3 - Wrong mandatory token color](https://app.clickup.com/t/86drvtekh)

src/renderer/src/helpers/UtilsHelper.ts
```diff
@@ -116,8 +116,8 @@ export class UtilsHelper {
     if (hash.length === 0) throw new Error('Invalid hash')
 
     const mandatoryColors: Record<string, string> = {
-      ef4073a0f2b305a38ec4050e4d3d28bc40ea63f5: '#56f33f',
-      d2a4cff31913016155e38e474a2c06d08be276cf: '#02c797',
+      '0xef4073a0f2b305a38ec4050e4d3d28bc40ea63f5': '#56f33f',
+      '0xd2a4cff31913016155e38e474a2c06d08be276cf': '#02c797',
```
raulduartep: A better solution for this would be to leave the hash without the 0x and create a new method within UtilsHelper called normalizeHash that remove the 0x from the string and call it before checking the object


```suggestion
       static normalizeHash(hash: string) {
        return hash.replace('0x', '').toLowerCase()
      }
      
     static generateTokenColor(hash: string) {
      const normalizedHash = this.normalizeHash(hash)
      
      const mandatoryColors: Record<string, string> = {
      '0xef4073a0f2b305a38ec4050e4d3d28bc40ea63f5': '#56f33f',
      '0xd2a4cff31913016155e38e474a2c06d08be276cf': '#02c797',
      }
      
      if (mandatoryColors[normalizedHash]) return mandatoryColors[normalizedHash]
      ... rest of method
```

PR #2686, opened 2024-03-28
CU-86drpnt0k - NEON2 - Implement changes on NWD to allow adding witne…

raulduartep:
…ss using the new version of WcSdk

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drpnt0k NEON2 - Implement changes on NWD to allow adding witness using the new version of WcSdk](https://app.clickup.com/t/86drpnt0k)

melanke (approved):

PR #2691, opened 2024-04-01
﻿CU-86dt1kmw4 - NEON3 - Wallets/Portfolio Screen - Transactions Design fixes

Thacryba:

- [ ] Unit tests written?


melanke:
Task linked: [CU-86dt1kmw4 NEON3 - Wallets/Portfolio Screen - Transactions Design fixes](https://app.clickup.com/t/86dt1kmw4)

src/renderer/src/locales/en/components.json
```diff
@@ -58,8 +58,8 @@
     "date": "DATE",
     "asset": "ASSET",
     "amount": "AMOUNT",
-    "from": "FROM",
-    "to": "TO",
+    "from": "ACCOUNT",
+    "to": "ADDRESS",
```
raulduartep, line 62: You can leave it as it was. I had already aligned with Balducci on this headers

PR #2701, opened 2024-04-08
CU-86drpnrpz - NEON3 - Implement Sentry or similar tool to track erro…

LeonardoDizConde:
…rs in production

- [ ] Unit tests written?


melanke:
Task linked: [CU-86drpnrpz NEON3 - Implement Sentry or similar tool to track errors in production](https://app.clickup.com/t/86drpnrpz)

PR #2703, opened 2024-04-08
CU-86dt26yh2 - NEON3 - Wallets Screen - Implement "Claim Gas" Feature

yumiuehara:

- [ ] Unit tests written?


melanke:
Task linked: [CU-86dt26yh2 NEON3 - Wallets Screen - Implement "Claim Gas" Feature](https://app.clickup.com/t/86dt26yh2)

src/renderer/src/routes/pages/Wallets/CommonAccountActions.tsx
```diff
@@ -15,6 +17,8 @@ export const CommonAccountActions = ({ account }: TProps) => {
 
   return account?.type !== 'watch' ? (
     <div className="flex gap-2">
+      {account.blockchain !== 'ethereum' && <ClaimGasButton account={account} />}
```
raulduartep: You should check if account.blockchain is claimable instead.

PR #2716, opened 2024-04-18
CU-86dta847p - NEON3 - Adjust the way installation works so users can…

raulduartep:
… select target folder (and default is not the same as NEON2)

- [ ] Unit tests written?

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -18,10 +18,12 @@ win:
       arch:
         - x64
 nsis:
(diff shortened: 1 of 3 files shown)


melanke:
Task linked: [CU-86dta847p NEON3 - Adjust the way installation works so users can select target folder (and default is not the same as NEON2)](https://app.clickup.com/t/86dta847p)

thiagocbalducci (approved):

PR #2728, opened 2024-05-21
86dthjvx1 - NEON2 - Investigate the problem on NWD's WC deeplink when using Web3modal

ricardoak25:


melanke:

main.js
```diff
@@ -260,7 +260,10 @@ ipcMain.handle('dialog', async (event, method, params) => {
 })
 
 ipcMain.handle('getInitialDeepLinkUri', async () => {
-  const uri = initialDeepLinkUri
+  let uri = initialDeepLinkUri
+  if (uri && uri.indexOf('/wc?uri=') !== -1) {
+    uri = uri.replace('/wc?uri=', '')
+  }
```
melanke:
```suggestion
  let uri = initialDeepLinkUri
  if (uri && uri.startsWith('/wc?uri=')) {
    // the new format comes with this prefix
    uri = uri.replace('/wc?uri=', '')
  } else {
    // the legacy format needs to be decoded
    uri = atob(uri)
  }
```

app/containers/ConnectDapp/ConnectDapp.jsx
```diff
@@ -55,12 +55,24 @@ const ConnectDapp = ({ showErrorNotification, history }: Props) => {
...
       if (!uri) return
 
-      const decoded = atob(uri)
-      handleOnURI(decoded)
+      if (isBase64(uri)) {
+        const decoded = atob(uri)
+        handleOnURI(decoded)
+      } else {
+        handleOnURI(uri)
+      }
```
melanke:
```suggestion
      handleOnURI(uri)
```

melanke (changes requested):
I think we can decode the uri only if we are sure it's the old format, which doesn't have the prefix. Can you make this changes and test again?

melanke (approved):

PR #2730, opened 2024-05-22
86dthjvx1 - NEON2 - Investigate the problem on NWD's WC deeplink when using Web3modal

ricardoak25:


melanke:

main.js
```diff
@@ -261,12 +261,12 @@ ipcMain.handle('dialog', async (event, method, params) => {
 
 ipcMain.handle('getInitialDeepLinkUri', async () => {
   let uri = initialDeepLinkUri
-  if (uri && uri.startsWith('/wc?uri=')) {
+  if (uri && uri.startsWith('neon://uri=/wc?uri=')) {
     // the new format comes with this prefix
```
melanke:
```suggestion
    // the new format comes with this prefix and it's not encoded. So, we are removing the prefix and encoding, to keep the old logic working for the old format
```

melanke (changes requested):

melanke (approved):

PR #2731, opened 2024-05-28
CU-86dthjvx1 - NEON2 - Investigate the problem on NWD's WC deeplink w…

raulduartep:
…hen using Web3modal

- [ ] Unit tests written?


melanke:

melanke (approved):

PR #2735, opened 2024-06-12
86dtnhvkm - NEON2 - Migration - Implement deeplinking so the user can continue migration on NEON3 automatically

ricardoak25:


melanke:
Task linked: [CU-86dtnhvkm NEON2 - Migration - Implement deeplinking so the user can continue migration on NEON3 automatically](https://app.clickup.com/t/86dtnhvkm)

raulduartep (approved):

